PR #8, opened 2021-02-19
Remodel input handling

kraflab:
Complete rewrite of input bindings - how they are configured, stored, and accessed during gameplay. Input is (mostly) abstracted behind an interface at `dsda/input`. The knowledge of key / button states, how many keys are configured, etc is isolated in the input file. In various places the raw input is still used (e.g., key presses going into text entry), but _bound actions_ go through the dsda input code.

- you can now bind mouse and joy buttons to everything
- you can now store 3 separate input profiles
- you can now configure multiple keys to the same action


fabiangreffrath:
Awesome!

PR #11, opened 2021-03-08
Fix missing newline in timedemo printout

Phytolizer:
I found a missing newline in the -timedemo printout. By the way, I'm Phytolizer. Hi! :)

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -4269,7 +4269,7 @@ dboolean G_CheckDemoStatus (void)
 
       M_SaveDefaults();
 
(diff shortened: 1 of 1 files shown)


kraflab:
Thanks, could you also fix this upstream? https://github.com/coelckers/prboom-plus/blob/master/prboom2/src/g_game.c#L4251

kraflab:
Actually maybe I_Error injects the `\n` on its own when it prints to the console...nevermind I guess 🤷

kraflab:
`lprintf(LO_ERROR, "%s\n", errmsg);` yep, it adds it here.

PR #20, opened 2021-04-03
New MBF21 DEHACKED codepointers (first batch)

XaserAcheron:
Got the first handful of codepointers done. There's a few more I may try and sneak in (melee attacks and some of skillsaw's suggestions), but some of 'em will need a bit of research to figure out how best to do 'em, and I don't want to hold up the essentials any longer than I must. ;)

Quick summary of the 'pointers:
- Actor pointers:
  - **A_SpawnFacing(type, height)** -- spawns an actor of `type` at `height` z units and sets its angle to the caller's angle.
  - **A_MonsterProjectile(type, angle)** -- generic monster projectile attack; always sets `tracer` field.
  - **A_MonsterBulletAttack(damage, spread)** -- generic monster bullet attack w/horizontal `spread`; if `spread` is negative, apply vertical spread equal to 2/3 of this value (approx. equal to SSG vert-spread)
  - **A_RadiusDamage(damage, radius)** -- generic A_Explode, w/customizable damage and radius (helll yeah)
- Weapon pointers:
  - **A_WeaponProjectile(type, angle)** -- generic weapon projectile attack; does not consume ammo
  - **A_WeaponBulletAttack(damage, spread)** -- generic weapon bullet attack; does not consume ammo; same `spread` behavior as A_MonsterBulletAttack
  - **A_WeaponSound(sound, fullvol)** -- same as A_PlaySound, but for weapons
  - **A_WeaponJump(state, chance)** -- same as A_RandomJump, but for  weapons
  - **A_ConsumeAmmo(amount)** -- subtract `amount` units of ammo. if `amount` is zero, use the weapon slot's `ammopershot`. will not reduce ammo below zero.
  - **A_CheckAmmo(state, amount)** -- jumps to `state` if ammo is below `amount`; if `amount` is zero, use the weapon slot's `ammopershot` value instead
  - **A_RefireTo(state, noammocheck)** -- jumps to `state` if trigger is still held down; will also check ammo unless `noammocheck` is set
  - **A_GunFlashTo(state, nothirdperson)** -- sets the weapon's flash state to `state`; also sets the player's 3rd-person sprite to the player actor's firing frame unless `nothirdperson` is set

A bit more than the 5 pointers I originally planned, but eh. Low-hanging fruit gets monched. :P


prboom2/src/m_random.c
```diff
@@ -146,6 +147,45 @@ void M_ClearRandom (void)
...
 }
 
+// [XA] Common random formulas used by codepointers
+
+//
+// P_RandomHitscanAngle
+// Outputs a random angle between (-spread, spread), as an int ('cause it can be negative).
+//   spread: Maximum angle (degrees, in fixed point -- not BAM!)
+//
+int P_RandomHitscanAngle(pr_class_t pr_class, fixed_t spread)
```
kraflab, line 157: If you're taking pr_class as an argument, you should use it for the P_Random calls, right? :^)

prboom2/src/p_map.c
```diff
@@ -2058,13 +2060,21 @@ dboolean PIT_RadiusAttack (mobj_t* thing)
...
     {
     // must be in direct path
-    P_DamageMobj (thing, bombspot, bombsource, bombdamage - dist);
+
+    // [XA] independent damage/distance calculation.
+    //      same formula as eternity; thanks Quas :P
+    if (bombdamage == bombdistance)
+      damage = bombdamage - dist;
+    else
+      damage = (bombdamage * (bombdistance - dist) / bombdistance) + 1;
```
kraflab, line 2075: Can bombdistance be 0?

prboom2/src/p_pspr.c
```diff
@@ -84,14 +84,63 @@ static const int recoil_values[] = {    // phares
...
+	NULL
+};
+
+//
+// P_IsParamWeaponPtr
+// Returns true if the specified action is a
+// parameterized weapon code pointer.
+//
+dboolean P_IsParamWeaponPtr(const actionf_t ptr)
+{
```
kraflab, line 111: Should we be fancy?

```c
const actionf_t* weapon_ptr;

if (ptr == NULL)
  return false;

for (weapon_ptr = param_weapon_ptrs; *weapon_ptr != NULL, weapon_ptr++)
  if (*weapon_ptr == ptr)
    return true;

return false;

File: prboom2/src/p_pspr.c [hunk 4]
```diff
@@ -84,14 +84,63 @@ static const int recoil_values[] = {    // phares
...
+  while (param_weapon_ptrs[i] != NULL);
+
+  return false;
+}
+
 //
 // P_SetPsprite
 //
 
 void P_SetPsprite(player_t *player, int position, statenum_t stnum)
```
kraflab, line 128: This code isn't something that gets called a lot in one frame, right? Just making sure adding a loop lookup through the weapon pointer array isn't going to affect performance.

prboom2/src/m_random.c
```diff
@@ -146,6 +147,45 @@ void M_ClearRandom (void)
...
+{
+  int t;
+  int_64_t spread_bam;
+
+  // FixedToAngle doesn't work for negative numbers,
+  // so for convenience take just the absolute value.
+  spread_bam = (spread < 0 ? FixedToAngle(-spread) : FixedToAngle(spread));
+  t = P_Random(pr_mbf21);
+  return (int)((spread_bam * (t - P_Random(pr_mbf21))) >> 8);
```
kraflab: Is `8` a magic number or is there a defined constant we could put here?

prboom2/src/p_pspr.c
```diff
@@ -257,6 +306,25 @@ int P_WeaponPreferred(int w1, int w2)
...
+//
+// P_GetAmmoPerShot
+// Returns ammo per shot for the currently selected weapon.
+//
+// [XA] Eventually this value will be customizable
+// instead of hardcoded, presuming everything goes
+// according to plan...
+//
+
+int P_GetAmmoPerShot(player_t *player)
```
kraflab, line 313: This isn't hit by any existing code, right? It will only be hit downstream of the new code pointers? Just making sure heretic doesn't need to be considered :^)

prboom2/src/m_random.c
XaserAcheron: Welp, brainfart. :P -- Fixed.

prboom2/src/p_map.c
XaserAcheron: Fortunately not -- the function always returns early if bombdistance is zero on the `//out of range` check a couple of lines before.

prboom2/src/p_pspr.c
XaserAcheron: Nah, should be fine -- there's only ever 2 psprites per player (weapon and flash) -- maybe only 2 ever? I dunno if classic doom MP even creates psprites for the other players. If this was something that applied to actor states, then maybe we'd have a problem. :P

prboom2/src/m_random.c
XaserAcheron: `>> 8` here is just a nicer `/ 255`, since we're effectively generating a random number between -255 and 255 and I wanna scale it to (-1,1). I guess there's no harm in defining a `RANDBITS` constant or something though.

prboom2/src/p_pspr.c
XaserAcheron: Heh, sure. :P

prboom2/src/m_random.c
XaserAcheron: On second thought, maybe it actually should just be `/ 255`, since the value can be negative, and a right shift on that might be UB? lemme googlify it

prboom2/src/p_pspr.c
XaserAcheron: Currently no, but the eventual plan is for P_CheckAmmo to call this function `if (mbf21)` in my eventual `ammopershot` PR. Means it still ought not affect Heretic though; presuming you can't set the new complevel there. ;)

prboom2/src/m_random.c
XaserAcheron: Replaced it with `/ 255` -- still just using the number instead of a constant, but eh.

kraflab: Sounds fine to me

prboom2/src/p_pspr.c
kraflab: Instead of a method, what do you think about using an array `shotammo` to match the existing `maxammo` and `clipammo`? That will fit in with the use of `maxammo` and `clipammo` in d_deh, assuming we will be making it all configurable the same way.

XaserAcheron: That makes sense -- though that seems like a thing to do for the new `ammopershot` PR. I'll probably rip out the P_GetAmmoPerShot function completely if it's indeed as simple as indexing an array. ;)

prboom2/src/p_enemy.c
```diff
@@ -2964,6 +2964,104 @@ void A_LineEffect(mobj_t *mo)
...
+// Basically just A_Spawn with a major quality-of-life tweak.
+//   misc1: Type of actor to spawn
+//   misc2: Height to spawn at, relative to calling actor
+//
+void A_SpawnFacing(mobj_t *actor)
+{
+  if (!mbf21 || !actor->state->misc1)
+    return;
+
+  mobj_t *mo = P_SpawnMobj(actor->x, actor->y, (actor->state->misc2 << FRACBITS) + actor->z, actor->state->misc1 - 1);
```
kraflab: We need to declare this variable at the top and then set it here to avoid a warning in C, right? 🤓

kraflab:
Please add a code pointers header in https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/mbf21.md and link this pr as well as list the ones added here. Just basic info is fine - it isn't meant to be a spec for public consumption so we can go into more detail once mbf21 is more finalized.

prboom2/src/p_enemy.c
XaserAcheron: Heh, maybe. :P -- Fixed for consistency at least. ;)

XaserAcheron:
Just pushed an update to mbf21.md -- in this PR; it's self-referential now. :P

kraflab (approved):

PR #21, opened 2021-04-07
Timer compile fix for MSVC

XaserAcheron:
Can't guarantee this is all kosher, but it was enough to get me up and running, so here's the PR. :P

### DIFF
--- a/prboom2/src/dsda/time.c
+++ b/prboom2/src/dsda/time.c
@@ -20,7 +20,7 @@
 #include "time.h"
 
 // [XA] MSVC hack: need to implement a handful of functions here.
-// this was yoinked from a handful of stackoverflow posts so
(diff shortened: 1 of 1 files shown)


kraflab:
Is `struct timespec` already defined?

XaserAcheron:
Ye, it seems to be -- compiler went "hey yo this already exists" on that line so I took it out and it worked. ;)

[EDIT] It's in Microsoft's version of `<time.h>`

kraflab:
The only remaining concern is "does it work" but since the official builds aren't made this way and you made it clear in the comment, better to have something that can compile for devs in this environment in the short term at least.

XaserAcheron:
Heh, yeah, this is definitely "swim at your own risk" territory, IMO.

PR #24, opened 2021-04-15
Add "Ammo per shot" DEHACKED field

XaserAcheron:
Adds a new `Ammo per shot` field to DEHACKED. Implementation mirrors Eternity's, for forward-compat n' whatnot.

A summary of sorts:
- Weapons now use this field to determine if the player has enough ammo to fire, instead of hardcoding the value for each weapon.
- Weapon autoswitch has been fixed in general to check the correct ammotype & ammo use for each weapon, rather than assume the vanilla ammotypes (this is an ooooold bug from ever since DEHACKED was introduced).
- The new `A_ConsumeAmmo` and `A_CheckAmmo` MBF21 DEHACKED codepointers will use the ready weapon's `Ammo per shot` field if the `amount` arg is set to zero.
- Vanilla doom weapon fire codepointer behavior varies on whether or not the `Ammo per shot` field is explicitly set on a weapon.
  - If set, the ready weapon's `Ammo per shot` field is subtracted from the weapon's ammo type.
  - If not set, the amount of ammo subtracted depends on the weapon codepointer being used (e.g. A_FireShotgun2 will subtract 2, A_FireBFG will subtract BFGCELLS -- this mirrors vanilla Doom behavior)
- For general sanity, the "BFG Cells/Shot" field will also set the BFG weapon's `Ammo per shot` field.

There's a little bit of refactoring of things that touches a bit of Heretic (namely, all the various ammo-per-shot arrays for the doom/heretic weapons were replaced with the new weaponinfo_t ammopershot field ), though this feature is not yet fully implemented for Heretic 'cause it doesn't do DEH. ;) -- I tested Heretic to make sure it all works as intended still (and the spec passes), so all should be well there.

[Will update mbf21.md here in a bit; need to make the PR first. ;]

### DIFF
--- a/prboom2/src/d_deh.c
+++ b/prboom2/src/d_deh.c
@@ -1252,7 +1252,8 @@ static const char *deh_weapon[] = // CPhipps - static const*
   "Select frame",   // .downstate
   "Bobbing frame",  // .readystate
   "Shooting frame", // .atkstate
-  "Firing frame"    // .flashstate
+  "Firing frame",   // .flashstate
+  "Ammo per shot",  // .ammopershot [XA] new to mbf21
 };
 
 // CHEATS - Dehacked block name = "Cheat"
@@ -2357,7 +2358,13 @@ static void deh_procWeapon(DEHFILE *fpin, FILE* fpout, char *line)
                 if (!deh_strcasecmp(key,deh_weapon[5]))  // Firing frame
                   weaponinfo[indexnum].flashstate = (int)value;
                 else
-                  if (fpout) fprintf(fpout,"Invalid weapon string index for '%s'\n",key);
+                  if (!deh_strcasecmp(key,deh_weapon[6]))  // Ammo per shot
+                    {
+                      weaponinfo[indexnum].ammopershot = (int)value;
+                      weaponinfo[indexnum].intflags |= WIF_ENABLEAPS;
+                    }
+                  else
+                    if (fpout) fprintf(fpout,"Invalid weapon string index for '%s'\n",key);
     }
   return;
 }
@@ -2635,7 +2642,7 @@ static void deh_procMisc(DEHFILE *fpin, FILE* fpout, char *line) // done
                                   idkfa_armor_class = (int)value;
                                 else
                                   if (!deh_strcasecmp(key,deh_misc[14]))  // BFG Cells/Shot
-                                    bfgcells = (int)value;
+                                    weaponinfo[MT_BFG].ammopershot = bfgcells = (int)value;
                                   else
                                     if (!deh_strcasecmp(key,deh_misc[15])) { // Monsters Infight
                                       // e6y: Dehacked support - monsters infight
--- a/prboom2/src/d_deh.c
+++ b/prboom2/src/d_deh.c
@@ -57,6 +57,18 @@
 #define TRUE 1
 #define FALSE 0
 
+static dboolean bfgcells_modified = false;
+
+void CheckDehConsistency(void)
+{
+  if (
+    bfgcells_modified &&
+    weaponinfo[MT_BFG].intflags & WIF_ENABLEAPS &&
+    bfgcells != weaponinfo[MT_BFG].ammopershot
+  )
+    I_Error("Mismatch between bfgcells and bfg ammo per shot modifications! Check your dehacked.");
+}
+
 // e6y: for compatibility with BOOM deh parser
 int deh_strcasecmp(const char *str1, const char *str2)
 {
@@ -2719,7 +2731,10 @@ static void deh_procMisc(DEHFILE *fpin, FILE* fpout, char *line) // done
                                   idkfa_armor_class = (int)value;
                                 else
                                   if (!deh_strcasecmp(key,deh_misc[14]))  // BFG Cells/Shot
+                                  {
                                     weaponinfo[MT_BFG].ammopershot = bfgcells = (int)value;
+                                    bfgcells_modified = true;
+                                  }
                                   else
                                     if (!deh_strcasecmp(key,deh_misc[15])) { // Monsters Infight
                                       // e6y: Dehacked support - monsters infight
--- a/prboom2/src/d_items.h
+++ b/prboom2/src/d_items.h
@@ -44,7 +44,7 @@
 //
 // Internal weapon flags
 //
-#define WIF_ENABLEAPS (0x0000000000000001) // [XA] enable "ammo per shot" field for native Doom weapon codepointers
+#define WIF_ENABLEAPS 0x00000001 // [XA] enable "ammo per shot" field for native Doom weapon codepointers
 
 // haleyjd 09/11/07: weapon flags
 //
--- a/prboom2/src/d_items.h
+++ b/prboom2/src/d_items.h
@@ -41,6 +41,10 @@
 #pragma interface
 #endif
 
+//
+// Internal weapon flags
+//
+#define WIF_ENABLEAPS (0x0000000000000001) // [XA] enable "ammo per shot" field for native Doom weapon codepointers
 
 /* Weapon info: sprite frames, ammunition use. */
 typedef struct
@@ -52,10 +56,11 @@ typedef struct
   int         atkstate;
   int         holdatkstate;
   int         flashstate;
+  int         ammopershot;
+  int         intflags;
 } weaponinfo_t;
 
 extern weaponinfo_t doom_weaponinfo[NUMWEAPONS+2];
-extern int ammopershot[NUMWEAPONS+2];
 
 // heretic
 
--- a/prboom2/src/p_pspr.c
+++ b/prboom2/src/p_pspr.c
@@ -303,17 +303,17 @@ int P_SwitchWeapon(player_t *player)
   int currentweapon, newweapon;
   int i;
 
-  prefer = weapon_preferences[demo_compatibility != 0]; // killough 3/22/98
-  currentweapon = player->readyweapon;
-  newweapon = currentweapon;
-  i = NUMWEAPONS + 1;   // killough 5/2/98
-
   // [XA] use fixed behavior for mbf21. no need
   // for a discrete compat option for this, as
   // it doesn't impact demo playback (weapon
   // switches are saved in the demo itself)
   if (mbf21)
     return P_SwitchWeaponMBF21(player);
+
+  prefer = weapon_preferences[demo_compatibility != 0]; // killough 3/22/98
+  currentweapon = player->readyweapon;
+  newweapon = currentweapon;
+  i = NUMWEAPONS + 1;   // killough 5/2/98
 
   // killough 2/8/98: follow preferences and fix BFG/SSG bugs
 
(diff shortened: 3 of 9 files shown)


XaserAcheron:
Quick-ish update: needed to add a check for `wp_noammo` in P_SubtractAmmo, and while I was in the neighborhood, I went ahead and made it so in `mbf21` ammo cannot go into the negatives, 'cause nobody wants that. :P

XaserAcheron:
Just updated `mbf21.md` with some info on the new stuffs.

prboom2/src/p_pspr.c
```diff
@@ -2228,19 +2293,19 @@ void P_UpdateBeak(player_t * player, pspdef_t * psp)
 dboolean Heretic_P_CheckAmmo(player_t * player)
 {
     ammotype_t ammo;
-    int *ammoUse;
+    weaponinfo_t *weaponinfo;
```
kraflab: Let's not call that `weaponinfo` - I think that's a recipe for future chaos when the global has the same name :smile: 
Maybe `checkweaponinfo` :shrug:

prboom2/src/p_pspr.c
```diff
@@ -365,6 +424,36 @@ dboolean P_CheckAmmo(player_t *player)
...
+
+  if (mbf21 && ammotype == am_noammo)
+    return; // [XA] hmm... I guess vanilla/boom will go out of bounds then?
+
+  if (mbf21 && (weaponinfo[player->readyweapon].intflags & WIF_ENABLEAPS))
+    amount = weaponinfo[player->readyweapon].ammopershot;
+  else
+    amount = compat_amt;
+
+  if (!mbf21 || player->ammo[ammotype] >= amount)
```
kraflab: Not sure I understand this. The implication is somewhere in the code we called this method despite not having enough ammo? Is it possible? At any rate we should subtract the ammo we do have left in that case, and set the ammo to zero instead of skipping the subtraction.

prboom2/src/p_pspr.c
```diff
@@ -365,6 +424,36 @@ dboolean P_CheckAmmo(player_t *player)
...
+// defined in dehacked; otherwise use the amount specified
+// by the codepointer instead (Doom behavior)
+//
+// [XA] NOTE: this function is for handling Doom's native
+// attack pointers; of note, the new A_ConsumeAmmo mbf21
+// codepointer does NOT call this function, since it doesn't
+// have to worry about any compatibility shenanigans.
+//
+
+void P_SubtractAmmo(struct player_s *player, int compat_amt)
```
kraflab: I'd like a better name for `compat_amt`. Maybe `vanilla_amount`?

prboom2/src/p_pspr.c
```diff
@@ -335,12 +390,16 @@ dboolean P_CheckAmmo(player_t *player)
...
+    count = weaponinfo[player->readyweapon].ammopershot;
   else
-    if (player->readyweapon == wp_supershotgun)        // Double barrel.
-      count = 2;
+    if (player->readyweapon == wp_bfg)  // Minimal amount for one shot varies.
+      count = BFGCELLS;
+    else
+      if (player->readyweapon == wp_supershotgun)        // Double barrel.
+        count = 2;
+	  else
```
kraflab: Whitespace :^)

prboom2/src/p_pspr.c
```diff
@@ -221,6 +221,73 @@ int weapon_preferences[2][NUMWEAPONS+1] = {
...
+
+    if (checkweapon != wp_nochange && player->weaponowned[checkweapon])
+    {
+      ammotype = weaponinfo[checkweapon].ammo;
+      if (ammotype == am_noammo ||
+        player->ammo[ammotype] >= weaponinfo[checkweapon].ammopershot)
+        newweapon = checkweapon;
+    }
+  }
+  while (newweapon==currentweapon && --i);
```
kraflab: Please give the `==` some space :angel:

prboom2/src/p_pspr.c
```diff
@@ -221,6 +221,73 @@ int weapon_preferences[2][NUMWEAPONS+1] = {
...
+// instead of blindly assuming both.
+
+static int P_SwitchWeaponMBF21(player_t *player)
+{
+  int *prefer = weapon_preferences[0];
+  int currentweapon = player->readyweapon;
+  int newweapon = currentweapon;
+  int i = NUMWEAPONS+1;
+
+  weapontype_t checkweapon;
```
kraflab, line 233: C declarations again:
```c
int *prefer;
int currentweapon, newweapon;
int i;
weapontype_t checkweapon;
ammotype_t ammotype;

prefer = weapon_preferences[0];
currentweapon = player->readyweapon;
newweapon = currentweapon;
i = NUMWEAPONS + 1;
```

If you want to fix that in the original method too, I'm all for it, but at least let's have it fixed in the new code.

prboom2/src/d_deh.c
```diff
@@ -2712,7 +2719,7 @@ static void deh_procMisc(DEHFILE *fpin, FILE* fpout, char *line) // done
                                   idkfa_armor_class = (int)value;
                                 else
                                   if (!deh_strcasecmp(key,deh_misc[14]))  // BFG Cells/Shot
-                                    bfgcells = (int)value;
+                                    weaponinfo[MT_BFG].ammopershot = bfgcells = (int)value;
```
kraflab, line 2735: I don't think we should set this here. I'd rather have the meaning of this dehacked field be the same as it has been, which is overwriting the bfgcells value. The way it is now, it doesn't set WIF_ENABLEAPS anyway, which could be confusing, and the alternative of setting it would have ramifications for compatibility. Better to let it rest as it has been.

prboom2/src/p_pspr.c
XaserAcheron: Yup, it's very easy to accidentally go into the negatives in vanilla DEHACKED -- random example, if you have a dual pistol weapon that calls A_FirePistol twice, and fire the weapon with 1 ammo remaining, you'll go to -1 ammo and the HUD will look bad. :P

Of course, preventing this kinda thing is exactly what the new `ammopershot` field is for, so this problem is avoidable simply by using mbf21 dehacked features correctly, but negative ammo is never a good thing in any case so I figured I'd guard against it outright.

This is really just a "nice to have", so if you feel strongly about it I can take it out -- or if you'd rather see the logic be "subtract, then set to zero if negative", that works for me -- bit cleaner that way, thinking on it :P

prboom2/src/p_pspr.c
XaserAcheron: Welpspace. Will fix.

prboom2/src/d_deh.c
XaserAcheron: This is mainly here for forwards-compatibility with Eternity -- EE does this exact thing, so if we don't do the same here, there may be divergent behavior.

prboom2/src/p_pspr.c
XaserAcheron: Good callout -- will fix.

prboom2/src/p_pspr.c
XaserAcheron: Works for me. :P

prboom2/src/p_pspr.c
XaserAcheron: It be fixed.

prboom2/src/p_pspr.c
XaserAcheron: Aaaand, done.

prboom2/src/p_pspr.c
XaserAcheron: All done.

prboom2/src/p_pspr.c
XaserAcheron: Done. Tim Curry is pleased.

prboom2/src/p_pspr.c
XaserAcheron: Fixed both, since I was in the neighborhood. :P

prboom2/src/p_pspr.c
kraflab: Ya - just substract it and then do "if negative and mbf21, = 0" I think. Then at least if there's some weird mixup, it won't make the same mistake repeatedly (i.e., 1 will drop to 0 and then it should detect there's no ammo, hopefully).

prboom2/src/d_deh.c
kraflab: I think that's something to adjust in ee then personally 🤷 
Having 2 fields that write the same value, but differently, seems like a dangerous path to take. And making an old key have a partially new meaning is also not safe to me.

prboom2/src/d_items.h
```diff
@@ -41,6 +41,10 @@
 #pragma interface
 #endif
 
+//
+// Internal weapon flags
+//
+#define WIF_ENABLEAPS (0x0000000000000001) // [XA] enable "ammo per shot" field for native Doom weapon codepointers
```
kraflab: This should just be `0x00000001` since it's 32 bits.

prboom2/src/p_pspr.c
```diff
@@ -229,10 +299,21 @@ int weapon_attack_alignment=0;
...
 {
-  int *prefer = weapon_preferences[demo_compatibility!=0]; // killough 3/22/98
-  int currentweapon = player->readyweapon;
-  int newweapon = currentweapon;
-  int i = NUMWEAPONS+1;   // killough 5/2/98
+  int *prefer;
+  int currentweapon, newweapon;
+  int i;
+
+  prefer = weapon_preferences[demo_compatibility != 0]; // killough 3/22/98
```
kraflab: Let's move these lines past the branch condition so we can save precious picoseconds 🙏

kraflab:
I'll resolve the conflicts, sec

prboom2/src/d_deh.c
kraflab: Let's do this:
1) Don't set ammopershot here, but set `bfgcells_modified = true`.
2) After processing the deh, do this:
```c
if (bfgcells_modified && 
    weaponinfo[MT_BFG].intflags & WIF_ENABLEAPS && 
    bfgcells != weaponinfo[MT_BFG].ammopershot)
  I_Error("Mismatch between bfgcells and bfg ammo per shot modifications! Check your dehacked.");
```

kraflab: I guess that doesn't solve the problem of using a new codepointer that will reference ammopershot though, is that the real problem? :thinking:

kraflab: If that's the case, then DO set ammopershot here, but I want to raise I_Error if someone tries to set both ammopershot in the weaponinfo and bfgcells and they don't match. As long as that happens I'm all aboard.

prboom2/src/p_pspr.c
kraflab: I took care of it 🎵

prboom2/src/d_items.h
kraflab: I took care of it 🎵

prboom2/src/p_pspr.c
kraflab: I took care of it 🎵

prboom2/src/d_deh.c
kraflab: https://github.com/kraflab/dsda-doom/pull/24/commits/70f7a405feaa440dd1fb8b504811a8723587ad8a took care of this here - we can drop it later if you think we should, but let's not get stuck on this point.

PR #30, opened 2021-04-17
MBF21 State Args

XaserAcheron:
Adds an `args` array to the state table, exposes it to DEHACKED, and updates the new codepointers to use args instead of misc1/misc2.  'Nuff said.

[Will update mbf21.md here in a bit; gotta get food n' do some things]

### DIFF
--- a/prboom2/src/d_deh.c
+++ b/prboom2/src/d_deh.c
@@ -59,16 +59,6 @@
 
 static dboolean bfgcells_modified = false;
 
-void CheckDehConsistency(void)
-{
-  if (
-    bfgcells_modified &&
-    weaponinfo[MT_BFG].intflags & WIF_ENABLEAPS &&
-    bfgcells != weaponinfo[MT_BFG].ammopershot
-  )
-    I_Error("Mismatch between bfgcells and bfg ammo per shot modifications! Check your dehacked.");
-}
-
 // e6y: for compatibility with BOOM deh parser
 int deh_strcasecmp(const char *str1, const char *str2)
 {
@@ -1362,6 +1352,7 @@ typedef struct {
   actionf_t cptr;  // actual pointer to the subroutine
   const char *lookup;  // mnemonic lookup string to be specified in BEX
   // CPhipps - const*
+  int argcount;  // [XA] number of mbf21 args this action uses, if any
 } deh_bexptr;
 
 static const deh_bexptr deh_bexptrs[] = // CPhipps - static const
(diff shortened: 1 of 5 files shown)


XaserAcheron:
Oop, something else I want to do before this gets merged is add a sanity check to the DEH parser to reject a patch that specifies too many nonzero args for a codepointer. Should be easy to sneak into the new CheckDEHConsistency function.

kraflab:
Looks good to me, whenever you finish the other stuff you mentioned, we can merge it up

XaserAcheron:
All done -- sorry it took me a week to get around to it. Must've been a long lunch break. :P

prboom2/src/d_deh.c
```diff
@@ -3522,3 +3513,47 @@ dboolean deh_GetData(char *s, char *k, uint_64_t *l, char **strval, FILE *fpout)
...
+    {
+      ++j;
+      if (states[i].action == deh_bexptrs[j].cptr)
+      {
+        bexptr_name = deh_bexptrs[j].lookup;
+        maxargs = deh_bexptrs[j].argcount;
+        found = TRUE;
+      }
+    } while (!found && (deh_bexptrs[j].cptr != NULL));
+
```
kraflab, line 3547: Don't we need an "if not found, continue" here?

XaserAcheron: Not really -- if there's no matching codepointer, we want to make sure there are zero args defined. Probably doesn't matter in practice for null-action states, but it's nice n' consistent.

That said, in practice, states with no codepointer match the A_NULL entry, so there's never not a match. :P

XaserAcheron: Not really -- if there's no matching codepointer, we want to make sure there are zero args defined. Probably doesn't matter in practice for null-action states, but it's nice n' consistent.

That said, in practice, states with no codepointer match the A_NULL entry, so there's never not a match. :P

kraflab: Oh, but then why do you have a found variable? 🤔 That's what was throwing me off. I.e., just `break` when you find something?

kraflab: Also, is something stopping this from just being `for (j = 0; deh_bexptrs[j].cptr != NULL; ++j)`? The `do while` doesn't feel natural here to me :^)

XaserAcheron: Changed to a `break`. A classic case of "not enough coffee" :P

kraflab: I updated the loop too 🧐 
If there's anything wrong let me know and we'll fix it - merging for now.

XaserAcheron: Nah, is good -- the do/while was me copying the old code style, but I'm all for de-Killough-izing the new bits if you are. :P

PR #36, opened 2021-04-25
Changed the IWAD directory for MacOS

Pedro-Beirao:
(MacOS only) The launcher uses one folder to find the IWADs (/Library/Application Support/PrBoom-Plus) and another folder to load them from (/.dsda-doom). This change makes the launcher use the same folder for both finding and loading the IWAD.


kraflab:
Thanks!

PR #40, opened 2021-05-01
MBF21 Melee Codepointers

XaserAcheron:
Adds two new MBF21 codepointers:
- A_MonsterMeleeAttack(damagebase, damagedice, sound, range)
- A_WeaponMeleeAttack(damagebase, damagedice, zerkfactor, sound, range)

Details in `mbf21.md` -- pretty straightforward now that `args` are a thing. :D

### DIFF
--- a/prboom2/src/p_pspr.c
+++ b/prboom2/src/p_pspr.c
@@ -1221,6 +1221,64 @@ void A_WeaponBulletAttack(player_t *player, pspdef_t *psp)
 }
 
 //
+// A_WeaponMeleeAttack
+// A parameterized player weapon melee attack.
+//   args[0]: Base damage of attack (e.g. for 2d10, customize the 2); if not set, defaults to 2
+//   args[1]: Attack damage modulus (e.g. for 2d10, customize the 10); if not set, defaults to 10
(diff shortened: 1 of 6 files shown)


prboom2/src/p_pspr.c
```diff
@@ -1220,6 +1220,64 @@ void A_WeaponBulletAttack(player_t *player, pspdef_t *psp)
...
+    damage = (damage * zerkfactor) >> FRACBITS;
+
+  // slight randomization; weird vanillaism here. :P
+  angle = player->mo->angle;
+
+  t = P_Random(pr_mbf21);
+  angle += (t - P_Random(pr_mbf21))<<18;
+
+  /* killough 8/2/98: make autoaiming prefer enemies */
+  // [XA] ain't touchin' this.
```
kraflab: I believe in you 😉 

I guess it should be this:
```c
slope = P_AimLineAttack(player->mo, angle, range, MF_FRIEND);

if (!linetarget)
  slope = P_AimLineAttack(player->mo, angle, range, 0);
```

XaserAcheron: Heh, welp -- well, I took a pass at de-killoughizing it and got the same result as you, so let's go with that. :P -- thanks for pre-verifying! If that's a thing. :P

PR #41, opened 2021-05-07
MBF21 Skillsaw Codepointers

XaserAcheron:
Adds Skillsaw's suggested codepointers (plus a few variants):
- A_NoiseAlert
- A_WeaponAlert (the above, for weapons)
- A_HealChase (generic A_VileChase)
- A_JumpIfHealthBelow
- A_JumpIfTargetInSight
- A_JumpIfTargetCloser
- A_JumpIfFlagsSet
- A_AddFlags
- A_RemoveFlags

re: Flags pointers, I decided the best way to handle the MF2 flags is to just not (for MBF21 anyway) :P -- seems like a port-support minefield since half the flags are Heretic things rather than "MBF21-spec" stuff, and the most useful flags are in the standard actor flags range anyway, IMO. Better to do it well later than do it poorly now.

Doc update coming soonish -- just wanted to get the code out there for reviewin'.


prboom2/src/p_enemy.c
```diff
@@ -3198,6 +3210,153 @@ void A_RadiusDamage(mobj_t *actor)
...
+  int state;
+  unsigned int flags;
+
+  if (!mbf21 || !actor)
+    return;
+
+  state = actor->state->args[0];
+  flags = actor->state->args[1];
+
+  if (actor->flags & flags)
```
kraflab: Should this be `(actor->flags & flags) == flags`?

XaserAcheron: Good catch -- is fixed now.

XaserAcheron:
Updated with a new MBF21 flags arg n' some fixins.

spec/support/utility.rb
```diff
@@ -2,7 +2,7 @@ module Utility
   extend self
 
   def play_demo(lmp:, iwad: "DOOM2.WAD", pwad: nil, extra: nil)
-    command = "./build/dsda-doom.exe -iwad spec/support/wads/#{iwad}"
+    command = "./build/Debug/dsda-doom.exe -iwad spec/support/wads/#{iwad}"
```
kraflab: Why is this change here? :^)

prboom2/.editorconfig
```diff
@@ -0,0 +1,10 @@
+# top-most EditorConfig file
```
kraflab: Not sure if this is intended to be here, but if it is then let's make that a separate PR so it doesn't add noise for anyone trying to see just the relevant changes.

prboom2/src/p_enemy.c
```diff
@@ -3198,6 +3210,162 @@ void A_RadiusDamage(mobj_t *actor)
...
+  unsigned int flags, flags2;
+
+  if (!mbf21 || !actor)
+    return;
+
+  state  = actor->state->args[0];
+  flags  = actor->state->args[1];
+  flags2 = actor->state->args[2];
+
+  if ((flags == 0 || (actor->flags & flags) == flags) &&
```
kraflab: I don't think we need these extra `== 0` checks.

kraflab (commented):
Besides the small comments, I think it's ready for docs

XaserAcheron:
WTF, I don't know how those two other files wound up in this commit. o_O

Re-pushed, with a better fix this time 'round.

kraflab:
All good 👍

XaserAcheron:
Just updated the docs, so hopefully this is all kosher-ish now. (Provided my half-asleep brain didn't make any obvious errors, heh :P )

kraflab (approved):

PR #43, opened 2021-05-21
MBF21 Seeker missile codepointers

XaserAcheron:
Adds the following MBF21 codepointers:
- A_SeekTracer(threshold, maxturnangle)
- A_FindTracer(fov, distance)
- A_ClearTracer
- A_JumpIfTracerInSight(state)
- A_JumpIfTracerCloser(state, distance)

More deets in mbf21.md as usual. Been a lot of fun testing this one. :P

### DIFF
--- a/prboom2/doc/mbf21.md
+++ b/prboom2/doc/mbf21.md
@@ -306,9 +306,10 @@ MBF21 defaults:
     - This function will no-op if the calling actor already has a tracer. To forcibly re-acquire a new tracer, call A_ClearTracer first.
     - This function uses a variant of Hexen's blockmap search algorithm (P_RoughMonsterSearch); refer to the implementation for specifics, but it's identical to Hexen's except for the rules for picking a valid target (in order of evaluation):
       - Actors without the SHOOTABLE flag are skipped
-      - Players are skipped in co-op multiplayer games
       - The projectile's owner (target) is skipped
-      - Friendly monsters will be skipped if missile is fired by a player or another friendly monster
+      - Friendly actors (i.e. co-op players and friendly monsters) will be skipped if missile is fired by a friendly actor
+      - Likewise, non-friendly actors will be skipped if missile is fired by a non-friendly actor
+        - Exception: candidate actor will not be skipped if missile's owner is targeting it (i.e. infighting)
       - If the `fov` arg is nonzero, actors outside of an `fov`-degree cone, relative to the missile's angle, are skipped
       - Actors not in line-of-sight of the missile are skipped
 
--- a/prboom2/doc/mbf21.md
+++ b/prboom2/doc/mbf21.md
@@ -207,7 +207,7 @@ MBF21 defaults:
 - For future-proofing, if more nonzero args are defined on a state than its action pointer expects (e.g. defining Args3 on a state that uses A_WeaponSound), an error will be thrown on startup.
 
 #### New DEHACKED Codepointers
-- [PR](https://github.com/kraflab/dsda-doom/pull/20), [PR](https://github.com/kraflab/dsda-doom/pull/38), [PR](https://github.com/kraflab/dsda-doom/pull/40), [PR](https://github.com/kraflab/dsda-doom/pull/41)
+- [PR](https://github.com/kraflab/dsda-doom/pull/20), [PR](https://github.com/kraflab/dsda-doom/pull/38), [PR](https://github.com/kraflab/dsda-doom/pull/40), [PR](https://github.com/kraflab/dsda-doom/pull/41), [PR](https://github.com/kraflab/dsda-doom/pull/43)
 - All new MBF21 pointers use the new "Args" fields for params, rather than misc1/misc2 fields
 - Arg fields are listed in order in the docs below, e.g. for `A_SpawnObject`, `type` is Args1, `angle` is Args2, etc.
 - Although all args are integers internally, there are effectively the following types of args:
@@ -296,7 +296,7 @@ MBF21 defaults:
       - The actor's `tracer` pointer is used as the seek target, rather than Heretic's `special1` field
       - On the z-axis, the missile will seek towards the vertical centerpoint of the seek target, rather than the bottom (resulting in much friendly behavior when seeking toward enemies on high ledges). Refer to the implementation for details.
 
-- **A_FindTracer(threshold, maxturnangle)**
+- **A_FindTracer(fov, distance)**
   - Searches for a valid tracer (seek target), if the calling actor doesn't already have one. Particularly useful for player missiles.
   - Args:
     - `fov (fixed)`: Field-of-view, relative to calling actor's angle, to search for targets in. If zero, the search will occur in all directions.
--- a/prboom2/doc/mbf21.md
+++ b/prboom2/doc/mbf21.md
@@ -285,6 +285,37 @@ MBF21 defaults:
     - `state (uint)`: State to jump to on the calling actor when resurrecting a corpse
     - `sound (uint)`: Sound to play when resurrecting a corpse
 
+- **A_SeekTracer(threshold, maxturnangle)**
+  - Generic seeker missile function.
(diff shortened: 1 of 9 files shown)


prboom2/src/p_maputl.c
```diff
@@ -725,6 +725,195 @@ dboolean P_PathTraverse(fixed_t x1, fixed_t y1, fixed_t x2, fixed_t y2,
...
+  return NULL;
+}
+
+//
+// P_RoughMonsterSearch
+// Searches though the surrounding mapblocks for monsters/players
+//
+// distance is in MAPBLOCKUNITS
+
+mobj_t *P_RoughMonsterSearch(mobj_t *mo, angle_t fov, int distance)
```
kraflab: Since this isn't searching for just monsters, let's call it `P_RoughTargetSearch` for consistency with `P_LookForTargets`.

prboom2/src/p_maputl.c
```diff
@@ -725,6 +725,195 @@ dboolean P_PathTraverse(fixed_t x1, fixed_t y1, fixed_t x2, fixed_t y2,
...
+  {
+    // skip non-shootable actors
+    if (!(link->flags & MF_SHOOTABLE))
+    {
+      link = link->bnext;
+      continue;
+    }
+
+    // skip other players in co-op
+    if (netgame && !deathmatch && link->player)
```
kraflab: Isn't this redundant with the friend check? "Does the player have MF_FRIEND assigned in deathmatch?" is a question I don't know the answer to btw, which is likely important for the check below.

prboom2/src/p_maputl.c
```diff
@@ -725,6 +725,195 @@ dboolean P_PathTraverse(fixed_t x1, fixed_t y1, fixed_t x2, fixed_t y2,
...
+
+    // skip the projectile's owner
+    if (link == mo->target)
+    {
+      link = link->bnext;
+      continue;
+    }
+
+    // skip friendly actors if owner is friendly (or a player)
+    if ((link->flags & MF_FRIEND) && mo->target != NULL &&
```
kraflab: Players have MF_FRIEND set. I'd say we shouldn't have an extra player check unless we need it. Someone may want to have a mechanism that shuts off this flag on the player, and then they would want this condition to change as well.

`if (mo->target && (link->flags & mo->target->flags & MF_FRIEND))`

If this should also exclude non-friends targeting non-friends:

`if (mo->target && !((link->flags ^ mo->target->flags) & MF_FRIEND)))`

prboom2/doc/mbf21.md
```diff
@@ -309,6 +340,22 @@ MBF21 defaults:
     - This function uses the same approximate distance check as other functions in the game (P_AproxDistance).
     - The jump will only occur if distance is BELOW the given value -- e.g. `A_JumpIfTargetCloser(420, 69)` will jump to state 420 if distance is 68 or lower.
 
+- **A_JumpIfTracerInSight(state)**
+  - Jumps to a state if caller's tracer (seek target) is in line-of-sight.
+  - Args:
+    - `state (uint)`: State to jump to
+  - Notes:
+    - This function only considers line-of-sight, i.e. independent of calling actor's facing direction.
```
kraflab, line 349: Should we add an fov arg here as well for completeness?

prboom2/src/p_mobj.c
```diff
@@ -2232,22 +2232,25 @@ void P_ThrustMobj(mobj_t * mo, angle_t angle, fixed_t move)
     mo->momy += FixedMul(move, finesine[angle]);
 }
 
-dboolean P_SeekerMissile(mobj_t * actor, angle_t thresh, angle_t turnMax)
+dboolean P_SeekerMissile(mobj_t * actor, angle_t thresh, angle_t turnMax, dboolean usetracer, dboolean seekcenter)
```
kraflab: How about `mobj_t ** tracer, dboolean seekcenter`?

Then the changes are:
`target = *tracer;`

`*tracer = NULL`

prboom2/doc/mbf21.md
XaserAcheron: May as well, now that we have the code for that written somewhere :P -- I'd like to do that in a separate PR though, if that's OK, since A_JumpIfTargetInSight needs it too (and I gotta fix a crash bug with that one anyway, lol).

prboom2/src/p_maputl.c
XaserAcheron: I'm not seeing anywhere in code offhand that switches MF_FRIEND off in deathmatch... I wonder if that leads to more subtle bugs in places. O_O

prboom2/src/p_maputl.c
XaserAcheron: Thanks! That second one ought to make this function work nicely for monsters, which is kinda niche but someone somewhere will find a use for it I'm sure. :P

I think the first one could handle the deathmatch player case by adding in `&& (!deathmatch || !link->player)` at the end. Presuming my brain's workin' right ATM (never a guarantee with me :P ).

prboom2/src/p_mobj.c
XaserAcheron: That's way better -- I need to stop brute forcing these things. ;P

prboom2/src/p_maputl.c
XaserAcheron: Works for me :)

prboom2/doc/mbf21.md
kraflab: Alrighty

XaserAcheron:
Pushed some updates -- A_FindTracer works for monsters now. :D

prboom2/src/p_maputl.c
```diff
@@ -753,23 +752,28 @@ static mobj_t *RoughBlockCheck(mobj_t *mo, int index, angle_t fov)
...
       continue;
     }
 
-    // skip friendly actors if owner is friendly (or a player)
-    if ((link->flags & MF_FRIEND) && mo->target != NULL &&
-      ((mo->target->flags & MF_FRIEND) || mo->target->player))
+    // likewise, skip non-friendly actors if owner is non-friendly
+    // (unless explicitly targeting another non-friendly, so infighting works)
+    if (mo->target &&
+      !((link->flags ^ mo->target->flags) & MF_FRIEND) &&
```
kraflab, line 775: This check is redundant with the one above (the xor accounts for all combinations). How about removing the separate one above and just adding this:
`&& !(deathmatch && link->player && mo->target->player)`

Other reason: friend monsters can infight, right? Then the friend / deathmatch condition above will skip friends who are infighting despite the second condition aiming to not skip this case.

prboom2/src/p_maputl.c
```diff
@@ -681,48 +681,242 @@ dboolean P_PathTraverse(fixed_t x1, fixed_t y1, fixed_t x2, fixed_t y2,
...
+        else
+          if ((xintercept >> FRACBITS) == mapx)
           {
             xintercept += xstep;
             mapy += mapystep;
           }
+      }
+
+      // go through the sorted list
+      return P_TraverseIntercepts(trav, FRACUNIT);
```
kraflab: 🤔 Why did this method get a bunch of changes? I'm totally fine with format fixes, but this method wasn't involved in this PR, was it?

kraflab: I reverted it

XaserAcheron: WTF? I have absolutely no idea how this happened -- none of this showed up in the diff before I committed it.

I'm probably going to go back and redo this commit and force-push; this mess needs to not exist.

prboom2/src/p_maputl.c
XaserAcheron: Sorry, all these rules make my head spin. It's probably better if you make the edits on this one.

XaserAcheron:
Closing this PR -- going to do another one without the mess

PR #53, opened 2021-06-09
Sound Quirks Fix

Terraformer9x:
Changes an 'or' statement to an 'and' in a check for stopping sound. The 'kill old sound' code is supposed to emulate a quirk with item pickup sounds however it was applying the quirk to all sounds. Changing the `or` to an `and` fixes this.


kraflab:
If the pr+ PR ends up merged, I can cherry-pick it directly, so I'll close this one so we don't need to duplicate anything 🙂

PR #57, opened 2021-07-28
Decouple P_CreateSecNodeList

kraflab:
The extra validcount++ injection into P_CheckPosition done in the scope of heretic and mbf21 fixed an issue introduced by MBF, briefly described here in a comment. Unfortunately, hexen _relies on_ buggy validcount usage. There are recursive calls to P_CheckPosition that cause the counter to become invalidated and lines to get skipped when they shouldn't. This means that a fix for the MBF change breaks hexen. The safest solution here is to completely decouple P_CreateSecNodeList from the normal validcount. This allows validcount to break correctly in hexen, and not break in heretic and mbf21. With luck, there isn't some place in mbf that relies on the count breaking via P_CreateSecNodeList...


kraflab:
More background here: https://github.com/kraflab/dsda-doom/commit/14074455c78ca817d13bd9ae3dd2a953d1af1967

PR #73, opened 2021-08-22
Allow desktop-fullscreen in OpenGL mode

jpernst:
I noticed some odd mode-switch delay when alt-tabbing in the latest builds and tracked it to this; exclusive fullscreen was being forced in opengl mode. Made this quick change and things are working smoothly again. Borderless window mode works perfectly fine with OpenGL and is a common setting in games these days.

### DIFF
--- a/prboom2/src/SDL/i_video.c
+++ b/prboom2/src/SDL/i_video.c
@@ -1154,10 +1154,9 @@ void I_UpdateVideoMode(void)
     init_flags = SDL_WINDOW_OPENGL;
   }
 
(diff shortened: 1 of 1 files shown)


kraflab:
Any idea why `// Fullscreen desktop for software renderer only - DTIED` was written / why it could have been a concern before?

jpernst:
I honestly have no idea. I followed the git-blame back to the port it was merged from, and there's no explanation of why this caveat was made in the original commit.

kraflab (approved):
👍

PR #74, opened 2021-08-26
Ghosts restart

Pedro-Beirao:
Ghosts not restarting was bothering me too much so here's a quick solution. Ive been testing this for awhile and dont see any problem with it.

### DIFF
--- a/prboom2/src/dsda.c
+++ b/prboom2/src/dsda.c
@@ -54,7 +54,7 @@ dboolean dsda_IsWeapon(mobj_t* thing);
 void dsda_DisplayNotification(const char* msg);
 void dsda_ResetMapVariables(void);
 
-int p;
+static int p;
 
 int dsda_getGhostIndex()
 {
--- a/prboom2/src/dsda.c
+++ b/prboom2/src/dsda.c
@@ -54,9 +54,14 @@ dboolean dsda_IsWeapon(mobj_t* thing);
 void dsda_DisplayNotification(const char* msg);
 void dsda_ResetMapVariables(void);
 
-void dsda_ReadCommandLine(void) {
-  int p;
+int p;
 
+int dsda_getGhostIndex()
+{
+  return p;
+}
+
+void dsda_ReadCommandLine(void) {
   dsda_track_pacifist = M_CheckParm("-track_pacifist");
   dsda_track_100k = M_CheckParm("-track_100k");
   dsda_analysis = M_CheckParm("-analysis");
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -81,6 +81,7 @@
 #include "hu_tracers.h"
 #include "dsda.h"
 #include "dsda/settings.h"
+#include "dsda/ghost.h"
 #ifdef GL_DOOM
 #include "gl_struct.h"
 #include "gl_intern.h"
@@ -366,6 +367,8 @@ void G_SkipDemoCheck(void)
 
 int G_ReloadLevel(void)
 {
+  dsda_InitGhostImport(dsda_getGhostIndex());
+
   int result = false;
 
   if ((gamestate == GS_LEVEL) &&
--- a/prboom2/src/dsda.h
+++ b/prboom2/src/dsda.h
@@ -47,4 +47,6 @@ int dsda_MaxKillRequirement(void);
 char* dsda_DemoNameBase(void);
 char* dsda_NewDemoName(void);
 
+int dsda_getGhostIndex();
+
 #endif


prboom2/src/dsda.c
```diff
@@ -54,9 +54,14 @@ dboolean dsda_IsWeapon(mobj_t* thing);
 void dsda_DisplayNotification(const char* msg);
 void dsda_ResetMapVariables(void);
 
-void dsda_ReadCommandLine(void) {
-  int p;
+int p;
```
fabiangreffrath: static?

Pedro-Beirao: Oh yeah that makes sense. done

prboom2/src/e6y.c
```diff
@@ -366,6 +367,8 @@ void G_SkipDemoCheck(void)
 
 int G_ReloadLevel(void)
 {
+  dsda_InitGhostImport(dsda_getGhostIndex());
```
fabiangreffrath, line 370: What if we didn't have an `"-import_ghost"` parameter? Does `dsda_getGhostIndex()` copy well with `0`?

prboom2/src/dsda.c
```diff
@@ -54,9 +54,14 @@ dboolean dsda_IsWeapon(mobj_t* thing);
 void dsda_DisplayNotification(const char* msg);
 void dsda_ResetMapVariables(void);
 
-void dsda_ReadCommandLine(void) {
-  int p;
+static int p;
+
+int dsda_getGhostIndex()
+{
+  return p;
```
kraflab, line 61: What does `p` mean? I know the answer, but the code doesn't tell me. What happens if I call this method? Is it always going to give me the right answer? (No.) What does `Index` mean? My assumption would be the index of a ghost (as there is an index for each ghost), but that's actually not the case. Is `p` relevant to multiple places in this file (does it need to be in this scope)?

You could approach things in this way to improve the messaging in the code and deduplicate things:

```c
int dsda_GetGhostArgumentIndex(void) {
  static int p = -1;

  if (p < 0)
    p = M_CheckParm("-import_ghost");

  return p;
}

...

if ((p = dsda_GetGhostArgumentIndex())) dsda_InitGhostImport(p)
```

This doesn't solve all the problems though. Why does e6y.c need to know anything about how ghosts are imported? I.e., it shouldn't need to look up the option index when it doesn't know anything about ghosts. There are also then 2 places that are fetching this argument and sending it to `dsda_InitGhostImport`. We can isolate the knowledge of the commandline to the importer and avoid this duplication.

That's all about clean implementation, but there's actually a more important problem. Let's look at some of what happens in the importer:

```c
  dsda_ghost_import.ghosts = calloc(dsda_ghost_import.count, sizeof(dsda_ghost_t));

  arg_i = option_i;
  while (++arg_i != myargc && *myargv[arg_i] != '-') {
    dsda_OpenGhostFile(arg_i, &ghost_file);
```

First, we're allocating some data here, but this method assumes it is only called once. Is the data freed somewhere, or will this lead to a memory leak? This method also opens file handles. Are those files closed somewhere? What happens if we open thousands of file streams (not uncommon for someone grinding a short map)? There's a limit to how many file handles you can have simultaneously open, so this will cause a crash for players. We can fix this by closing files and freeing the data, but do we need to? Why reimport data that we've already read and stored?

What we need here is a couple things: cleaning up the implementation of commandline handling (probably extract export / import argument checking to ghost.c) and adding a reset method in ghost.c that will handle restarting things cleanly. I don't know how much effort that is - I haven't done it yet because it's not a quick fix and will take some investigation into how exactly ghost.c is working to do it cleanly.

kraflab (changes requested):

prboom2/src/e6y.c
Pedro-Beirao: It works fine without that parameter or even multiple ghosts.

prboom2/src/dsda.c
Pedro-Beirao: Just wanted a quick solution to this problem that messes the least with the code.
I thought `p` would always give you the right answer. When testing it always worked fine.

e6y.c seemed like a decent file to add this restart, but you should organise it the way you want.

I agree that it shouldn't reimport the data that was already read and stored, but I didn't think about it when doing this.

Would it be possible to change the `dsda_UpdateGhosts()` function of dsda.c to make it rewind the ghost?
This might be a cleaner implementation (But maybe it would be harder to do, idk)

kraflab: Like I said, I don't know what the best approach is because I haven't looked into it :^)
I'm not interested in quick fixes to things like this, because that approach leads to a messy code base that is also prone to bad side effects, like the file handle crash I mentioned 😉 
If you want to try out a more substantial solution, then I will welcome it - otherwise I can take care of this myself soon.

PR #76, opened 2021-09-01
merge pain palette changing for Chex

fabiangreffrath:


prboom2/src/st_stuff.c
```diff
@@ -791,15 +791,26 @@ static void ST_doPaletteStuff(void)
...
   if (cnt)
     {
-      palette = (cnt+7)>>3;
-      if (palette >= NUMREDPALS)
-        palette = NUMREDPALS-1;
+      // In Chex Quest, the player never sees red.  Instead, the
+      // radiation suit palette is used to tint the screen green,
+      // as though the player is being covered in goo by an
+      // attacking flemoid.
+      if (gamemission == chex)
```
kraflab, line 798: We should still have a check if the pain palette is enabled right?

fabiangreffrath: If it wasn't, we wouldn't even enter the `if (cnt)` block.

kraflab: Ah right

kraflab (approved):
Thanks 🥣

PR #79, opened 2021-09-11
Fix for "Cannot open file" issue on macOS

Pedro-Beirao:
macOS has a limit of 256 files open at the same time by default.
With the new crosshair variants introduced in 0.21.3, it needed to open 258 files to compile.

@fabiangreffrath you might want to take look, as prboom+ might end up needing to open that many files


kraflab (approved):
Thanks 🤠

PR #80, opened 2021-09-14
fix respawning with the IDDQD cheat for the Raven games

fabiangreffrath:


prboom2/src/m_cheat.c
```diff
@@ -306,8 +306,11 @@ static void cheat_god()
 
       // spawn a teleport fog
       an = plyr->mo->angle >> ANGLETOFINESHIFT;
-      P_SpawnMobj(plyr->mo->x+20*finecosine[an], plyr->mo->y+20*finesine[an], plyr->mo->z, MT_TFOG);
-      S_StartSound(plyr, sfx_slop);
+      P_SpawnMobj(plyr->mo->x + 20*finecosine[an],
+                  plyr->mo->y + 20*finesine[an],
+                  plyr->mo->z + g_telefog_height,
+                  g_mt_tfog);
+      if (!raven) S_StartSound(plyr, sfx_slop);
```
kraflab: Maybe we can use the teleport sound for heretic / hexen and add something like a `g_sfx_revive` parameter?

kraflab (approved):
Thanks 🤠

PR #87, opened 2021-11-10
Separate automap colors for revealed and unrevealed secrets

jackrjli:
This adds a separate color for revealed secrets on the automap, so that revealed and unrevealed secrets may be distinguished more easily.


kraflab:
Cool, thanks 👌

PR #88, opened 2021-11-10
Fix seg texture clipping

kraflab:
There's a visual bug caused by the floor waggle effect from hexen, where a texture will be drawn much larger than it actually is when its visibility is ~0 as it rises into view above the adjacent sector. This PR fixes the bug by fixing a comparison in r_segs that inadvertently ignores the lowest 4 bits in the world variables. The issue only existed in software mode.

### DIFF
--- a/prboom2/src/r_segs.c
+++ b/prboom2/src/r_segs.c
@@ -937,32 +937,41 @@ void R_StoreWallRange(const int start, const int stop)
         markceiling = false;
     }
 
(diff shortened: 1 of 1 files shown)


kraflab:
@fabiangreffrath this might be worth fixing in crispy hexen, although I know you aren't necessarily maintaining that part of the trilogy. Consider it an FYI :^)

PR #94, opened 2022-01-12
Manage deh thing indices

kraflab:
The current dsdhacked implementation allocates _up to_ the required index. Adding 1 new thing could require billions of things worth of space, if the dehacked file uses a large index. This PR adds dehacked thing index management so that values in the dehacked file are translated into the smallest available indices.

### DIFF
--- a/prboom2/src/d_deh.c
+++ b/prboom2/src/d_deh.c
@@ -1418,9 +1418,22 @@ typedef struct {
   actionf_t cptr;  // actual pointer to the subroutine
   const char *lookup;  // mnemonic lookup string to be specified in BEX
   // CPhipps - const*
-  int argcount;  // [XA] number of mbf21 args this action uses, if any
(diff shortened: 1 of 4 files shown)


kraflab:
@rfomin FYI. This is something that should eventually be done for all the dsdhacked entities. I'm doing it for things right now because I'll need to have some control over dehacked thing indices when I start adding zdoom things to dsda-doom (need to make sure a dehacked file that thinks something is "new" doesn't overwrite an existing thing). I'm not sure when I'll do the other entities because it isn't strictly necessary for me at this moment - and you need to make sure EVERY instance of a given entity is captured.

If you notice something I didn't account for here, please let me know 🙏

rfomin:
LGTM, but I haven't tested it. Are you planning to add ZDoom codepointers?

kraflab:
> Are you planning to add ZDoom codepointers?

No plans for that, but map things like MapSpot and the extra teleport destinations for instance need to be added.

PR #98, opened 2022-01-23
Completely initialize blocklinks in P_UnArchiveBlockLinks

janchochol:
When loading save game / restoring keyframe on maps with lot's of
monsters and friction floor, it can happen that Doom is blocked on
indefinite loop.

Problem is loop in 'blocklinks' linked list. It is cased by fact, that
'blocklinks' is initialized 2 times (in 'P_SetThingPosition' and in
'P_TrueUnArchiveThinkers', both called from 'P_TrueUnArchiveThinkers')
and these two sources can be unsynchronized (in case of friction floors,
'P_Move' updates coordinates "after" calling 'P_TryMove', so it is
possible that coordinates does not correspond to position in
'blocklist').

### DIFF
--- a/prboom2/src/p_saveg.c
+++ b/prboom2/src/p_saveg.c
@@ -620,6 +620,22 @@ void P_UnArchiveBlockLinks(mobj_t** mobj_p, int mobj_count)
     memcpy(&count, save_p, sizeof(count));
     save_p += sizeof(count);
 
+    // Initialize 'blocklinks[i]' to empty value.
+    //
+    // We are restoring whole blocklist snapshot, so there is no need to keep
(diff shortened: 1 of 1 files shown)


janchochol:
Problem can be easily reproduced e.g. on NoSp3 map 28 with fight on friction floor.

kraflab:
Aha, the problem occurs when `count` is zero, because then the start of the chain is never updated. Nice catch :^)

janchochol:
> the problem occurs when `count` is zero

Yeah, I was not thinking about it that way. In that case my comment was really useless, as it was meant to be completely rewritten (and it really was if `count > 0`).

PR #99, opened 2022-01-23
Remove the old prboom macOS launcher

Pedro-Beirao:
Besides removing the old prboom macOS launcher, this PR also fixes some #ifdef statements related to mac.
`#ifdef MACOSX` does not work and should be changed to `#ifdef __APPLE__`

The file `i_system.m` tries to change some paths, but for some reason it is never called and is not even present on `src/CMakeLists.txt`, so I think its better to keep the Linux paths, which is the current behaviour 

The `KMOD_META` key is not the macOS Command key, in fact, it isn't even listed in the documentation (https://wiki.libsdl.org/SDL_Keymod), so I changed it to `KMOD_GUI` which are the right and left Command keys


kraflab:
`-20k` 😮 

Beautiful, thanks!

PR #100, opened 2022-02-07
Enter cheats on the console

Pedro-Beirao:


prboom2/src/dsda/console.c
```diff
@@ -254,6 +260,13 @@ static void dsda_ExecuteConsole(void) {
         break;
       }
     }
+
+    char prefix[2];
```
kraflab: In C, variables need to be declared at the top of the scope. Doing it in the middle of a scope is a c++ feature (though some compilers will let it slide).

prboom2/src/dsda/console.c
```diff
@@ -254,6 +260,13 @@ static void dsda_ExecuteConsole(void) {
...
+    memcpy(prefix, command, 2);
+    if(!stricmp(prefix, "id"))
```
kraflab: There shouldn't be any exceptions in the console, so we don't want to have an extra specific comparison here. It adds extra complexity that is avoidable. Rather we should follow the existing behaviour. I'd suggest adding a new command "cheat.enter" with the arg for the cheat. Define it so that it automatically covers all current and future cheat codes.

I should be able to enter for example "cheat.enter idclev23" or "cheat.enter tntem"

kraflab: You can follow the pattern of console_CommandLock (with console_CheatEnter).

Pedro-Beirao: Entering "cheat.enter" takes some time, that's why I wanted to avoid it.

"cheat.enter idclev xx" does not work because this cheat has the "not_menu" flag

kraflab: If you're worried about time, why would you open the console? What's the problem we're trying to solve here? There may be a different solution than using the console.

prboom2/src/dsda/console.c
```diff
@@ -213,6 +215,18 @@ static dboolean console_Exit(const char* args) {
   return true;
 }
 
+static dboolean console_CheatEnter(const char* args) {
+  char element[CONSOLE_ENTRY_SIZE];
+  char value[3];
```
kraflab: What happens if the string size of value is greater than 2? You can protect against this by using `%2s` in `sscanf` to restrict the length. It isn't used like this elsewhere because the code that fills the console buffer already restricts the total to `CONSOLE_ENTRY_SIZE`.

prboom2/src/dsda/console.c
Pedro-Beirao: Typing cheats normally makes me press some keys that move the player, rewind or even restart the level.
I know, that I can bind a certain key to a certain cheat, but sometimes I want to use a cheat that I have no keybind for.
The console seemed like the best solution for this problem.

prboom2/src/dsda/console.c
```diff
@@ -26,6 +26,8 @@
 
 #include "console.h"
 
+#include "m_cheat.h"
```
kraflab: Could you put this above `#include "m_menu.h"`? I try to organize the header includes in order by group: doom / prboom+ files first, then dsda files, then the current file.

prboom2/src/m_cheat.c
```diff
@@ -999,6 +999,21 @@ dboolean M_CheatResponder(event_t *ev)
   return false;
 }
 
+dboolean M_CheatEntered(char element[], char value[3])
+{
+  cheatseq_t* cheat_i;
+
+  for (cheat_i = cheat; cheat_i->cheat; cheat_i++)
+  {
+    if (!strcmp(cheat_i->cheat, element) && M_CheatAllowed(cheat_i->when))
```
kraflab, line 1008: So this will work for everything _except_ idclev, is that correct? Any idea why this cheat is disabled in the menu? Maybe we can address that later on.

prboom2/src/dsda/console.c
kraflab: Hm, one alternative would be a menu that shows all the cheats and their states active / inactive and you can toggle them. Or there could be a specific key to prompt the user to enter a cheat (that blocks other responders from seeing the inputs). Would either of those be useful for you? Adding the stuff to the console like this is something I wanted to do anyway in the long run, so it's still worth merging, but I may consider one of those things for the future.

prboom2/src/m_cheat.c
Pedro-Beirao: I removed the "not_menu" flag and everything worked correctly, but when using idclev on the console it would warp to the level, but not update the display until the console was exited.
So this cheat was probably disabled to avoid this visual problem.

prboom2/src/dsda/console.c
```diff
@@ -213,6 +214,18 @@ static dboolean console_Exit(const char* args) {
   return true;
 }
 
+static dboolean console_CheatEnter(const char* args) {
+  char element[CONSOLE_ENTRY_SIZE];
+  char value[3];
+
+  if (sscanf(args, "%s %2s", element, value) > 0)
```
kraflab: If the string only has `element` and not `value`, then `value` is uninitialized. It will have random data in it that will then get sent to `M_CheatEntered`. If someone enters a cheat and forgets to add the `value` where it is needed, that may cause a problem. But we don't know in this scope whether or not something _should_ have a `value` or not, which is a bit unpleasant 🤔 

Maybe it is better in the end to explicitly make the cheats direct commands like `iddqd` instead of `cheat.enter iddqd` so that each one can validate the arguments.

Are there specific ones that you want? You could just add a few as a proof of concept and then I will add the others later.

Pedro-Beirao: That will make console.c pretty big pretty quickly if there are different functions for each cheat.
But sure I can do that 👍

prboom2/src/m_cheat.c
fabiangreffrath: > So this will work for everything _except_ idclev, is that correct? Any idea why this cheat is disabled in the menu? Maybe we can address that later on.

Check the 5th bullet point:
https://doomwiki.org/wiki/IDCLEV_idiosyncrasies

prboom2/src/dsda/console.c
Pedro-Beirao: A key to prompt the user to input a cheat would be perfect.

prboom2/src/dsda/console.c
kraflab: That's alright, it's expected the console command list will eventually be large - I will split things into topic files when it becomes necessary.

Pedro-Beirao:
Before I start that, what do you think about this https://github.com/Pedro-Beirao/dsda-doom/commit/e50bcf0c307e75a11ab16a1ed8be304b3bc40e6f ?
Would there still be problems with `value` having random data?

prboom2/src/dsda/console.c
Pedro-Beirao: Before I start that, what do you think about this https://github.com/Pedro-Beirao/dsda-doom/commit/e50bcf0c307e75a11ab16a1ed8be304b3bc40e6f ?
Would there still be problems with `value` having random data?

kraflab: There won't be random data if it's interpreted as a string, because of the line `if (scan_count == 1) args[0] = '\0';`

However, we should still verify the data. So for example (gimme takes 2 characters):

```c
static dboolean console_Gimme(const char* args) {
  char value[3];

  if (sscanf(args, "%2s", value) == 1)
    return M_CheatEnter("gimme", value);

  return false;
}

...

{ "gimme", console_Gimme } // in the command list
```

This might be begging for a refactor so that the `console_*` functions get sent the command along with the args, and then we can just have 3 methods: `console_CheatNoArgs`, `console_CheatOneArg` and `console_CheatTwoArgs`. E.g., `{ "iddqd", console_CheatNoArgs }`. If you're only planning on adding a few, then don't worry about the refactor, I will do that before adding all the cheats.

Pedro-Beirao: In the latest commit I added the `console_CheatNoArgs`, `console_CheatOneArg` and `console_CheatTwoArgs`.
`tntweap` does not work because, depending on whether it is given a argument, it should call `cheat_tntweap` or `cheat_tntweapx`

kraflab: That's just a hack to prompt the user for a number when tntweap is entered (the "cheat" function is just printing a notification). It's the same as with `idbehold`. The console should work with `tntweap 4` or `idbehold h` I think for consistency of the `base_cheat cheat_args` pattern (and do nothing with `idbeholdh`). Right now I'm thinking we should just do the basic ones that have no arguments using `console_BasicCheat` and leave idbehold / tntweap / idmus / etc for later (and thus remove the one arg / two arg stuff). I want to try out some options there and may refactor everything, but there would be a good baseline in place here. What do you think?

Pedro-Beirao: Yeah that would be much better 👍

prboom2/src/m_cheat.c
```diff
@@ -999,6 +999,24 @@ dboolean M_CheatResponder(event_t *ev)
...
 
+dboolean M_CheatEntered(const char* element, const char* value)
+{
+  cheatseq_t* cheat_i;
+
+  for (cheat_i = cheat; cheat_i->cheat; cheat_i++)
+  {
+    if (!strcmp(cheat_i->cheat, element) && M_CheatAllowed(cheat_i->when))
+    {    
+      if (cheat_i->arg > 0)
```
kraflab: This should be `>=` right?

Pedro-Beirao: You are right. Didn't see that the arguments like pw_invulnerability could be 0.

kraflab (approved):
Nice work 👍

PR #101, opened 2022-02-20
implement SNDFONT lump

rfomin:


prboom2/src/MUSIC/flplayer.c
```diff
@@ -105,9 +107,51 @@ static const char *fl_name (void)
 #include <delayimp.h>
 #endif
 
+static void *fl_sfopen(const char *lumpname)
```
kraflab, line 110: Let's pass lumpnum instead of lumpname here and save a lookup.

prboom2/src/memio.c
```diff
@@ -198,7 +198,7 @@ int mem_fseek(MEMFILE *stream, signed long position, mem_rel_t whence)
       return -1;
   }
 
-  if (newpos < stream->buflen)
+  if (newpos <= stream->buflen)
```
kraflab, line 201: Could you explain this change / what consequences there might be?

prboom2/src/MUSIC/flplayer.c
rfomin: This is a callback, so we need to pass `const char *`. We can save an int to a string, or convert pointers, or make `lumpnum` global, but I don't think it's worth it.

prboom2/src/memio.c
rfomin: Without this change SEEK_END file will never work. Same fix in Odamex: https://github.com/odamex/odamex/blob/7190b0217abfd7fb22ad6f9f10026324367e8f3a/common/m_memio.cpp#L200

prboom2/src/MUSIC/flplayer.c
kraflab: Ah ok, then all good 👍

kraflab (approved):
Thanks 🙌

PR #102, opened 2022-03-03
Allow compilation of universal MacOS binaries

Pedro-Beirao:
On MacOS, to compile universal (Intel + Silicon) binaries, we need to give the `-DCMAKE_OSX_ARCHITECTURES="x86_64;arm64"`  argument to cmake.
Because we are targeting 2 architectures, the `foreach` loop is needed.


kraflab (approved):
👍

PR #103, opened 2022-03-31
Update math for mouse stutter correction

mikeday0:
FRACUNIT / 2 should instead be 1000000 / GAMETIC, i.e. the number of microseconds per gametic. The math error probably isn't noticeable, but it's nice when code matches intention. :smile:


kraflab (approved):
This makes a bit more sense 😉

PR #104, opened 2022-04-12
Cleanup DSDA Time's MSVC compatibility layer

Lovey847:
Compiled & tested in MSVC x86, MSVC x64, MinGW x64, GCC x64 and Clang x64

This pull request removes the MSVC hack in prboom2/src/dsda/time.c and adds a cleaner win32 interface to replace it.

It's not a huge pull request, just something I've wanted to do for a while now.

### DIFF
--- a/prboom2/src/dsda/time.c
+++ b/prboom2/src/dsda/time.c
@@ -15,58 +15,37 @@
 //	DSDA Time
 //
 
+#ifdef _WIN32
+#include <windows.h>
+#else
 #include <time.h>
+#endif
 
 #include "i_system.h"
 
 #include "time.h"
 
-// [XA] MSVC hack: need to implement a handful of functions here.
-// this was yoinked from https://stackoverflow.com/a/31335254 so
-// no guarantees on if this stuff is accurate/kosher/whatever,
-// so maybe don't cut an official release with MSVC just yet.
-// it's enough to hack it into compiling for me, at least. ;)
+// lovey: Win32 interface
+#ifdef _WIN32
 
-#ifdef _MSC_VER
+static unsigned long long dsda_time[DSDA_TIMER_COUNT];
 
-#include <windows.h>
-
-#define CLOCK_MONOTONIC -1
-#define exp7           10000000i64 //1E+7
-#define exp9         1000000000i64 //1E+9
-#define w2ux 116444736000000000i64 //1.jan1601 to 1.jan1970
-
-void unix_time(struct timespec *spec) {
-  __int64 wintime; GetSystemTimeAsFileTime((FILETIME*)&wintime);
-  wintime -= w2ux;  spec->tv_sec = wintime / exp7;
-  spec->tv_nsec = wintime % exp7 * 100;
+void dsda_StartTimer(int timer) {
+  QueryPerformanceCounter((LARGE_INTEGER*)&dsda_time[timer]);
 }
 
-int clock_gettime(int _skip, struct timespec *spec)
-{
-  static struct timespec startspec;
-  static double ticks2nano;
-  static __int64 startticks, tps = 0;
-  __int64 tmp, curticks;
+unsigned long long dsda_ElapsedTime(int timer) {
+  static unsigned long long freq = 0;
+  unsigned long long now;
 
-  QueryPerformanceFrequency((LARGE_INTEGER*)&tmp);
+  if (!freq) QueryPerformanceFrequency((LARGE_INTEGER*)&freq);
 
-  if (tps != tmp) {
-    tps = tmp;
-    QueryPerformanceCounter((LARGE_INTEGER*)&startticks);
-    unix_time(&startspec); ticks2nano = (double)exp9 / tps;
-  }
-
-  QueryPerformanceCounter((LARGE_INTEGER*)&curticks); curticks -= startticks;
-  spec->tv_sec = startspec.tv_sec + (curticks / tps);
-  spec->tv_nsec = startspec.tv_nsec + (double)(curticks % tps) * ticks2nano;
-  if (!(spec->tv_nsec < exp9)) { spec->tv_sec++; spec->tv_nsec -= exp9; }
-  return 0;
+  QueryPerformanceCounter((LARGE_INTEGER*)&now);
+  return (now-dsda_time[timer])*1000000ll/freq;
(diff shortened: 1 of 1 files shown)


prboom2/src/dsda/time.c
```diff
@@ -15,58 +15,37 @@
...
-// this was yoinked from https://stackoverflow.com/a/31335254 so
-// no guarantees on if this stuff is accurate/kosher/whatever,
-// so maybe don't cut an official release with MSVC just yet.
-// it's enough to hack it into compiling for me, at least. ;)
-
-#ifdef _MSC_VER
-
-#include <windows.h>
+// lovey: Win32 interface
+#ifdef _WIN32
```
kraflab: Why `_WIN32` instead of `_MSC_VER`? In msys2 the unix code works and I'd like to keep it that way. What does _WIN32 evaluate to in that context?

prboom2/src/dsda/time.c
```diff
@@ -15,58 +15,37 @@
...
+unsigned long long dsda_ElapsedTime(int timer) {
+  static unsigned long long freq = 0;
+  unsigned long long now;
 
-  if (tps != tmp) {
-    tps = tmp;
-    QueryPerformanceCounter((LARGE_INTEGER*)&startticks);
-    unix_time(&startspec); ticks2nano = (double)exp9 / tps;
-  }
+  if (!freq) QueryPerformanceFrequency((LARGE_INTEGER*)&freq);
```
kraflab: Please follow the style elsewhere that adds spaces in between operators, casts, etc. (I realize the code you are replacing wasn't matching either.)

prboom2/src/dsda/time.c
```diff
@@ -15,58 +15,37 @@
...
 
-  QueryPerformanceCounter((LARGE_INTEGER*)&curticks); curticks -= startticks;
-  spec->tv_sec = startspec.tv_sec + (curticks / tps);
-  spec->tv_nsec = startspec.tv_nsec + (double)(curticks % tps) * ticks2nano;
-  if (!(spec->tv_nsec < exp9)) { spec->tv_sec++; spec->tv_nsec -= exp9; }
-  return 0;
+  QueryPerformanceCounter((LARGE_INTEGER*)&now);
+  return (now-dsda_time[timer])*1000000ll/freq;
```
kraflab: Where does this constant come from? Can you give it a name?

kraflab (commented):
Could you explain what this PR *does* and what makes it preferable? Having `#else` here is a code smell to me. The existing code provides a consistent interface and simply abstracts the missing method, which leads to identical code on all platforms :^)

Lovey847:
I don't exactly know in what way you think the `#else` is bad, but my reasoning for not including it on windows is, well, you don't really need time.h.

I gave a very brief description of what the PR does when I made it so I'll give a more descriptive version this time:
This PR aims to remove the copy-pasted `clock_gettime` implementation that xaser added to support MSVC at the time of their contribution, because it was hard to read, and was inefficient.

I chose to replace it with a more readable & efficient win32 implementation that would also work in mingw, there's only a couple of routines that need to be replaced so it ended up being quite small, and shouldn't be too hard to rewrite in the case of having to rewrite the entire timing library, if you ever need to do that for some reason.

Here is a reminder that I allowed edits from maintainers, don't exactly know what that means but I assume it allows you to modify the formatting and whatnot, if you need to.

prboom2/src/dsda/time.c
Lovey847: The unix code doesn't specify what it means in this context either, so I didn't think it mattered.

It's converting the "counts" (what you get from `QueryPerformanceCounter`) to microseconds.

The frequency of these "counts" is queried from `QueryPerformanceFrequency`, which is what freq is filled with.

prboom2/src/dsda/time.c
Lovey847: Sorry, I wasn't really thinking of code formatting when I made this, but now that I think about it it's a big deal.

I'll take this as a learning experience

prboom2/src/dsda/time.c
Lovey847: `_WIN32` because this is more efficient on windows, in both mingw and msvc.

If you look in the MinGW source ([https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v10.0.0.zip/download](url), at mingw-w64-libraries/winpthreads/src/clock.c), `CLOCK_MONOTONIC` in `clock_gettime` is implemented with `QueryPerformanceFrequency` & `QueryPerformanceCounter`

Again, this was tested with MinGW x64, that was one of the first things I tested to make sure I didn't mess anything up.

kraflab:
To explain a bit more why this approach is "less clean" to me, the hack from stack overflow externalizes the compatibility with windows from the business logic. Since it's abstracted away, it isn't part of the actual logic and thus it becomes more or less irrelevant as long as it is functional. We can move it into a separate file and it is simply a compatibility patch.

If we want to change how the patch works at any time, we can do so without changing the business logic. At the same time, if we want to change the business logic, we can do so without touching the compatibility patch. This is the effect of properly separated concerns and keeping code agnostic.

In contrast, the solution in this PR has different methods for different operating systems, so changing one may require changing the other. This is the code smell.

If your main concern is efficiency, you're going to need to supply detailed data on how the efficiency is affected by this PR, and what impact it has on actual gameplay fps. I suspect it will be very difficult to demonstrate accurately.

Lovey847:
I could argue that, in my original PR, I moved the compatibility layer from being `clock_gettime` to `dsda_StartTimer` and `dsda_ElapsedTime,` but there's really no point in that seeing how easy it is to just replace `clock_gettime`.

I would've liked to keep it `_WIN32` instead of `_MSC_VER`, but I don't know a clean way of preventing the conflict between my `clock_gettime` and time.h's `clock_gettime`.

EDIT: Actually, I think I see the issue with my line of thinking. I'm so used to `clock_gettime` being used in unix-only scenarios because it's a part of posix, so my head just sectioned it into it's own layer. But you seem to be using it as if it will always be present, which... it will, since you have compatibility layers for platforms that don't have it.

kraflab (approved):
Now that looks squeaky clean 😉 🎉 
Thanks!

PR #105, opened 2022-04-24
GL: Non-exclusive fullscreen render-to-texture

elim2g:
When rendering in OpenGL using non-exclusive fullscreen mode on a resolution not matching the window, we now render the game to a texture and stretch that (by its aspect ratio) to fill at least one dimension of the screen. This will stretch and center the scene within the window.


prboom2/src/d_main.c
```diff
@@ -322,6 +322,9 @@ void D_Display (fixed_t frac)
     oldgamestate = -1;            // force background redraw
   }
 
+  if (V_IsOpenGLMode() && !exclusive_fullscreen && !dsda_SkipMode())
```
kraflab: Is skip mode in particular relevant here? Does `!nodrawers` satisfy the condition?

elim2g: Most likely. I didn't look too far into it, I was just making sure I wouldn't step on SkipMode's toes by performing glClear on framebuffers that weren't going to have any work done to them.

kraflab (approved):
Magnifique

PR #107, opened 2022-04-27
Software: Allow taking screenshots at game scene resolution

elim2g:
Added a menu option for taking screenshots at the game's rendered resolution. By default this option uses the pre-existing behaviour of taking the screenshot at the full resolution of the window.


kraflab:
As discussed "offline" I think this isn't worth adding extra complexity for now, closing 🚪

PR #108, opened 2022-05-05
remove armsbg binicon and corresponding typedef, not actually used anywhere

fabiangreffrath:
See https://github.com/fabiangreffrath/woof/commit/c27abe130028e909c100934a9e684f0217b5da8c


kraflab (approved):
Thanks 😎

PR #109, opened 2022-05-19
off64_t is not defined on macOS

Pedro-Beirao:
Instead, we should use off_t.
I don't understand why it this only happens when using a newer version of Dumb (It compiled fine using Dumb 0.9.3)

### DIFF
--- a/prboom2/src/MUSIC/dumbplayer.c
+++ b/prboom2/src/MUSIC/dumbplayer.c
@@ -68,9 +68,11 @@ const music_player_t db_player =
 #if !defined(_FILE_OFFSET_BITS) || (_FILE_OFFSET_BITS < 64)
 #ifdef _MSC_VER
 #define DUMB_OFF_T_CUSTOM __int64
-#else /* !_MSC_VER */
+#elif defined(__APPLE__)
+#define DUMB_OFF_T_CUSTOM off_t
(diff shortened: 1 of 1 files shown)


kraflab (approved):
Thanks! 😎

kraflab:
@fabiangreffrath this might affect prboom+ at some point as well FYI (started getting random failures in the CI yesterday)

Pedro-Beirao:
To clarify the problem, a few days ago, the Dumb library was updated on the brew package manager from version 0.9.3 to 2.0.3.

This newer version states: 
```
/*
 * If the build fails here, it does so, because we need a 64-bit-type for
 * defining offsets. To fix this do either of the following:
 *
 * 1. Compile your code with -D_FILE_OFFSET_BITS=64, so that off_t is 64-bit
 *    (recommended, but make sure the rest of your code can handle it)
 * 2. Supply your own definition of a signed 64-bit integer
 *    such as off64_t or int64_t before including dumb.h as follows:
 *    #define DUMB_OFF_T_CUSTOM int64_t
 */
```

On macOS, off_t is always 64-bit and there is no off64_t.

Discussion about this problem: https://trac-test.macports.org/ticket/24067?cversion=0&cnum_hist=35

PR #110, opened 2022-05-19
Implement support for 8-bit and 16-bit WAV files

rfomin:
This is based on Woof. Alternatively there is PrBoom+ PR: https://github.com/coelckers/prboom-plus/pull/42/files

### DIFF
--- a/prboom2/src/SDL/i_sound.c
+++ b/prboom2/src/SDL/i_sound.c
@@ -627,8 +627,8 @@ static void I_UpdateSound(void *unused, Uint8 *stream, int len)
         }
         else
         {
-          s = (ci->data[0] * (0x10000 - ci->stepremainder))
-            + (ci->data[1] * (ci->stepremainder))
+          s = ((unsigned int)ci->data[0] * (0x10000 - ci->stepremainder))
+            + ((unsigned int)ci->data[1] * (ci->stepremainder))
             - 0x800000; // convert to signed
         }
 
--- a/prboom2/src/SDL/i_sound.c
+++ b/prboom2/src/SDL/i_sound.c
@@ -109,6 +109,7 @@ typedef struct
   // ... and a 0.16 bit remainder of last step.
   unsigned int stepremainder;
   unsigned int samplerate;
+  unsigned int bits;
   // The channel data pointers, start and end.
   const unsigned char *data;
   const unsigned char *enddata;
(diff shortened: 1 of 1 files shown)


prboom2/src/SDL/i_sound.c
```diff
@@ -163,21 +164,70 @@ static void stopchan(int i)
...
+  if (len > 44 && memcmp(data, "RIFF", 4) == 0 && memcmp(data + 8, "WAVEfmt ", 8) == 0)
+  {
+    SDL_RWops *RWops;
+    SDL_AudioSpec wav_spec;
+    Uint8 *wav_buffer = NULL;
+    int bits, samplelen;
+
+    RWops = SDL_RWFromMem(data, len);
+
+    Z_ChangeTag(data, PU_CACHE);
```
kraflab: What's this line about? All sound lumps in dsda-doom are locked at launch.

rfomin: `SDL_LoadWAV_RW` loads the data into memory and decodes it if necessary, so the original data buffer is no longer used. Woof can use `Z_ChangeTag(data, PU_CACHE)` here, not sure about PrBoom+.
The reason for using SDL is because the WAV format is complex and there are many problematic WAV files (pistol sound in UACrebellion.wad for example).

rfomin: Fixed in https://github.com/kraflab/dsda-doom/pull/110/commits/3fc17ea07dddb481efa3648375848a8585200464

prboom2/src/SDL/i_sound.c
```diff
@@ -163,21 +164,74 @@ static void stopchan(int i)
...
+
+    bits = SDL_AUDIO_BITSIZE(wav_spec.format);
+    if (bits != 8 && bits != 16)
+    {
+      lprintf(LO_WARN, "Only 8 or 16 bit WAV files are supported");
+      SDL_FreeWAV(wav_buffer);
+      return channel;
+    }
+
+    Z_Free(data);
```
kraflab: I'm even more confused now 😅
This method is receiving a piece of data that is a lump. The lump itself is locked and allocated in zone memory. Every time the sound effect is played, that lump data will be sent to this method. Why are we freeing the lump data here?

rfomin: Because if that lump is a WAV file, it might be compressed and needs to be decoded.

kraflab: My question is why are we freeing the raw lump data in this function. It's an unexpected side effect to me (outside this method, the data is locked, but then it gets freed before it returns).

rfomin: > outside this method, the data is locked, but then it gets freed before it returns

But immediately after the free, we allocate a new buffer and copy the decoded WAV into it. BTW, what the difference between PU_STATIC and PU_LOCKED?

kraflab: I understand you are allocating a new buffer, but outside this method we have a pointer to data that now doesn't exist, which is not a safe practice 😅

What I'm telling you is that all sound effect data in dsda-doom is cached when the program launches and does not need to be freed manually at any time. We use lump locking / unlocking to manage lump data (which takes care of cleanup if and when it is needed).

kraflab: You can think of the locking as a way of managing zone memory and swapping the tags based on the number of locks. This is done so one place in the code base can lock a lump, and another can, and they will unlock it when they are done. Then no part needs to know if any other part has locked the data, as the zone is managing the lock counts.

kraflab: If for some reason another part of the code base has locked the sound effect lump, and you manually free it, then that part now has an invalid pointer but doesn't know it.

fabiangreffrath: > What I'm telling you is that all sound effect data in dsda-doom is cached when the program launches and does not need to be freed manually at any time. We use lump locking / unlocking to manage lump data (which takes care of cleanup if and when it is needed).

But in the case at hand the sound *lump* does not contain the actual sound *data*, but some WAV format stuff that first needs to be converted into your actual sound data format. And once this conversion is done, you will never need the sound *lump* again.

rfomin: Right, sorry, Woof does it differently. So is it safe just write `ci->data = wav_buffer`? Or we need to call `SDL_FreeWAV(wav_buffer)` somewhere?

kraflab: > But in the case at hand the sound _lump_ does not contain the actual sound _data_, but some WAV format stuff that first needs to be converted into your actual sound data format. And once this conversion is done, you will never need the sound _lump_ again.

There are two problems here.

This data has been locked by another part of the program. Whether or not that is "correct" or meaningful is besides the point. Freeing data that is reserved somewhere else is a fundamentally dangerous action and relies on assumptions that may or may not be incorrect either now or in the future.

Unless I'm mistaken, `addsfx` is called by `I_StartSound` every time a sound is played, which means the lump data is accessed every time the sound plays and this whole conversion is happening every time a sound is played. Is there somewhere that short circuits this flow of events if it has happened before?

fabiangreffrath: In Woof, all sounds are pre-cached at start-up unless sound pitch randomization is enabled. Maybe I am mixing up its concepts with PrBoom+/DSDA. Sorry if I added even more confusion.

rfomin: > Unless I'm mistaken, `addsfx` is called by `I_StartSound` every time a sound is played, which means the lump data is accessed every time the sound plays and this whole conversion is happening every time a sound is played. Is there somewhere that short circuits this flow of events if it has happened before?

I think we need some sort of caching data structure. I'll try to add it a little later.

prboom2/src/SDL/i_sound.c
```diff
@@ -502,19 +607,30 @@ static void I_UpdateSound(void *unused, Uint8 *stream, int len)
...
-              + ((unsigned int)channelinfo[chan].data[1] * (channelinfo[chan].stepremainder))
-              - 0x800000; // convert to signed
+        if (ci->bits == 16)
+        {
+          s = (short)(ci->data[0] | (ci->data[1] << 8)) * (255 - (ci->stepremainder >> 8))
+            + (short)(ci->data[2] | (ci->data[3] << 8)) * (ci->stepremainder >> 8);
+        }
+        else
+        {
+          s = (ci->data[0] * (0x10000 - ci->stepremainder))
```
kraflab: What is the significance of removing the cast to unsigned int from these lines?

prboom2/src/SDL/i_sound.c
kraflab: 👍

prboom2/src/SDL/i_sound.c
rfomin: I just copied it from PrBoom+ without thinking. I'll restore it just in case.

kraflab (approved):
Thanks! 💪

PR #112, opened 2022-06-21
Z_FreeTag: always free memory from the given tag up to PU_MAX

fabiangreffrath:
Also, some minor unrelated clean-up.

### DIFF
--- a/prboom2/src/z_zone.c
+++ b/prboom2/src/z_zone.c
@@ -60,15 +60,8 @@
 #include <dpmi.h>
 #endif
 
-// Tunables
-
 // signature for block header
 #define ZONEID  0x931d4a11
-
-// Number of mallocs & frees kept in history buffer (must be a power of 2)
-#define ZONE_HISTORY 4
-
-// End Tunables
 
 typedef struct memblock {
   unsigned id;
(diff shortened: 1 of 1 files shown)


prboom2/src/z_zone.c
```diff
@@ -163,24 +156,23 @@ void Z_FreeTag(int tag)
...
-
   if (tag < 0 || tag >= PU_MAX)
     I_Error("Z_FreeTag: Tag %i does not exist", tag);
 
-  block = blockbytag[tag];
-  if (!block)
-    return;
-  end_block = block->prev;
-  while (1)
+  for (; tag < PU_MAX; tag++)
```
kraflab, line 162: Why? There is no "order" to tags, so this would be an unexpected side effect.

fabiangreffrath: He, there is. They are ordered in an `enum` from "static" to "static for the duration of a level" to "throw away". Note that Vanilla code always calls `Z_FreeTags(a,b)` with `b = PU_CACHE` (or `PU_PURGELEVEL-1`, that is).

fabiangreffrath: So, at the beginning of a new level, all level-specific and cache data is cleared.

kraflab: I understand that they happen to be in an order, and I understand that vanilla worked differently, but they do not have a guaranteed order in dsda-doom, it's incidental, and any one tag has no bearing on any other tag, so this operation has no value.

kraflab: I.e., let's say I introduced a tag for storing data related to graphics settings, and when the graphics are reinitialized we clear that tag. That would be nice and convenient, but how would it fit into the way you would clear tags?

kraflab: This would also mean that clearing the level tag would unload all the unlocked lumps, so it would just degrade performance.

fabiangreffrath: > This would also mean that clearing the level tag would unload all the unlocked lumps, so it would just degrade performance.

Yes, that's how it's done in, let's say, every other Doom source port. 😉 But I buy that argument about introducing another special-purpose tag that won't line up in the classic tag order.

PR #113, opened 2022-06-21
Cleanup zone interface

kraflab:
This PR does a few things:

1) Remove data locks
Data that gets "unlocked" is not freed (despite many comments mentioning that this happens). That's because unlocked data is just marked as being unimportant, in case we _run out of memory completely_. If we somehow run out of memory completely, then something catastrophic has happened. So, in all other cases, all the locking / unlocking code doesn't actually do anything.

2) Remove zone users
The zone user pointer was used to change the "owner" of the data to NULL when unlocked data was purged. Since data locking no longer exists, this functionality is obsolete.

3) Change the wording everywhere away from "cache" for the associated data.
There was no data cache so this was misleading.

4) Hide zone tags
The zone interface was a bit too low-level. Any given file doesn't need to know about zone tagging or how the internals work. There are currently two uses for zone memory: regular malloc with error handling (doesn't care about tags), and level-scoped data allocation / deletion (doesn't care about tags). Since neither of these cases needs to specify tags, we can scrub them out of the outer code base entirely.

### DIFF
--- a/prboom2/src/MUSIC/flplayer.c
+++ b/prboom2/src/MUSIC/flplayer.c
@@ -112,7 +112,7 @@ static void *fl_sfopen(const char *lumpname)
   MEMFILE *instream;
   int lumpnum = W_GetNumForName(lumpname);
   int len = W_LumpLength(lumpnum);
-  const void *data = W_CacheLumpNum(lumpnum);
+  const void *data = W_LumpByNum(lumpnum);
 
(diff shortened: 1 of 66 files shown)


kraflab:
@fabiangreffrath FYI

I'm sure you won't agree with all of these changes, but it's more food for thought anyway (and the commits should be well-separated).

PR #117, opened 2022-07-31
CMake: Set _FILE_OFFSET_BITS=64 on Linux

BigBigos:
By default without _GNU_SOURCE, off64_t is not defined. This breaks
MUSIC/dumbplayer.c workaround for making off_t 64-bit. Even 64-bit Linux
is affected.

Define _FILE_OFFSET_BITS=64 explicitly so that off_t is defined as a
64-bit type. This also skips the workaround in dumbplayer.c source file.

The `UNIX AND NOT APPLE` is assumed to mean `Linux`, though it might
trigger on non-Linux systems as well.

---

I only tested this on my Gentoo x86-64 machine, building both in 32-bit and 64-bit mode.

The idea of adding `_GNU_SOURCE` came to me by comparing the compiler flags while building dsda-doom and coelckers/prboom-plus. In the latter, there is an unrelated change that adds `_GNU_SOURCE`:

https://github.com/coelckers/prboom-plus/commit/c87abe0fc10e3a3c9d5a94db3f7881b086dad971

### DIFF
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -50,6 +50,10 @@ if(NOT HAVE_STRNICMP)
     add_definitions("-Dstrnicmp=strncasecmp")
 endif()
 
+if(UNIX AND NOT APPLE)
+    add_definitions("-D_FILE_OFFSET_BITS=64")
+endif()
+
 check_symbol_exists(getopt "unistd.h" HAVE_GETOPT)
 check_symbol_exists(mmap "sys/mman.h" HAVE_MMAP)
 check_symbol_exists(CreateFileMapping "windows.h" HAVE_CREATE_FILE_MAPPING)


kraflab:
As far as I'm aware, defining `_GNU_SOURCE` is a pretty big change. Is there a less intrusive change we can make?
Take a look at how the missing `off64_t` was fixed for mac here: https://github.com/kraflab/dsda-doom/commit/f8495a81549cb62c749729afe232edfb46c9dd5a. Is there an equivalent (targeted) fix for this case?

fabiangreffrath:
Why not just `int64_t` for all arches? All we'd need is `stdint.h` for that.

BigBigos:
We can define `_FILE_OFFSET_BITS=64` explicitly. This would both make sure the `#if` tree is skipped and `off_t` is defined as 64-bit.

I've confirmed it works on both 32-bit and 64-bit Linux.

fabiangreffrath:
Isn't this usually defined by a system header instead of by the user?

Edit: `features.h`?

BigBigos:
According to [this](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html), the macro is meant to be defined by the user. It tells the standard library how to define functions and types that deal with file sizes.

It also recommends defining these in source files instead of using `-D`. Since `<dumb.h>` is included in a single source file only, it could be a viable alternative. Should I pursue such a solution?

fabiangreffrath:
> Why not just `int64_t` for all arches? All we'd need is `stdint.h` for that.

```c
 * 2. Supply your own definition of a signed 64-bit integer
 *    such as off64_t or int64_t before including dumb.h as follows:
 *    #define DUMB_OFF_T_CUSTOM int64_t
```

https://github.com/kode54/dumb/blob/396caa4d31859045ccb5ef943fd430ca4026cce8/include/dumb.h#L133-L135

kraflab:
Fixed in https://github.com/kraflab/dsda-doom/pull/119

PR #119, opened 2022-08-02
supply int64_t as the dumb library's custom offset type

fabiangreffrath:


prboom2/src/MUSIC/dumbplayer.c
```diff
@@ -68,10 +68,9 @@ const music_player_t db_player =
 #if !defined(_FILE_OFFSET_BITS) || (_FILE_OFFSET_BITS < 64)
 #ifdef _MSC_VER
 #define DUMB_OFF_T_CUSTOM __int64
-#elif defined(__APPLE__)
-#define DUMB_OFF_T_CUSTOM off_t
 #else
-#define DUMB_OFF_T_CUSTOM off64_t
+#include <stdint.h>
+#define DUMB_OFF_T_CUSTOM int64_t
```
fabiangreffrath, line 73: Alternative: `long long`

kraflab:
@BigBigos could you confirm this works fine for you?

prboom2/src/MUSIC/dumbplayer.c
kraflab: I prefer `int64_t` for sure :^)

fabiangreffrath: 👍

BigBigos:
This builds correctly on both 32-bit and 64-bit Linux, thanks!

kraflab:
Thanks 👌

PR #122, opened 2022-08-14
Fix unknown type name ‘uint64_t’ error.

FtZPetruska:
## Description

This error would happen in docker containers such as `ubuntu:latest`.

This was first noticed in https://github.com/Homebrew/homebrew-core/pull/108013 when running the CI.

## Alternative solution

Another solution could be to include `stdint.h` if preferred.

### DIFF
--- a/prboom2/src/d_deh.c
+++ b/prboom2/src/d_deh.c
@@ -1124,7 +1124,7 @@ struct deh_flag_s {
   uint_64_t value;
 };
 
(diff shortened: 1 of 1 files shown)


kraflab:
Heh, this was just a typo I guess?

kraflab:
Thanks!

Pedro-Beirao:
Made a PR to add dsda-doom to brew about a year ago, but the repo was not popular enough for them to accept.

Great to see that the port grew so fast!

Once it gets merged, I'll delete my custom tap

PR #123, opened 2022-08-17
Add separate install option for `dsda-doom.wad`

FtZPetruska:
## Description

This is a suggestion from Homebrew/homebrew-core#108013.

This PR adds a CMake cache entry named `EXTRA_DOOMWADDIRS` which consists in a list of additional directories to check for WADs similar to `DOOMWADDIR`.

## Motivation

While working on the Homebrew integration we noticed the following issues:
- When leaving `DOOMWADDIR` to its default (`<prefix>/share/games/doom`), the `games/doom` part gets symlinked into `HOMEBREW_PREFIX`. The issue is that when reinstalling/updating/uninstalling dsda-doom, this folder would get wiped and all your WADs erased.
- When setting `DOOMWADDIR` to `HOMEBREW_PREFIX/share/games/doom`, build fails on macOS due to quarantine as CMake tries to create the missing directories when it doesn't have the privileges to do so.

The current makeshift solution is the following:
- Install step:
  - Leave `DOOMWADDIR` to its default.
  - After CMake install, move `dsda-doom.wad` away. 
- Post-install step:,
  - Create `HOMEBREW_PREFIX/share/games/doom`.
  - Symlink `dsda-doom.wad` into it
  - Symlink `HOMEBREW_PREFIX/share/games/doom` into dsda's prefix.

With this PR, `HOMEBREW_PREFIX/share/games/doom` could be set as an `EXTRA_DOOMWADDIRS` and be searched while not being deleted when dsda gets updated.

It's worth mentioning that `HOMEBREW_PREFIX` is not the same on all platforms and can be chosen by the user, here are the defaults:
- `/opt/homebrew` (M1 mac)
- `/usr/local` (x86_64 mac)
- `/home/linuxbrew/.linuxbrew` (Linux)

### DIFF
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -135,6 +135,7 @@ set(CMAKE_REQUIRED_INCLUDES ${CMAKE_REQUIRED_INCLUDES_PREV})
 set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES_PREV})
 
 set(DOOMWADDIR "${CMAKE_INSTALL_PREFIX}/share/games/doom" CACHE PATH "Path to look for WAD files")
+set(DSDAPWADDIR "${DOOMWADDIR}" CACHE PATH "Path to install DSDA-Doom internal WAD")
 
 option(SIMPLECHECKS "Enable checks which only impose significant overhead if a posible error is detected" ON)
 
--- a/prboom2/cmake/config.h.cin
+++ b/prboom2/cmake/config.h.cin
@@ -5,6 +5,7 @@
 #cmakedefine PACKAGE_STRING "@PACKAGE_STRING@"
 
 #cmakedefine DOOMWADDIR "@DOOMWADDIR@"
+#cmakedefine DSDAPWADDIR "@DSDAPWADDIR@"
 
 #cmakedefine WORDS_BIGENDIAN
 
--- a/prboom2/data/CMakeLists.txt
+++ b/prboom2/data/CMakeLists.txt
@@ -425,4 +425,4 @@ add_custom_command(
     DEPENDS rdatawad ${WAD_SRC}
 )
 add_custom_target(dsda-doom-wad DEPENDS ${WAD_DATA_PATH})
-install(FILES ${WAD_DATA_PATH} DESTINATION ${DOOMWADDIR} COMPONENT "PrBoom-Plus internal WAD")
+install(FILES ${WAD_DATA_PATH} DESTINATION ${DSDAPWADDIR} COMPONENT "DSDA-Doom internal WAD")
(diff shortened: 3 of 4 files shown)


kraflab:
What do the other ports in homebrew do about this? Do they use the same solution?

FtZPetruska:
The other ports available in Homebrew are:
- Crispy Doom
- Chocolate Doom
- GZDoom (as a Cask, a native macOS App bundle)

For the first two, they do not ship with an additional pwad, or provide a way to add additional `DOOMWADDIR` at compile time. For Homebrew, this means nothing gets linked in `HOMEBREW_PREFIX/share/games/doom` and as such nothing gets overwritten during updates.

GZDoom being shipped as a cask, it isn't built by Homebrew (plus GZDoom uses a config file to search for WADs).

kraflab:
Is this line the essential problem? The fact that it's writing to that directory?
```c
install(FILES ${WAD_DATA_PATH} DESTINATION ${DOOMWADDIR} COMPONENT "PrBoom-Plus internal WAD")
```
Would changing where the internal wad is placed be another option? (like `share/games/dsda-doom` 🤷)

I'm not familiar enough with cmake to completely understand the situation. If there should just be one common directory for general wad files, it feels like it should be unnecessary to build in the option of defining extra ones.

@fabiangreffrath do you have any insight into this topic?

FtZPetruska:
In the case of [Pedro-Beirao/homebrew-dsda-doom](https://github.com/Pedro-Beirao/homebrew-dsda-doom)'s homebrew Formula, they don't rely on CMake's install command but manually install the executable and the pwad to the bin directory.

While it works, and would solve the issue, the pwad gets linked to the bin directory which is not desired by the maintainers.

> Would changing where the internal wad is placed be another option? (like share/games/dsda-doom 🤷)

This could work if that path is configurable, such as a `DSDAPWADDIR` variable, even if its default value is the same as `DOOMWADDIR`. This could work like:

In the top level `CMakeLists.txt`:
```cmake
set(DSDAPWADDIR "${DOOMWADDIR}" CACHE PATH "Path to install DSDA-Doom internal WAD")
```
In `data/CMakeLists.txt`:
```cmake
install(FILES ${WAD_DATA_PATH} DESTINATION ${DSDAPWADDIR} COMPONENT "PrBoom-Plus internal WAD")
```
In `config.h.cin`:
```c
#cmakedefine DSDAPWADDIR "@DSDAPWADDIR@"
```
In `src/SDL/i_system.c`:
```c
search0[] = {
...
{DSDAPWADDIR}, // build-time configured location of dsda-doom.wad
...
};
```

Would this change be preferable ?

kraflab:
Ya, that change looks more natural to me 👍 
By the way, since you're touching that line, feel free to change "PrBoom-Plus" to "DSDA-Doom" if you want, I missed this one :^)

kraflab:
Thanks 👍

PR #131, opened 2022-09-12
Fix sound attenuation when distance overflows

decino:
Sound emitters play their sounds at full volume when the listener-to-sound distance approximation is greater than 32,767 due to the value overflowing to a negative one. An example where this can happen is in large maps where enemies are awake in faraway out of bounds teleporter closets.

### DIFF
--- a/prboom2/src/s_sound.c
+++ b/prboom2/src/s_sound.c
@@ -727,7 +727,8 @@ void S_StopChannel(int cnum)
 int S_AdjustSoundParams(mobj_t *listener, mobj_t *source,
                         int *vol, int *sep, int *pitch)
 {
-  fixed_t adx, ady,approx_dist;
+  fixed_t adx, ady;
+  ufixed_t approx_dist;
   angle_t angle;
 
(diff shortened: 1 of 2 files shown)


prboom2/src/s_sound.c
```diff
@@ -773,7 +773,9 @@ int S_AdjustSoundParams(mobj_t *listener, mobj_t *source,
       return *vol > 0;
     }
 
-  if (approx_dist > S_CLIPPING_DIST)
+  // Sound emitter is too far away from listener.
+  // The <= 0 check is done in case the approx_dist overflows to a negative value.
+  if (approx_dist > S_CLIPPING_DIST || approx_dist <= 0)
```
kraflab: How about adding a typedef for `ufixed_t` as an `unsigned int` in `m_fixed.h` and changing `approx_dist` to `ufixed_t`? Then we fix the problem at its core and don't have to edit the check.

kraflab:
Thanks 😎

PR #132, opened 2022-09-12
New death use action to restart demo/level

JokieW:
This adds a death use action that fully restarts the demo/level, like the "Restart Demo"/"Restart Level" button does.

That button is super helpful, but I have been doing nightmare runs, which involves a lot of dying. It's easy in the heat of the moment to press "use" instead of "restart" after dying and not realizing you're still recording in the old demo. Or sometimes you try to open a door or something at the same time you die, and you forget to restart for a new demo. The "nothing" action works but it breaks the flow a bit.

When the "death use action" is set to "restart", and a demo isn't being played back, the death use action will restart the demo if recording, or the level if not, just like the button usually does. Another advantage is that since it's a full restart, you start with the same rng everytime, unlike the regular respawn.

### DIFF
--- a/prboom2/src/dsda/death.c
+++ b/prboom2/src/dsda/death.c
@@ -32,12 +32,16 @@ typedef enum {
   death_use_default,
   death_use_nothing,
   death_use_reload,
+  death_use_restart
 } death_use_action_t;
 
 int dsda_death_use_action;
 
 static int dsda_DeathUseAction(void)
 {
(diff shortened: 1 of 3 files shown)


kraflab:
It's not an accident that this feature doesn't already exist. I don't like the idea of changing the behavior of basic input like this while recording a demo. Maybe I'll change my mind at some point, but that's how it is right now. The attempts message at demo start (or lack thereof) can be an indicator if you're unsure if you started a new demo or not.

PR #134, opened 2022-09-13
Migrate config interface

kraflab:
Previously, configuration variables were treated in various ways by the code base. Some were psuedo-constant globals, others were used as "default values" that could be overwritten into emergent variables (`use_fullscreen` vs `desired_fullscreen`), some were incorporated into the menu via update callbacks, etc.

This PR seeks to create a consistent interface for configuration with clearly separated responsibilities and more flexibility.

Requirements:
1. Every config variable must have a persistent value (what the file sees) and a transient value (what the code sees).
2. Every config variable must trigger necessary side effects when updated.
3. Every config variable must account for strict mode in its emergent value.
4. Config reads and writes must be variable-agnostic.

What this does:
- Extracts configuration details (names, defaults, callbacks) into `dsda/configuration`
- Removes all config information from the menu except the config IDs
- Removes all direct config variable reads and writes
- Removes many superfluous config variables
- Significantly reduces the number of global variables and dependencies
- Leaves the file presentation (what settings are listed where, with what headings) in `m_misc`

This isn't quite as far as I would like to go - the configuration file shouldn't know the names of any global variables for instance. But it's a good stopping point along the way. As it is, this is probably the biggest single change to the code base.


kraflab:
@fabiangreffrath I imagine this is way too overkill for woof but maybe some ideas here will interest you some day 😄

PR #137, opened 2022-09-14
Fix intermission screen split data time diff

### LINKED ISSUE #136: Wrong split data time difference format
https://www.twitch.tv/videos/1590977799?t=1h8m38s

If the user has "Show Split Data" enabled then the difference would display as "+1:5.20" instead of the expected value of "+1:05.20".

kmeaw:
If the number of minutes is less than 10 we should have a leading '0'.

Fixes #136

### DIFF
--- a/prboom2/src/dsda/intermission_display.c
+++ b/prboom2/src/dsda/intermission_display.c
@@ -80,7 +80,7 @@ static void dsda_UpdateIntermissionTime(dsda_split_t* split) {
     if (diff >= 2100) {
       snprintf(
         delta, sizeof(delta),
-        " (%s%d:%04.2f)",
+        " (%s%d:%05.2f)",
         sign, diff / 35 / 60, (float)(diff % (60 * 35)) / 35
       );
     }


prboom2/src/dsda/intermission_display.c
```diff
@@ -80,14 +80,14 @@ static void dsda_UpdateIntermissionTime(dsda_split_t* split) {
     if (diff >= 2100) {
       snprintf(
         delta, sizeof(delta),
-        " (%s%d:%04.2f)",
+        " (%s%d:%05.2f)",
```
kraflab, line 83: It's just this one right? Why is the other case changed?

kmeaw: I just thought it would be nice to have a consistent 0 padding in both cases, like on stopwatches: +00.03, ..., +59.97, +1:00.00
If you don't think so then I can just leave this case and discard the other one.

kraflab: Yes please do so. If I decide to line then up somehow I may do that later on

kmeaw: Sure, I have left just this one and discarded the other.

kraflab:
Thanks 😎

PR #143, opened 2022-09-30
Purge demoex file

kraflab:
Previously, loading the demo footer involved these steps:

- Load the demo file
- Locate the end marker
- Check for the pwad signature
- Copy the demo footer as a pwad into a temporary file
- Add the temporary file to the wad list
- Load _all_ wad files
- Read the necessary lump from the demo ex pwad
- Unload _all_ wad files
- Unload the demo

This PR changes it to instead read the demo footer in memory directly.

There is another quirk of the startup procedure:

- The port tries to guess the IWAD (defaults to doom2 if it finds it)
- The demo footer can add an IWAD (-iwad DOOM.WAD for instance)
- Now the wad file list has multiple iwads, and the autoload may load wads for the wrong iwad

This PR marks obsoleted IWADs to be skipped, which causes them to be ignored for autoload as well as when loading the wad files themselves in W_Init.

### DIFF
--- a/prboom2/src/SDL/i_main.c
+++ b/prboom2/src/SDL/i_main.c
@@ -192,10 +192,6 @@ static void I_EssentialQuit (void)
 static void I_Quit (void)
 {
   M_SaveDefaults ();
-
-  // This function frees all WAD data as a side effect (!!!)
-  // You MUST NOT call this function before any code that touches lump data (e.g., music shutdown)
-  I_DemoExShutdown();
 }
 
 //
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1439,6 +1439,27 @@ static void HandleClass(void)
   randomclass = dsda_Flag(dsda_arg_randclass);
 }
 
+static void HandlePlayback(void)
+{
+  const char* name;
+
+  name = dsda_ParsePlaybackOptions();
+
+  if (name)
+  {
+    char *file = Z_Malloc(strlen(name) + 4 + 1); // cph - localised
+    strcpy(file, name);
+    AddDefaultExtension(file, ".lmp");     // killough
+    D_AddFile (file, source_lmp);
+    //jff 9/3/98 use logical output routine
+    lprintf(LO_INFO, "Playing demo %s\n", file);
+    Z_Free(file);
+
+    //e6y
+    G_CheckDemoEx();
+  }
+}
+
 const char* doomverstr = NULL;
 
 static void EvaluateDoomVerStr(void)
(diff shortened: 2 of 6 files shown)


kraflab:
@fabiangreffrath There's a bug that this PR fixes which is also present in prboom+:
```
cp Eviternity.wad autoload/DOOM2.WAD
prboom-plus.exe -iwad DOOM.WAD -record mydemo.lmp
prboom-plus.exe -playdemo mydemo.lmp # crash
```
Basically, have some wad in your DOOM2.WAD autoload. Then play a demo for doom 1. The autodetect iwad logic will grab DOOM2.WAD because there is no explicit `-iwad` argument. Then the footer parsing will find `-iwad DOOM.WAD` in the demo itself and add DOOM.WAD to the load order. But this happens after the autoload occurs - which means the doom 2 autoload wads are already in the load order. In the case above this will cause a crash. In other cases the user will just end up with the wrong files.

The code is getting to be quite different between the ports in D_DoomMainSetup, but the same idea could be applied there.

1st: https://github.com/kraflab/dsda-doom/pull/143/commits/8bde77a2ca30b41259d1a1562ba7b1d0dbc0026b (mark existing iwad entries for skipping when adding a new one)

2nd: Move the code that checks playback files and demo footers _above_ the code that does autoloading

Personally, D_DoomMainSetup is the scariest function in the source for me, and changing the order of events there is not something I like to do. I'm less apprehensive in dsda-doom because a ton of the code there has been cleaned up, and various parts simplified. There also is no embedded launcher in dsda-doom anymore, but the footer reading in prboom-plus also affects how the launcher works. So moving things around may also change how and when that can pop up? I don't really know - basically I would just leave it bugged in prboom-plus, but maybe you are more confident about that function than I am 😄

fabiangreffrath:
Oh my! I was hoping to never have to touch that code again. :laughing: 

Your solution in https://github.com/kraflab/dsda-doom/commit/8bde77a2ca30b41259d1a1562ba7b1d0dbc0026b looks very clever. However, it looks to me that if a second IWAD is added to the `wadfiles[]` array, then all former IWADs lose their status. Is this the correct order? Shouldn't the first IWAD that was added keep its status and all subsequent ones get ignored?

kraflab:
The initial search for a default iwad can only find one iwad (which is either the first it finds or one set by `-iwad`). If a demo footer specifies an iwad being required, then the demo footer "wins". However, if the user supplied `-iwad` themselves, then the information in the footer is ignored. So, we do want the latest iwad added to be the one we keep.

fabiangreffrath:
Understood, thanks!

PR #157, opened 2022-11-13
Fix red translucent sprites in indexed lightmode

rrPKrr:


kraflab:
This fixes the sprites but not the fact that fuzz is drawn on top of the pain screen, is that right?

rrPKrr:
Yes, that's right.

jackrjli:
This also fixes an issue I saw with translucent textures in Bauhaus MAP01:
doom00

kraflab:
Thanks 😸

kraflab:
@XaserAcheron FYI - depending on how the fuzz is fixed later on, we may want to revert this

PR #158, opened 2022-11-14
Send all notes/sound off prior to SysEx reset

### LINKED ISSUE #153: PortMidi volume spikes
Reported by zeromaster on discord:

There is occasionally a music volume spike when quitting dsda-doom. Also happens when entering the intermission screen. It only happens with PortMidi and not when using FluidSynth.

ceski-1:
Fixes https://github.com/kraflab/dsda-doom/issues/153


kraflab (approved):
Thanks 👏

PR #164, opened 2022-12-12
Add setting for showing BFG tracers

janchochol:
BFG tracers is quite complicated mechanism and difficult to properly learn.
This new option and new TAS setting (only for OpenGL) to show BFG tracers (and if they hit), which will make learning usage of BFG much more easier.

### DIFF
--- a/prboom2/src/dsda/configuration.c
+++ b/prboom2/src/dsda/configuration.c
@@ -196,6 +196,9 @@ void dsda_TrackConfigFeatures(void) {
   if (dsda_IntConfig(dsda_config_gl_health_bar))
     dsda_TrackFeature(uf_healthbar);
 
+  if (dsda_IntConfig(dsda_config_gl_bfg_tracers))
+    dsda_TrackFeature(uf_bfg_tracers);
+
   if (dsda_IntConfig(dsda_config_movement_strafe50))
     dsda_TrackFeature(uf_alwayssr50);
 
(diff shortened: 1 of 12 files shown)


janchochol:
Ia m not sure, if I used correct "code structure" (editing correct files at correct places), so I have no problem to alter the PR if needed.

kraflab:
Conceptually it's a cool idea, but I'm not sure about the current implementation.

- Sometimes the lines don't show up, not sure why
- If you're standing still, the lines render as parallel lines, which is inaccurate
- The lines go through walls, so they aren't accurate indicators of where the tracers can go
- In general, the lines don't hit their targets visually - a line may run into the floor or a wall with the enemy above

I'm not sure if this has a ton of value in its present form - it could help with getting an idea of the angle spread of the bfg, but this is quite a limited application.

Something I am planning eventually is to track all line attacks generally - that would cover a lot of use cases and the bfg would be one of them. This PR is good inspiration for that, although I may not merge it on its own.

janchochol:
Thanks for the review!

* It never happened to me - I'll try investigate more
* Never spot this, but it is surprisingly correct - as center of camera is same as originating point, tracer should look parallel when viewed from same place (try moving little bit after tracers are draw)
* Yes, that is true (fix will require some more changes in `P_AimLineAttack`, `PTR_AimTraverse` and similar)
* Yes, that is unfortunate, but I did not find a way how to make it better.

Biggest problem is, how to make it intuitive. I tried point tracers to `linetarget` if tracer hit, but it was even more confusing.
Currently it is intuitive for me, as I implemented it in a way, which suits me.

Anyway, I definitely do no insist of merging the PR - if you find it usable with some modification, I definitely can try to implement them.

Even if you close this PR, this is useful at least for me (I found why I am that bad in two-hitting Cybers :smile:) and it will not be loss of time for me.

kraflab:
Going to close this one, but may return to it some day.

PR #170, opened 2022-12-17
Fixes and addition about macOS

fraansg:
- In the macOS guide:
  - Changed the brew formula relative to libmad (alias for mad package)
  - Fixed the options in zip command (--recurse-paths to include the .dylib files and --exclude .\* for Desktop files from the compressed archive).
- Ignore .DS_Store and /release*

Cc: @Pedro-Beirao


kraflab:
Thanks 👍

PR #171, opened 2022-12-20
Indexed Lightmode Spectre Fuzz Fix

### LINKED ISSUE #152: Spectres appear on top of pain palette in Indexed light mode
See screenshot reported by Vile.

doom01

XaserAcheron:
Fixes #152 (well, the actual reported issue at least, not any of the other random things in comments). 

It's still a bit fudgy in how it works internally (i.e. the spectre fuzz isn't using the colormap yet 'cause that's hard), but it's Good Enough(tm) for a first release, at least.

There's a couple of vague ideas brewing in my head on how to maybe make a better fuzz effect in general, but it does involve some changes to the new lightmode that are way out of scope for 0.25 (and I don't yet know if it'll even work ;P ).


kraflab:
Nice 😎

Issue #161, opened 2022-12-08
V_FillRect broken in indexed light mode when ui mode is active


kraflab:
This affects some miscellaneous effects like the demo progress bar and console background (but also the heretic start screen).

Possible fix:
https://github.com/kraflab/dsda-doom/compare/master...fix-v-fill-rect

PR #174, opened 2022-12-21
Fix Raw screens (e.g. Heretic titlepic) in indexed lightmode

XaserAcheron:
See title -- this partially addresses #161, but I haven't yet covered the general V_FillRect case just yet (got another train of thought on how to fix that which I need to play with).

Nice bonus: this fix actually uses the shader, rather than disabling it, so gamma works for the titlepic:
2022-12-21 00_07_31-dsda-doom 0 24 3

Heretic's E2 endscreen and E3 scroller are working with this new method, BTW -- still no gamma in intermission screens but that's a separate thing to address; if nothing else, it's consistent. ;)


kraflab:
Nice 🥳

PR #175, opened 2022-12-23
Fix IDDIT/IDDKT not being repeatable with only one item/monster

jackrjli:
This fixes a bug where the IDDIT and IDDKT (but not IDDST) cheat codes would only work once if there were only one item/monster remaining (the cheat would work the first time, but subsequent times would not recenter the automap on the item/monster).

Also made it so that follow mode isn't turned off if there are no items/monsters remaining, like how IDDST behaves with no secrets remaining.

### DIFF
--- a/prboom2/src/m_cheat.c
+++ b/prboom2/src/m_cheat.c
@@ -646,7 +646,6 @@ static void cheat_reveal_secret()
   }
 }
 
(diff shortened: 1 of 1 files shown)


kraflab:
Thanks 👌

Issue #168, opened 2022-12-14
Intermission does not account for gamma in indexed light mode


kraflab:
Compare the menu (does account for gamma) with the intermission elements:
![doom04](https://user-images.githubusercontent.com/25529060/207671891-b01b3420-9055-49f8-911f-6b4a82debaba.png)

kraflab:
Fixed by https://github.com/kraflab/dsda-doom/pull/176

PR #176, opened 2022-12-23
Misc Indexed Lightmode UI Fixes

### LINKED ISSUE #161: V_FillRect broken in indexed light mode when ui mode is active

XaserAcheron:
Fixes #161 and #168 and polishes up a few more things for the indexed lightmode. The highlights:

- The shader is used for intermission & finale screens now, so gamma works.
- UI drawing now uses the palette directly, rather than the colormap at full-brightness, since there's no guarantee that the brightest range in the colormap is 1:1 with the palette -- Heretic and Hexen exhibit exactly this 'problem', which resulted in some inaccurate color mappings (the Heretic E2 endmap in particular was very borked).
- V_FillRect is fixed; it now temporarily suspends shaders so it can do its thing -- not really an ideal fix (at some point I need to take a look at why the shader ain't working), but it'll do for now.
- Flushing textures while not in a map would accidentally ditch the colormap textures, resulting in none of the UI drawing. The Heretic E2 Endscreen's palette change is what triggered the flush in this case, but this was certainly ungood and may have led to other oopsies. :P

This handles all the known issues at time of writing. Here's hoping we don't find too many after it hits the shelves. :P


kraflab:
😎

PR #183, opened 2022-12-24
Update PortMidi with PrBoom+ changes

### LINKED ISSUE #177: Title music in DBP37 don't work with portmidi backend
@ceski-1 changed the default SysEx reset in PrBoom+ because the GS reset requires "capital tone fallback" emulation. A quick fix is to change the default reset to `"gm"` here:https://github.com/kraflab/dsda-doom/blob/23a0633e7808a1bcd069fa5888f16f0e40431122/prboom2/src/dsda/configuration.c#L753

ceski-1:
Fixes https://github.com/kraflab/dsda-doom/issues/177


kraflab:
Thanks!

PR #187, opened 2022-12-30
Add fix for MS GS Synth bug

ceski-1:
Reported by El Juancho on Discord. MS GS Synth resets pitch bend sensitivity to the wrong value so this must be set manually after a GM/GS reset, not before. Test wad: [pbs.zip](https://github.com/kraflab/dsda-doom/files/10326262/pbs.zip)
```
dsda-doom.exe -iwad doom2.wad -file pbs.wad -warp 1
```
The "reed organ" at ~23s should change pitch.


kraflab:
Thanks 🥳

Issue #191, opened 2023-01-10
Fluidsynth sample-rate out of range


BlipBlerp:
I've been having this issue where Fluidsynth seems to be ignoring my settings. This only happens to me on 0.25.x. I have deleted my cfg multiple times but it keeps happening. These are my sound settings:
```
# Sound settings
snd_pcspeaker                  0
pitched_sounds                 0
full_sounds                    0
snd_samplecount                512
snd_samplerate                 44100
sfx_volume                     1
music_volume                   1
mus_pause_opt                  1
snd_channels                   32
snd_midiplayer                 "fluidsynth"
snd_soundfont                  "RLNDGM.SF2"
snd_mididev                    ""
mus_fluidsynth_chorus          0
mus_fluidsynth_reverb          0
mus_fluidsynth_gain            50
mus_opl_gain                   50
mus_portmidi_reset_type        "gs"
mus_portmidi_reset_delay       0
mus_portmidi_filter_sysex      1
mus_portmidi_reverb_level      40
mus_portmidi_chorus_level      0
```

Whenever I launch dsda-doom I get this error message in the console logs, which I guess could be related
```
fluidsynth: error: requested set value for 'synth.sample-rate' out of range
fl_init: Couldn't set synth.sample-rate
```

kraflab:
@ceski-1 do you have any idea about this one?

ceski-1:
I can reproduce this by setting my sound card's sample rate to 192kHz. Setting it to 48kHz is fine. If I compile with MSYS2 (MinGW 64-bit), to grab newer versions of FluidSynth, etc., I can use any sample rate and there's no error. I'll take a closer look in a bit.

ceski-1:
Here's the problem: https://github.com/kraflab/dsda-doom/blob/144709bf29dc8477220e3c9cdd6ed37786861385/prboom2/src/SDL/i_sound.c#L787

Here's what `Mix_QuerySpec()` returns for `snd_samplerate` for different versions of SDL2 with `snd_samplerate 44100` in the config and the sound card set to 192kHz:

| SDL2 ver. | Mix_QuerySpec() |
| --- | --- |
| 2.0.18 | 44100 |
| 2.0.20 | 44100 |
| 2.0.22 | 192000 |
| 2.24.0 | 192000 |
| 2.24.1* | 192000 |
| 2.24.2 | 192000 |
| 2.25.1 | 44100 |
| 2.26.0 | 44100 |
[...]

PR #201, opened 2023-01-28
Remove `Mix_QuerySpec()`; increase max `snd_samplerate`

ceski-1:
Draft for now, pending discussion here: https://github.com/kraflab/dsda-doom/issues/191

### DIFF
--- a/prboom2/src/SDL/i_sound.c
+++ b/prboom2/src/SDL/i_sound.c
@@ -783,9 +783,6 @@ void I_InitSound(void)
     return;
   }
 
-  // [FG] feed actual sample frequency back into config variable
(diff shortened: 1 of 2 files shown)


fabiangreffrath:
Please don't remove this. You even gave a proper explanation why this is needed: sometimes SDL initializes the soundcard with a different samplerate than requested. If we don't feed the actual value back into the variable, then for example the sound sample mixing code will still calculate with the requested value, instead of the actual one.

ceski-1:
Is there a way to query the actual sample rate being used by SDL and not the sound card's sample rate? Because that's the issue. If SDL's behavior changes again two versions from now, this problem will pop up again.

PR #202, opened 2023-01-28
PortMidi: Send "reset all controllers" when song loops

ceski-1:
Matches Vanilla Doom behavior and prevents hanging notes (NIN.WAD E1M8). Fixes a very old PrBoom+ bug: https://github.com/coelckers/prboom-plus/issues/559


kraflab:
Thanks!

PR #204, opened 2023-01-28
Support UTF8 locale with UCRT build

rfomin:
Note that non-Latin paths will only work on Windows 10+. It seems there are still many Win7 users in the Doom community.

### DIFF
--- a/prboom2/src/w_mmap.c
+++ b/prboom2/src/w_mmap.c
@@ -40,6 +40,7 @@
 #ifdef _WIN32
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
+#include "dsda/utility.h"
 #else
 #include <sys/mman.h>
 #endif
@@ -102,34 +103,6 @@ void W_DoneCache(void)
   mapped_wad = NULL;
 }
 
-static wchar_t *ConvertUtf8ToWide(const char *str)
-{
-    wchar_t *wstr = NULL;
-    int wlen = 0;
-
-    wlen = MultiByteToWideChar(CP_UTF8, 0, str, -1, NULL, 0);
-
-    if (!wlen)
-    {
-        return NULL;
-    }
-
-    wstr = malloc(sizeof(wchar_t) * wlen);
-
-    if (!wstr)
-    {
-        return NULL;
-    }
-
-    if (MultiByteToWideChar(CP_UTF8, 0, str, -1, wstr, wlen) == 0)
-    {
-        free(wstr);
-        return NULL;
-    }
-
-    return wstr;
-}
-
 void W_InitCache(void)
 {
   // Wipe any existing cache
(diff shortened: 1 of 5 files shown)


prboom2/src/w_mmap.c
```diff
@@ -102,6 +102,34 @@ void W_DoneCache(void)
   mapped_wad = NULL;
 }
 
+static wchar_t *ConvertUtf8ToWide(const char *str)
```
rfomin: Not sure where to place `ConvertUtf8ToWide` function, `m_misc.c`?

prboom2/src/w_mmap.c
```diff
@@ -128,9 +129,11 @@ void W_InitCache(void)
 #endif
       if (!mapped_wad[wad_index].data)
       {
-        mapped_wad[wad_index].hnd = CreateFile(wadfiles[wad_index].name,
+        wchar_t *wname = dsda_ConvertUtf8ToWide(wadfiles[wad_index].name);
```
kraflab, line 132: What decides when this conversion needs to happen and when it doesn't? Why in this file but not for instance when opening a demo file?

rfomin: This conversion is needed only for Win API functions, `CreateFileW()` in this case.

kraflab: Ok thanks

kraflab:
Just to double check, people not using UCRT can still compile this code, and builds created in win10 can still be used on win7, they'll just have the existing issue with non-latin paths. Is that right?

rfomin:
> people not using UCRT can still compile this code

Yes.

> builds created in win10 can still be used on win7, they'll just have the existing issue with non-latin paths

Yes, but win7 users may need redistributable/update: [docs](https://learn.microsoft.com/en-us/cpp/windows/universal-crt-deployment?view=msvc-170). Also you can't mix DLLs built with UCRT and "standard" CRT, for example, if the user downloads the DLLs from libsdl.org it won't work.

kraflab:
Hm, not sure I want to create builds that will require many users to take extra steps to actually use 😅

rfomin:
> Hm, not sure I want to create builds that will require many users to take extra steps to actually use 😅

I know 3 users with Win7 and everything immediately worked for them. We can wait for the next release of Woof, currently we have migrated from MSYS2 to MSVC with UCRT.
The alternative is the current Choco/Woof/PrBoom+ solution, I've successfully tested it on WinXP.

kraflab:
I'm not against going with the prboom+/woof solution, I just don't know how much work it will take (in general file handling in dsda-doom needs some work).

rfomin:
> I'm not against going with the prboom+/woof solution, I just don't know how much work it will take (in general file handling in dsda-doom needs some work).

I don't know how much this has changed in dsda-doom, but it should be easy. Just replace the file/path related functions with M_ wrappers (e.g. `fopen()` with `M_fopen()`). I can do that, but I think we should wait until the file handling changes?

kraflab:
Alright let me take a look at how things are working and try to consolidate things better 😄

kraflab:
Alright @rfomin I did a sweep and did some cleanup in advance. I tried to get all path-related opening into this file to start with: https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/m_file.c

There may still be things I missed, and I also didn't touch the special files like glob and windir because I figured these would be in the scope of the changes you will make, if they need updates.

kraflab:
Closing in favor of https://github.com/kraflab/dsda-doom/pull/234

PR #205, opened 2023-02-03
Fix weapon bobbing interpolation

ceski-1:
This uses an implementation similar to Crispy Doom to decouple weapon bob update rate from sprite timing. A smoother version is drawn on the screen while the original values are safely retained. Some related discussion [here](https://www.doomworld.com/forum/topic/121183-which-of-these-chainsaw-animations-is-the-correct-one/?tab=comments#comment-2292505).

https://user-images.githubusercontent.com/56656010/216520408-1f4ecc26-0c1c-47ea-9e2b-bf0e60507c11.mp4

### DIFF
--- a/prboom2/src/p_pspr.c
+++ b/prboom2/src/p_pspr.c
@@ -44,6 +44,7 @@
 #include "sounds.h"
 #include "d_event.h"
 #include "smooth.h"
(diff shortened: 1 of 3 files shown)


ceski-1:
I'm not convinced that this is the best solution.

PR #213, opened 2023-02-19
Grid layout for `weapon_text` HUD component

bkoropoff:
I found this easier to fit into my custom HUD layout.  Image attached.  Not sure if you'll like it or if you already have other plans for the HUD.

grid

 The first commit fixes a bug with rendering multi-line text.

### DIFF
--- a/prboom2/src/dsda/hud_components/weapon_text.c
+++ b/prboom2/src/dsda/hud_components/weapon_text.c
@@ -20,32 +20,58 @@
 #include "weapon_text.h"
 
 static dsda_text_t component;
+static dboolean grid;
 
 static void dsda_UpdateComponentText(char* str, size_t max_size) {
   player_t* player;
 
   player = &players[displayplayer];
 
-  snprintf(
-    str,
-    max_size,
-    "WPN \x1b%c%c\x1b%c %c %c %c %c %c %c %c %c",
-    player->powers[pw_strength] ? HUlib_Color(CR_BLUE) : HUlib_Color(CR_GREEN),
-    player->weaponowned[0] ? '1' : ' ',
-    HUlib_Color(CR_GREEN),
-    player->weaponowned[1] ? '2' : ' ',
-    player->weaponowned[2] ? '3' : ' ',
-    player->weaponowned[3] ? '4' : ' ',
-    player->weaponowned[4] ? '5' : ' ',
-    player->weaponowned[5] ? '6' : ' ',
-    player->weaponowned[6] ? '7' : ' ',
-    player->weaponowned[7] ? '8' : ' ',
-    player->weaponowned[8] ? '9' : ' '
-  );
+  if (grid)
+    snprintf(
+      str,
+      max_size,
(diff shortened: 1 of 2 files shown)


prboom2/src/hu_lib.c
```diff
@@ -145,6 +145,7 @@ void HUlib_drawTextLine
   int     i;
   int     w;
   int     x;
+  int     xbase;
```
kraflab: The question is whether there is any text that relies on the old behaviour. It's not exactly a bug, but that the old text start x was treated as an indentation for the first line, like in a paragraph.

prboom2/src/dsda/hud_components/weapon_text.c
```diff
@@ -20,32 +20,56 @@
...
+      player->weaponowned[5] ? '6' : ' ',
+      player->weaponowned[6] ? '7' : ' ',
+      player->weaponowned[7] ? '8' : ' ',
+      player->weaponowned[8] ? '9' : ' '
+    );
 }
 
 void dsda_InitWeaponTextHC(int x_offset, int y_offset, int vpt, int* args) {
   dsda_InitTextHC(&component, x_offset, y_offset, vpt);
+  grid = (dboolean) args[0];
```
kraflab: You can drop the cast.

prboom2/src/dsda/hud_components/weapon_text.c
```diff
@@ -20,32 +20,56 @@
...
-    player->weaponowned[5] ? '6' : ' ',
-    player->weaponowned[6] ? '7' : ' ',
-    player->weaponowned[7] ? '8' : ' ',
-    player->weaponowned[8] ? '9' : ' '
-  );
+  if (grid)
+    snprintf(
+      str,
+      max_size,
+      "W \x1b%c%c\x1b%c %c %c\n\x1b%cP\x1b%c %c %c %c\n\x1b%cN\x1b%c %c %c %c",
```
kraflab: Let's make this line slightly more digestible by breaking it up. I believe this is the syntax:
```c
"line 1\n"
"line 2\n"
"line 3",
```
If you're feeling fancy you could see how it looks to include the color reversion at the tail of the line instead of the start of the next line, so the W P N are all lined up at the start of the substrings.

prboom2/src/hu_lib.c
kraflab: I couldn't find anything relying on that behaviour, but I was also concerned about how this change interacts with negative x values, which cause right justification. That is, it would not be a complete fix since the text across lines would not be properly justified.

But I couldn't find anything actually using that, so I just cleaned it up entirely.

https://github.com/kraflab/dsda-doom/commit/09f2b33de8d6f9f41c3d8e4f3b46a1f7234a5d06
https://github.com/kraflab/dsda-doom/commit/214f6e534b43b6eef12e15fff20038c865b4cb56

You may want to rebase and just add the grid argument commit on top 😺

bkoropoff:
Pushed fixes

kraflab:
Thanks 👌

PR #214, opened 2023-02-19
Explicit rekkr support

bkoropoff:
Currently used to change ammo names in HUD component.

### DIFF
--- a/prboom2/src/dsda/global.c
+++ b/prboom2/src/dsda/global.c
@@ -144,6 +144,7 @@ const char* g_skyflatname;
 dboolean hexen = false;
 dboolean heretic = false;
 dboolean raven = false;
+dboolean rekkr = false;
 
 extern patchnum_t hu_font[HU_FONTSIZE];
 extern patchnum_t hu_font2[HU_FONTSIZE];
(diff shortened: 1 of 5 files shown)


kraflab:
There are good ideas in here but I don't believe the port should be updated for every wad that changes ammo (we can't single out rekkr alone). This is indicative of a need to define short ammo names. There are a lot of ways to do that but I haven't decided yet (in fact, I want to make all text in the port configurable for translation purposes as well). Whenever that happens you'll be able to just add an autoload lump for particular wads or embed it in the wad itself.

bkoropoff:
Rekkr is unique in that it is standalone (doesn't require doom.wad at all) and was blessed by Bethesda for the Unity port, so it could be viewed as a distinct official doom engine game like Heretic.  I agree that hard-coding ammo names for every wad would be too hacky.

Even if you don't decide to take the Rekkr change, do you still want the wad detection refactoring?

bkoropoff:
Pushed some fixes, let me know what parts you want, if any.

prboom2/src/dsda/global.c
```diff
@@ -609,31 +610,73 @@ static void dsda_InitHexen(void) {
...
+  const char* start;
+  const char* dot;
+  const char* fs;
+  const char* bs;
+  size_t len;
+  size_t baselen;
+
+  baselen = strlen(base);
+
+  // Find final path component
```
fabiangreffrath, line 624: Isn't this mostly a dup of https://github.com/kraflab/dsda-doom/blob/908f0135b0eb9614abe7f9b71ae04650a3032235/prboom2/src/d_main.c#L1251-L1264 ?

bkoropoff: Yeah, except for the extension removal (which catches `-iwad heretic`, etc.).  I'm not familiar with the code base so I didn't know that existed.

bkoropoff: If you want the function moved to a non-static, where should it go?

kraflab:
> it could be viewed as a distinct official doom engine game like Heretic

If it was distinct, it would be impossible to run it in dsda-doom. If it was official, it would be published by id software. So it's neither of those things. The custom gzdoom package of rekkr might be a distinct engine, but that version of rekkr isn't what we're talking about here.

The refactor doesn't look like it helps much in its current state, so I think we can drop this one for now :^)

PR #217, opened 2023-02-23
Add possibility to display map name in exhud

janchochol:
This is generally usefull for streamers, so it is clear on which level they currently are.


janchochol:
I am especially not sure about position of new element - any comments are welcomed.

kraflab:
As we already discussed on discord, I will be making a bunch of adjustments to the hud for 0.26. I will probably add map titles at some point but not quite like this. In general I suggest not creating a PR out of the blue like this to save yourself some time in the future, unless you have a strong insight for what I would do and when, or if we previously talked about the issue, or if it is a particularly simple / small task (in this case you made a deep change to an interface affecting many places in the code).

janchochol:
OK.

>  general I suggest not creating a PR out of the blue like this to save yourself some time in the future

This is not a problem for me - it does not take too much time, and I used to learn some things, so it is not wasted time, even if PR was closed.

But I definitely will ask next time.

PR #222, opened 2023-02-25
Bleedthrough fixes

### LINKED ISSUE #221: OpenGL rendering glitch in kdikdizd
Level 3 (actually map 15) uses flat bleeding to create deep water, but it is misrendered in GL:

doom00

There is actually a great deal of code in [prboom2/src/gl_missingtexture.c](https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/gl_missingtexture.c) and elsewhere that attempts to handle this case, but it's broken.  I think I have a patch to fix it.

bkoropoff:
Closes #221.

doom00

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -2193,10 +2193,11 @@ static void gld_AddFlat(int sectornum, dboolean ceiling, visplane_t *plane)
 
   indexed = V_IsWorldLightmodeIndexed();
 
+  if (plane->picnum & PL_SKYFLAT || plane->picnum == skyflatnum) // don't draw if sky
+    return;
+
   if (!ceiling) // if it is a floor ...
   {
-    if (sector->floorpic == skyflatnum) // don't draw if sky
-      return;
     // get the texture. flattranslation is maintained by doom and
     // contains the number of the current animation frame
     flat.gltexture=gld_RegisterFlat(flattranslation[plane->picnum], true, indexed);
@@ -2301,8 +2302,6 @@ static void gld_AddFlat(int sectornum, dboolean ceiling, visplane_t *plane)
   }
   else // if it is a ceiling ...
   {
-    if (sector->ceilingpic == skyflatnum) // don't draw if sky
-      return;
     // get the texture. flattranslation is maintained by doom and
     // contains the number of the current animation frame
     flat.gltexture=gld_RegisterFlat(flattranslation[plane->picnum], true, indexed);
--- a/prboom2/src/gl_missingtexture.c
+++ b/prboom2/src/gl_missingtexture.c
@@ -133,22 +133,22 @@ static void gld_PrepareSectorSpecialEffects(void)
 
           /* now mark the sectors that may require HOM bleed-through */
           if (needs_front_upper && side0->toptexture == NO_TEXTURE) {
-            side1->sector->flags |= MISSING_TOPTEXTURES;
+            side0->sector->flags |= MISSING_TOPTEXTURES;
             gld_RegisterBleedthroughSector(side1->sector,side0->sector,1);
           }
 
           if (needs_back_upper && side1->toptexture == NO_TEXTURE) {
-            side0->sector->flags |= MISSING_TOPTEXTURES;
+            side1->sector->flags |= MISSING_TOPTEXTURES;
             gld_RegisterBleedthroughSector(side0->sector,side1->sector,1);
           }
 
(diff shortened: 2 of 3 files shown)


prboom2/src/gl_main.c
```diff
@@ -2193,10 +2193,11 @@ static void gld_AddFlat(int sectornum, dboolean ceiling, visplane_t *plane)
 
   indexed = V_IsWorldLightmodeIndexed();
 
+  if (plane->picnum & PL_SKYFLAT || plane->picnum == skyflatnum) // don't draw if sky
```
kraflab, line 2196: Are we sure the PL_SKYFLAT bits are always there?

prboom2/src/gl_missingtexture.c
```diff
@@ -133,22 +133,22 @@ static void gld_PrepareSectorSpecialEffects(void)
 
           /* now mark the sectors that may require HOM bleed-through */
           if (needs_front_upper && side0->toptexture == NO_TEXTURE) {
-            side1->sector->flags |= MISSING_TOPTEXTURES;
+            side0->sector->flags |= MISSING_TOPTEXTURES;
```
kraflab, line 136: Why were the sides flipped before? Was it a bug?

prboom2/src/gl_missingtexture.c
```diff
@@ -171,53 +171,46 @@ static void gld_PrepareSectorSpecialEffects(void)
...
-      (
-        (ceiling && bleedsectors[source_idx].target->ceilingheight > target->ceilingheight) ||
-        bleedsectors[source_idx].target->floorheight < target->floorheight
-      )
-    )
-  )
-  {
-    bleedsectors[source_idx].target = target;
-  }
+  if (bleedsectors[idx].source == NULL ||
```
kraflab, line 201: This is equivalent right? I think that'll be easier to parse.
```c
!bleedsectors[idx].source ||
(ceiling && bleedsectors[idx].source->ceilingheight > source->ceilingheight) ||
bleedsectors[idx].source->floorheight < source->floorheight
```
By the way just looking at this...should it be `(!ceiling && bleedsectors[idx].source->floorheight < source->floorheight)` as well, or is it just not a symmetric condition?

prboom2/src/r_bsp.c
```diff
@@ -717,8 +717,13 @@ static void R_Subsector(int num)
          * a future patch by refactoring this into the renderer and tagging
          * visible candidate sectors during drawing.
          */
-        if (frontsector->floorheight >= viewz && (frontsector->flags & MISSING_BOTTOMTEXTURES))
-        {
+        if ((frontsector->floorheight >= viewz ||
+             viewz - frontsector->floorheight >= (500 << FRACBITS)) &&
```
kraflab: Could you add a constant like `FLOOR_BLEED_THRESHOLD` and put the fixme comment above a define for that?

prboom2/src/gl_main.c
bkoropoff: Yeah, this is the same logic used for detecting sky planes elsewhere, e.g. in `R_FindPlane`.

prboom2/src/gl_missingtexture.c
bkoropoff: Yes, as far as I can tell.  I don't see how it could have worked before.

prboom2/src/gl_missingtexture.c
bkoropoff: The original had the same asymmetry, but I think you're right that it's a bug.

kraflab:
Are there some other bleedthrough cases you're aware of that we can double check are still working?

bkoropoff:
The pool of water with a backpack sitting in it in MAP02 of TNT

prboom2/src/gl_main.c
kraflab: There is a texture flag for skies as well, and that flag is not always set for sky textures. I learned to never trust anything in this code base, no matter how obvious 😄

bkoropoff: If you look at the logic in `R_Subsector`, it passes `frontsector->sky` to `R_FindPlane` if and only if it has `PL_SKYFLAT` set and the floor/ceiling is `skyflatnum`, and just `floorpic`/`ceilingpic` otherwise.  `R_FindPlane` sets this in the resulting `visplane_t` verbatim.  So `PL_SKYFLAT` should always be set for sectors with non-standard skies.

kraflab:
Thanks 🥳

PR #223, opened 2023-02-25
Attempt to repair unclosed sectors to fix GL rendering glitches

### LINKED ISSUE #216: OpenGL rendering glitch in Alien Vendetta (av.wad)
A self-referencing sector used to create deep water in MAP01 shows a HOM effect:

doom00

The outer sector (which should be rendered) fails to be tessellated in `gl_preprocess.c` because it is not closed:

```
sector 36 is not closed at vertex 188 ! 10 lines in sector
sector 36 is not closed at vertex 189 ! 10 lines in sector
sector 36 is not closed at vertex 190 ! 10 lines in sector
sector 36 is not closed at vertex 191 ! 10 lines in sector
```

Candidate lines that would close it are actually tagged with the inner self-referencing sector (37), e.g. the highlighted one here:

Screenshot from 2023-02-23 00-09-03

bkoropoff:
Closes #216 

doom00

This ended up being kind of complicated, and I'm not sure it's the best approach.  The problem is as follows:

- A self-referencing sector is "invisible" to the software renderer because no linedef encountered during BSP traversal has a sidedef referencing it.  This causes the containing sector to be rendered instead.
- This normally works fine in dsda-doom GL, because the outer sector will be tessellated (converted to triangles) without regard for the self-referencing sector, and the same BSP traversal logic is used to build the scene.
- However, tessellation can fail if the containing sector is not closed.  In this case, dsda-doom falls back on tessellating all the subsectors of the sector.  Since these are guaranteed to be closed convex polygons, this always succeeds.
- Although the self-referencing sector is invisible to the renderer, the node builder will dutifully "carve" it out of the containing sector when generating subsectors
- This results in only the difference between the inner and outer sector being tessellated, leaving a hole that gets filled with HOM
- Mappers are not particularly careful about closing all their sectors since the software renderer is so forgiving, so an unclosed sector surrounding a self-referencing sector can happen (AV map 01)

Rather than trying to make the renderer/BSP traversal aware of self-referencing sectors (it's complicated code I'm not familiar with), I made the sector tessellation logic "try harder" to tessellate sectors.  It uses a search algorithm on connected lines to try to find a way to close the sector.


kraflab:
Tbh, I'm not sure I'm willing to commit this much code just to fix that bug, although there may be nothing wrong with the solution 😅

bkoropoff:
Yeah, the code ended up being fairly elaborate.  Some of the refactoring is nice, and it might fix other rendering problems arising from unclosed sectors that have yet to be caught, but the size of the change makes it riskier, which is why I've marked it as a draft for now.  I've been trying to think of a more surgical fix for just the precise problem in av.wad.

bkoropoff:
Some maps have extremely highly connected geometry that can cause an explosion in the search space, so I'll have to add iteration limits in addition to depth limits.

bkoropoff:
Closed in favor of #225, but this code can be dusted off if some other glitch due to unclosed sectors shows up.

Issue #220, opened 2023-02-24
Rendering error on first frame (pre-melt) in opengl


kraflab:
Occasionally (not always but not super rare either), opengl has rendering errors on the pre-melt frame. The problem is frozen in place as the background during the melt, but is resolved as soon as the map starts. This happens both on restarts and during continuous play.

![opengl-melt](https://user-images.githubusercontent.com/25529060/221205064-8af5491f-4c29-4b57-a2ce-31d0acc7d78f.jpg)

kraflab:
@JadingTsunami another one of these, maybe you have an idea 😄 
I don't know if it's a symptom of the same problem that affected decino, but this one is specifically about the render behind the melt effect.

JadingTsunami:
Hmm, unfortunately I've never seen this one. What hardware is in use? A long shot, but any chance it's known whether it's limited to a particular hardware type?

JadingTsunami:
Ha, never mind. I see it. I'll fix this in a bit.

kraflab:
Fixed by https://github.com/kraflab/dsda-doom/pull/224

PR #224, opened 2023-02-25
Supply correct texture format for GL wipe.

JadingTsunami:
Reported in #220. 

The call to `glTexImage2D` in `gl_wipe.c` used an incorrect internal image format. This should fix it.


kraflab:
🥳

PR #225, opened 2023-02-25
gl_preprocess: handle more self-referencing sectors

### LINKED ISSUE #216: OpenGL rendering glitch in Alien Vendetta (av.wad)

bkoropoff:
When a sector can't be tessellated as a whole because it's not closed, the GL renderer falls back on subsector tesselation. If the sector in question contains self-referencing sectors (used for deep water and other tricks), these won't be tesselated, causing HOM.

Explicitly detect self-referencing sectors and what sector contains them (i.e. what should be rendered in their place), and account for them during subsector tessellation.

Closes #216 

doom00

### DIFF
--- a/prboom2/src/gl_preprocess.c
+++ b/prboom2/src/gl_preprocess.c
@@ -41,6 +41,7 @@
 #define CALLBACK
 #endif
 
+#include <assert.h>
 #include <math.h>
 
 #include "gl_opengl.h"
@@ -243,6 +244,7 @@ static void gld_FlatConvexCarver(int ssidx, int num, divline_t *list)
       if (flats_vbo)
       {
         int currentsector=ssec->sector->iSectorID;
+        sector_t* container = ssec->sector->gl_pp;
         GLLoopDef **loop;
         int *loopcount;
 
@@ -250,6 +252,12 @@ static void gld_FlatConvexCarver(int ssidx, int num, divline_t *list)
         {
           loop = &subsectorloops[ ssidx ].loops;
           loopcount = &subsectorloops[ ssidx ].loopcount;
+        }
+        else if (container != NULL)
+        {
+          // Give triangles to ultimate container of self-referencing sector
+          loop = &sectorloops[container - sectors].loops;
+          loopcount = &sectorloops[container - sectors].loopcount;
         }
         else
         {
@@ -295,8 +303,13 @@ static void gld_CarveFlats(int bspnode, int numdivlines, divline_t *divlines)
     // the partition lines that carve out the subsector.
     // special case for trivial maps (no nodes, single subsector)
     int ssidx = (numnodes != 0) ? bspnode & (~NF_SUBSECTOR) : 0;
+    subsector_t* ssec = &subsectors[ssidx];
+    sector_t* sec = ssec->sector;
+    sector_t* cont = sec->gl_pp;
 
-    if (!(subsectors[ssidx].sector->flags & SECTOR_IS_CLOSED) || triangulate_subsectors)
+    if (!(sec->flags & SECTOR_IS_CLOSED) ||
+        (cont != NULL && !(cont->flags & SECTOR_IS_CLOSED)) ||
+        triangulate_subsectors)
       gld_FlatConvexCarver(ssidx, numdivlines, divlines);
     return;
   }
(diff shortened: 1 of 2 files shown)


prboom2/src/gl_preprocess.c
```diff
@@ -695,6 +706,78 @@ static void gld_PrecalculateSector(int num)
   Z_Free(lineadded);
 }
 
+static dboolean gld_SectorIsSelfReferencing(sector_t* sector)
+{
+  int i;
+
+  for (i = 0; i < sector->linecount; ++i) {
```
kraflab: I realize it's a bit annoying because the style is different in different files, but please put the `{` below the line here (and in the other cases).

kraflab:
This looks a lot better for sure 😄

prboom2/src/gl_preprocess.c
```diff
@@ -695,6 +706,83 @@ static void gld_PrecalculateSector(int num)
...
+    // We're self-referencing, but we couldn't find another sector across
+    // a self-referencing linedef.  It appears that this can happen due
+    // to the way the node builder generates subsectors, causing
+    // R_PointInSubsector not to find what we're looking for
+    // (e.g. Alien Vendetta MAP01 sector 39). Look for a neighboring
+    // self-referencing sector and point at that instead.
+    for (i = 0; i < sector->linecount; ++i)
+    {
+      line_t* l = sector->lines[i];
+      if (l->sidenum[0] != NO_INDEX &&
```
kraflab: Can we check `l->frontsector` and `l->backsector` directly?

bkoropoff: The existing code was not consistent about using it, but yeah, it looks equivalent.

prboom2/src/gl_preprocess.c
```diff
@@ -695,6 +706,83 @@ static void gld_PrecalculateSector(int num)
...
+      if (s1 != NULL && s1 != sector)
+        return s1;
+      if (s2 != NULL && s2 != sector)
+        return s2;
+    }
+  }
+
+  if (selfref)
+  {
+    // We're self-referencing, but we couldn't find another sector across
```
bkoropoff: I think this part of the logic is susceptible to false positives, so I'm going to rework it

bkoropoff:
I reworked the logic a bit to be more conservative.  It involves a pre-pass to decide which sectors are "real" (i.e. non-self-referencing), but this makes false positives exceptionally unlikely, and simplifies the rest.

bkoropoff:
Now with even more paranoia.

prboom2/src/gl_preprocess.c
```diff
@@ -695,6 +705,163 @@ static void gld_PrecalculateSector(int num)
...
+    if (sector->lines[i]->v1 == line->v1 || sector->lines[i]->v2 == line->v1)
+      v1c++;
+    if (sector->lines[i]->v2 == line->v2 || sector->lines[i]->v2 == line->v2)
+      v2c++;
+  }
+
+  // Line is still real if:
+  // - It's isolated (no neighbors), i.e. it's floating mid-sector
+  // - It has more than 1 neighbor per vertex, i.e. it divides the sector
+  return (v1c == 0 && v2c == 0) || (v1c > 1 && v2c > 1);
```
kraflab: Should it just be `v1c != 1 && v2c != 1`? I.e., what about a line that has more than one neighbor on one vertex and no neighbors on the other vertex?

prboom2/src/gl_preprocess.c
```diff
@@ -695,6 +705,163 @@ static void gld_PrecalculateSector(int num)
...
+      if (!gld_IsRealLine(&sectors[i], sectors[i].lines[j])) {
+        real = false;
+        break;
+      }
+    }
+    if (real)
+      sectors[i].flags |= SECTOR_IS_REAL;
+  }
+
+  // Now mark all transitively real sectors in passes (to avoid recursive calls)
```
kraflab: I'm concerned about a loop that may check tens of thousands of sectors repeatedly 😅

kraflab:
Well, I am once again not sure I want to commit this just to fix that bug 😅

prboom2/src/gl_preprocess.c
bkoropoff: Yeah, good point

prboom2/src/gl_preprocess.c
bkoropoff: The vast majority of sectors will be tagged by the first loop, meaning this loop will run at most once or twice in practice.  This does have worst case N^2 runtime, but it would only occur on a map which consists of N sectors chained end-to-end, with the first N-1 being self-referencing.  That's not a very realistic map.

This loop could be removed, but I was trying to be extremely paranoid about treating a sector as self-referencing when it isn't.

bkoropoff: I'll rework this to use an explicit queue and avoid visiting unnecessary sectors.

bkoropoff:
I think this is fairly straightforward compared to some of the other hacks to support software rendering quirks that are already in the code (and don't actually cover all cases, as I've found...)

kraflab:
Alright, looks great now 👍 
Thanks!

PR #229, opened 2023-02-26
gl_missingtexture: more bleed fixing/refactoring

bkoropoff:
My previous PR broke a sector in Hell Revealed MAP01.  I realized that bleedthrough is fundamentally bi-directional depending on player view height, so I fixed and simplified the code.  Tested on kdikdizd MAP15, TNT MAP 02, HR MAP01.

---

- Don't assume side0 is the front side (facing into current sector)
- Always register bleedthrough bidirectionally, as both cases can happen depending on the view:
  * A floor or ceiling could be invisible because it is above/below the view height
  * A floor or ceiling could be invisible because it is so high/low that it is fully occluded from view
- Since all sectors are visited, avoid duplicate work on each iteration
- Refactor/reformat code a little

### DIFF
--- a/prboom2/src/gl_missingtexture.c
+++ b/prboom2/src/gl_missingtexture.c
@@ -67,7 +67,7 @@ typedef struct
 {
   sector_t *source; // The sector to receive a fake bleed-through flat
   sector_t *target; // The floor sector whose properties should be copied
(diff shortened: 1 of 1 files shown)


kraflab:
👍

PR #230, opened 2023-02-26
Blood: The Bleedening

bkoropoff:
Fixes Sunder 2047 MAP31.  Tested on that and all the previous maps.

----

Distinguish occlusion and non-occlusion bleeding when registering and looking up sectors.  Avoids false-positives.


kraflab:
👌

PR #233, opened 2023-02-26
gl: clip bottom texture at ceiling

### LINKED ISSUE #196: Some OpenGL rendering glitch
The wall that should hide the lowering bar in this area is glitchy and just straight up disappear, revealing parts of the bar that shouldn't be visible

WAD & Map : [Sunder 2407.wad](https://www.dropbox.com/s/vi47z1a4e4c4980/Sunder%202407.zip?dl=1), MAP31

Here is what it looks like in 0.24.3

And here is in 0.25.4

[Here is also a video of the bug in action](https://www.youtube.com/watch?v=OLGKBy6lJHY)

The lighting mode for both instances are the Shaders mode

bkoropoff:
Closes #196 

-----

This avoids Z fighting when the sector's floor is higher than its ceiling.

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -2062,6 +2062,13 @@ bottomtexture:
         if (seg->linedef->flags & ML_DONTPEGBOTTOM)
           specific_rowoffset += frontsector->ceilingheight - floor_height;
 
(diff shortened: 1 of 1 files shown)


kraflab:
Thanks 👍

PR #234, opened 2023-02-27
Add support for non-Latin paths on Windows

rfomin:
I did not change the `data/rd_*` files, and there is `fopen()` under the `#TEST` macros in `midifile.c`.

Alternative to #204

### DIFF
--- a/prboom2/src/m_file.c
+++ b/prboom2/src/m_file.c
@@ -83,7 +83,7 @@ static wchar_t *ConvertMultiByteToWide(const char *str, UINT code_page) {
     return NULL;
   }
 
-  wstr = malloc(sizeof(wchar_t) * wlen);
+  wstr = Z_Malloc(sizeof(wchar_t) * wlen);
 
   if (!wstr) {
     lprintf(LO_INFO, "ConvertMultiByteToWide: Failed to allocate new string\n");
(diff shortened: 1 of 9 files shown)


prboom2/src/m_file.c
```diff
@@ -65,22 +70,238 @@
...
+  wchar_t *wstr = NULL;
+
+  wstr = ConvertSysNativeMBToWide(str);
+
+  if (!wstr)
+    return NULL;
+
+  ret = ConvertWideToUtf8(wstr);
+
+  free(wstr);
```
kraflab: Could you use zone functions for all the frees and allocations?

rfomin: Then we also need to use the zone memory in `i_glob.c` Is this necessary?

kraflab: Yes, otherwise we'll be using mixed memory management, which sounds like a recipe for future bugs 😄

rfomin: Fixed in https://github.com/kraflab/dsda-doom/pull/234/commits/db2a68cf318320ce2ab36805d8645cf3040967d9

kraflab:
This is awesome stuff, thanks! 🥳

PR #237, opened 2023-02-27
r_bsp: add symmetric logic for ceiling occlusion

### LINKED ISSUE #236: GL Rendering Glitch in BSTX1 MAP01
Also not my fault
doom01

bkoropoff:
I didn't bother with this before because you would think there's no point in having a fake "deep" ceiling.  But BTSX1 has one for no discernible reason.

Closes #236 

----

No idea why mappers would do it, but they did.

### DIFF
--- a/prboom2/src/r_bsp.c
+++ b/prboom2/src/r_bsp.c
@@ -50,6 +50,8 @@
 // bleedthrough normally occurs when the sector's floor is so
 // low that it is completely occluded from the current view.
 #define FLOOR_BLEED_THRESHOLD 500
+// Same, but for ceiling
+#define CEILING_BLEED_THRESHOLD 500
 
 int currentsubsectornum;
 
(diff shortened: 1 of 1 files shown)


prboom2/src/r_bsp.c
```diff
@@ -732,6 +734,7 @@ static void R_Subsector(int num)
           if (frontsector->floorheight >= viewz)
             tmpsec = GetBestBleedSector(frontsector, BLEED_NONE);
           if (tmpsec == NULL &&
+              viewz > frontsector->floorheight &&
```
kraflab: Isn't this line redundant? If the difference is greater than the floor bleed threshold it must also be nonzero.

prboom2/src/r_bsp.c
```diff
@@ -751,9 +754,15 @@ static void R_Subsector(int num)
...
 
-        if (frontsector->ceilingheight <= viewz && (frontsector->flags & MISSING_TOPTEXTURES))
+        if (frontsector->flags & MISSING_TOPTEXTURES)
         {
-          tmpsec = GetBestBleedSector(frontsector, 1);
+          tmpsec = NULL;
+          if (frontsector->ceilingheight <= viewz)
+            tmpsec = GetBestBleedSector(frontsector, BLEED_CEILING);
+          if (tmpsec == NULL &&
+              viewz < frontsector->ceilingheight &&
```
kraflab: Same question here

prboom2/src/r_bsp.c
bkoropoff: Yeah, I was thinking in terms of unsigned math where you need to avoid overflow.

kraflab:
Thanks again 😸

PR #238, opened 2023-02-27
gl_preprocess: fix some logic bugs

bkoropoff:


prboom2/src/gl_preprocess.c
```diff
@@ -922,7 +922,7 @@ static void gld_GetSubSectorVertices(void)
   {
     subsector_t* ssector = &subsectors[i];
 
-    if (gld_TriangulateSubsector(ssector))
+    if (!gld_TriangulateSubsector(ssector))
```
kraflab, line 925: What happened here? Btw the comment on this method says it only applies to glnodes, do you have a test map for this?

bkoropoff: This got inverted during the refactoring that created `gld_TriangulateSubsector`.  I just happened to notice it while looking at something else.  I don't have any test map specifically for this code path.

kraflab: Originally:
```c
    if (!(sec->flags & SECTOR_IS_CLOSED) ||
        (cont != NULL && !(cont->flags & SECTOR_IS_CLOSED)) ||
        triangulate_subsectors)
```

```c
static dboolean gld_TriangulateSubsector(subsector_t *ssec)
{
  sector_t *container = ssec->sector->gl_pp;
  return !(ssec->sector->flags & SECTOR_IS_CLOSED) ||
         (container && !(container->flags & SECTOR_IS_CLOSED)) ||
         triangulate_subsectors;
}
```
Am I missing something?

bkoropoff: Huh.  Well, `continue` clearly skips triangulating the current subsector, so I guess the original logic was also reversed.

kraflab: Strange 😅 
But let's leave it as it was before then, unless an issue comes up.

bkoropoff:
This will get subsumed by #239

PR #239, opened 2023-02-28
Better sector tests

### LINKED ISSUE #235: GL rendering glitch in Hell Revealed II MAP01
I thought I caused this somehow, but it happens on 0.25.6 and 0.25.5 as well.
doom00

bkoropoff:
Closes #235

Opening as draft because I still need to convince myself the algorithm is correct and test more.  I did try a huge Sunder map to confirm the performance is OK with sectors with huge numbers of lines.

The real problem this is fixing is that the existing check for sector closedness in the code is... just wrong.  It marks sectors as closed when they aren't, then attempts GLU tessellation when it can't possibly generate any useful triangles.  Fixing the test makes it fall back on subsector tessellation as it should and fixes the bug.

Unfortunately, fixing the test means some self-referencing sectors become visible when they shouldn't be, because they were accidentally being made invisible by the bug!  So the test for self-referencing sectors needed to be fixed/improved to catch more cases.

Fixing anything in this renderer is a game of whack-a-mole.

### DIFF
--- a/prboom2/src/gl_preprocess.c
+++ b/prboom2/src/gl_preprocess.c
@@ -750,7 +750,7 @@ static dboolean gld_IsRealLine(sector_t* sector, line_t* line)
       continue;
     if (sector->lines[i]->v1 == line->v1 || sector->lines[i]->v2 == line->v1)
       v1c++;
-    if (sector->lines[i]->v2 == line->v2 || sector->lines[i]->v2 == line->v2)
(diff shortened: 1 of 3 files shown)


bkoropoff:
It's very difficult to get all cases working at the same time within the current GL renderer framework.  I'm going to submit a PR to revert out my hacks so far and go back to the drawing board.

PR #240, opened 2023-03-01
gl_preprocess: revert out self-ref sector hacks

bkoropoff:
I think it's going to take a more radical approach to fix self-referencing sectors.  Putting things back the way they were while I do some experimenting.

### DIFF
--- a/prboom2/src/gl_preprocess.c
+++ b/prboom2/src/gl_preprocess.c
@@ -90,18 +90,12 @@ static void gld_TurnOffSubsectorTriangulation(void)
 
 static dboolean gld_TriangulateSubsector(subsector_t *ssec)
 {
-  sector_t *container = ssec->sector->gl_pp;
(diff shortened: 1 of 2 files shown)


kraflab:
So this undoes the self referencing improvements, but the bleed fixes are still there right? So av01 is broken again but kdikdizd is still fixed? Just making sure I know where this puts us 😄

kraflab:
I thought this code was still an improvement, and didn't introduce any new bugs. Is that not the case? Just thinking if that's true it doesn't need to be reverted on master until something else is ready.

bkoropoff:
Yeah, the bleeding fixes seem fine.  The current self-ref detection code is conceptually broken (e.g. the test for a line being interior or exterior is obviously wrong in retrospect).  I'm not sure when you're planning to do another release, so I didn't want to leave known-incorrect code in master.

kraflab:
Alright then, let's clean it up for the time being.

PR #241, opened 2023-03-02
Self-ref sector fixes (and more)

bkoropoff:
Summary of the problems leading to this point:

1. HR2 MAP01 has a bad rendering glitch (issue #235)
2. It's caused by a broken test for sector closedness that assumes linedef side 0 was facing the sector.  Since it falsely thinks the sector is closed, it uses GLU tessellation, which generates no triangles and results in HOM.
3. Fixing the broken test causes it to fall back on subsector tessellation, which fixes the glitch in combination with tweaking a bleeding threshold.
4. Fixing the broken test also opens a can of worms: a lot of self-referencing sectors were *accidentally* being rendered correctly because of it, including some in HR2 MAP01 and kdikdizd.  This was mostly dumb luck:
    - Trying to tessellate an unclosed sector can generate no triangles.  If the unclosed sector was self-referencing, this caused it to be invisible (as it should be)
    - Trying to tessellate an unclosed sector can succeed anyway because it has a closed subset.  This can fill in a self-referencing sector contained within it (also good)
5. Fixing the above requires making the GL rendering path act like the software path by not rendering flats of self-referencing sectors, and restoring self-ref sector tests in `gl_preprocess.c` so that their subsectors can "fill in" their parents.

This was the only way I could get all of AV MAP01, HR MAP01, HR2 MAP01, AVMAP01, and kdikdizd MAP15 to render correctly at the same time.  The new self-ref sector test is very precise by taking a graph-theoretic approach.  I guess it always comes back to graph search.

Note that the sector closedness test and GLU tessellation logic are also doing similar but more ad-hoc graph searches to find the sector contours.  These could be subsequently refactored to use results from the initial search.

### DIFF
--- a/prboom2/src/r_bsp.c
+++ b/prboom2/src/r_bsp.c
@@ -49,9 +49,9 @@
 // an approximation of software rendering behavior.  Presumably
 // bleedthrough normally occurs when the sector's floor is so
 // low that it is completely occluded from the current view.
-#define FLOOR_BLEED_THRESHOLD 500
+#define FLOOR_BLEED_THRESHOLD 400
 // Same, but for ceiling
-#define CEILING_BLEED_THRESHOLD 500
+#define CEILING_BLEED_THRESHOLD 400
 
 int currentsubsectornum;
 
--- a/prboom2/src/r_bsp.c
+++ b/prboom2/src/r_bsp.c
@@ -375,6 +375,18 @@ static void R_AddLine (seg_t *line)
 
   if (V_IsOpenGLMode())
   {
+    line_t* l = line->linedef;
+    sector_t* sec = subsectors[currentsubsectornum].sector;
+
+    // Don't add plane to drawing list until we encounter a
+    // non-self-referencing linedef in the subsector.
+    if (sec->gl_validcount != validcount && (l == NULL || l->frontsector != l->backsector))
+    {
+      sec->gl_validcount = validcount;
+
+      gld_AddPlane(currentsubsectornum, floorplane, ceilingplane);
+    }
+
     angle1 = R_PointToPseudoAngle(line->v1->x, line->v1->y);
     angle2 = R_PointToPseudoAngle(line->v2->x, line->v2->y);
 
@@ -619,7 +631,7 @@ static void R_Subsector(int num)
   sub = &subsectors[num];
   currentsubsectornum = num;
 
-  if (V_IsSoftwareMode() || !gl_use_stencil || sub->sector->validcount != validcount)
+  if (V_IsSoftwareMode() || sub->sector->gl_validcount != validcount)
   {
     frontsector = sub->sector;
 
@@ -667,7 +679,7 @@ static void R_Subsector(int num)
   // e6y
   // New algo can handle fake flats and ceilings
   // much more correctly and fastly the the original
-    if (V_IsOpenGLMode())
+    if (V_IsOpenGLMode() && sub->sector->gl_validcount != validcount)
     {
       // check if the sector is faked
       sector_t *tmpsec = NULL;
@@ -797,9 +809,6 @@ static void R_Subsector(int num)
     sub->sector->validcount = validcount;
 
     R_AddSprites(sub, (floorlightlevel+ceilinglightlevel)/2);
-
-    if (V_IsOpenGLMode())
-      gld_AddPlane(num, floorplane, ceilingplane);
   }
 
   // hexen
--- a/prboom2/src/gl_preprocess.c
+++ b/prboom2/src/gl_preprocess.c
@@ -90,12 +90,18 @@ static void gld_TurnOffSubsectorTriangulation(void)
 
 static dboolean gld_TriangulateSubsector(subsector_t *ssec)
 {
+  sector_t *container = ssec->sector->gl_pp;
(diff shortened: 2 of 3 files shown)


kraflab:
"assumes linedef side 0 was always the front side"
Isn't side 0 the front by definition? Or do you mean "the side facing the sector" or something like that?

bkoropoff:
Yeah, the side facing the current sector.

prboom2/src/r_bsp.c
```diff
@@ -794,12 +809,7 @@ static void R_Subsector(int num)
 
   if (sub->sector->validcount != validcount)
   {
-    sub->sector->validcount = validcount;
```
kraflab, line 797: Why is this line gone? Does this mean we'll be calling R_AddSprites more than we should?

bkoropoff: I'll refactor this a bit to avoid messing with validcount in multiple places

prboom2/src/r_bsp.c
```diff
@@ -619,7 +631,7 @@ static void R_Subsector(int num)
   sub = &subsectors[num];
   currentsubsectornum = num;
 
-  if (V_IsSoftwareMode() || !gl_use_stencil || sub->sector->validcount != validcount)
+  if (V_IsSoftwareMode() || sub->sector->gl_validcount != validcount)
```
kraflab, line 634: So what exactly is going on in this function 🤔 
Why does gl only want to do this once, but for software it happens each time? Or does software only reach this once per sector?

prboom2/src/r_bsp.c
```diff
@@ -667,7 +679,7 @@ static void R_Subsector(int num)
   // e6y
   // New algo can handle fake flats and ceilings
   // much more correctly and fastly the the original
-    if (V_IsOpenGLMode())
+    if (V_IsOpenGLMode() && sub->sector->gl_validcount != validcount)
```
kraflab, line 682: Why was this happening every time before and now should only happen once?

prboom2/src/r_bsp.c
```diff
@@ -375,6 +375,18 @@ static void R_AddLine (seg_t *line)
...
   if (V_IsOpenGLMode())
   {
+    line_t* l = line->linedef;
+    sector_t* sec = subsectors[currentsubsectornum].sector;
+
+    // Don't add plane to drawing list until we encounter a
+    // non-self-referencing linedef in the subsector.
+    if (sec->gl_validcount != validcount && (l == NULL || l->frontsector != l->backsector))
+    {
+      sec->gl_validcount = validcount;
```
kraflab, line 385: I'm a bit concerned about this decoupling. There are sector-related things in R_Subsector we want to only happen once, but that once is contingent on the conditions here, which may not be true. Doesn't that mean we may execute code from R_Subsector multiple times?

prboom2/src/r_bsp.c
bkoropoff: I think this could be changed to only do it once per sector for software mode as well, but I'd have to test it.  It's basically doing a bunch of hash lookups for no reason after the first subsector in a sector.

prboom2/src/r_bsp.c
bkoropoff: It was never necessary to do it every time before, and operating on potentially stale values of `floorplane/ceilingplane` made me nervous.

prboom2/src/r_bsp.c
bkoropoff: `R_Subsector` only uses `validcount` to avoid adding sprites multiple times per sector.  It should also probably use it in software mode when creating visplanes as I mentioned above (edit: wrong, see other comment), but that's harmless because visplane creation is idempotent.

Otherwise, everything else in `R_Subsector` really needs to happen for every subsector, namely drawing wall segs through this function.  So `gl_validcount` just ensures we don't call `gld_AddPlane` twice for a sector, which is important because it's *not* idempotent like software visplane lookup is.

prboom2/src/r_bsp.c
kraflab: Those plane globals need to be updated any time the sector changes, and the order subsectors are processed is probably not guaranteed to be contiguous by sector, so that may be why it has to be redone.

kraflab: But actually, wouldn't that be the same for gl then? 🤔 I guess not because the plane only gets added once, and when that happens the gl_validcount changes?

bkoropoff: Ah, right.  `floorplane` and `ceilingplane` are always needed for software because they get used by `R_StoreWallRange` in `r_segs.c`, but GL only needs them when it could potentially call `gld_AddPlane`.  So this can be left as-is.

kraflab:
I'm going to do a small refactor in r_bsp just to make it easier for me to think about what's going on there. One sec 😅

kraflab:
Ok, please adopt your r_bsp changes with this as a base: https://github.com/kraflab/dsda-doom/commit/22157589981f02eee81a9bf1e6f83eaad0935ed7

bkoropoff:
Everything up to the BSP changes is now in #243.

I'll submit the self-ref sector stuff on top of it once I'm done tweaking it some more.

PR #243, opened 2023-03-04
Fix HR2 MAP 01 (standalone)

bkoropoff:
I figured out how to tweak the sector closedness check to fix this map without causing any regressions (that I know of), so I'm submitting this alone while I tweak the self-ref sector code some more.


kraflab:
Looks clean to me, thanks!

PR #244, opened 2023-03-04
Restore self-ref sector checks, clean up preprocessing

bkoropoff:
This restores the self-referencing sector detection logic.  Compared to past revisions, it:

- Is (mostly) graph-theoretic
- Is fairly well factored (e.g. sector traversal is generic)
- Use cycle directionality and BSP tests to determine which lines are interior (cycle length was a dead end)
- Operates on subsectors to prevent sector tags shared across distant geometry from creating glitches

The second commit refactors tessellation and a few odds and ends to happen in the same traversal as
self-reference detection.  This makes things a lot cleaner (the inner tessellation loop is actually legible now), but it's obviously a larger change.

### DIFF
--- a/prboom2/src/r_defs.h
+++ b/prboom2/src/r_defs.h
@@ -293,7 +293,7 @@ typedef byte r_flags_t;
 #define RF_BOT_TILE 0x04 // Lower texture needs tiling
 #define RF_IGNORE   0x08 // Renderer can skip this line
 #define RF_CLOSED   0x10 // Line blocks view
-#define RF_ISOLATED 0x20 // Isolated line
+#define RF_LINKED   0x20 // Linked to at least one other line by a vertex
 #define RF_REAL     0x40 // Real (not self-referencing trick) line
 
 typedef enum
--- a/prboom2/src/r_defs.h
+++ b/prboom2/src/r_defs.h
@@ -81,6 +81,9 @@ typedef struct
   // they are *only* used in rendering
   fixed_t px;
   fixed_t py;
+  // dsda/traverse.c
+  uint16_t id:15;
+  dboolean visited:1;
 } vertex_t;
 
 // Each sector has a degenmobj_t in its center for sound origin purposes.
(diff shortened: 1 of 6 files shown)


andrey-budko:
perfect on hr2.wad, thank you, but something went wrong on requiem.wad map13

image

bkoropoff:
Thanks for the pointer, I'll take a look

andrey-budko:
freezes on hr2final map12 and map19

bkoropoff:
OK, I know the problem with requiem 13.  The sector closedness logic is too optimistic.

bkoropoff:
> perfect on hr2.wad, thank you, but something went wrong on requiem.wad map13

By the way, the fix for hr2 is already in master.  This is to fix AV map01.

andrey-budko:
> By the way, the fix for hr2 is already in master. This is to fix AV map01.

I forgot about hr2final and tested on outdated hr2.wad, and there was a difference:
<details>

 image  
</details>

Also you can check sector280@map11 on this wad
<details>

image
</details>

https://prboom-plus.sourceforge.io/HR2.WAD.zip

bkoropoff:
I pushed a quick fix for the disappearing flats problem.  The real issue is the hangs on some maps, which is due to needing a cycle enumeration algorithm with better time complexity.  I know how to implement one, it's just kind of obnoxious in C.

prboom2/src/r_defs.h
```diff
@@ -351,6 +353,12 @@ typedef struct line_s
   int healthgroup;
   const byte* tranmap;
   float alpha;
+
+  // gl_preprocess
+  dboolean tess:1;
```
kraflab: Please reuse the `r_flags` field if we're adding more temporary flags like this. You'll have to make `r_flags_t` an `unsigned short` instead of a `byte` in that case.

bkoropoff:
I have a known tessellation problem affecting sector 86 in requiem map 13, but this is starting to come together.  I can probably submit a PR for just the first commit and have the refactoring land later after baking more.

bkoropoff:
Nothing is ever easy

bkoropoff:
I misread the paper for the algorithm I'm using as running in O(n + e).  It's actually that *between each cycle in the graph*, i.e. O((n + e)(c + 1)).  So it completely chokes on HR2 MAP12 because it has a structure with ridiculous cyclic complexity and it takes a million years to compute.

Back to the drawing board.

PR #248, opened 2023-03-05
gl_preprocess: restore self-ref sector detection

bkoropoff:
This does not include the refactoring to handle GLU tessellation using the same traversal code, which has a corner case I need to work out first  (the existing code seems to handle them with some weird angle checking which doesn't make much sense to me).

I have completely implemented Donald Johnson's cycle search algorithm for good time complexity, with extra logic to avoid hangs like on HR2 MAP12.  There appears to be no way to handle self-referencing sectors that doesn't involve a pile of graph search code. :upside_down_face: 

-------------------------------------

- Use graph-based approach
- Detect containers at the level of subsectors to avoid artifacts from distant geometry sharing sector tags

### DIFF
--- a/prboom2/src/r_defs.h
+++ b/prboom2/src/r_defs.h
@@ -71,7 +71,7 @@
 // Note: transformed values not buffered locally,
 // like some DOOM-alikes ("wt", "WebView") do.
 //
-typedef struct
+typedef struct vertex_s
 {
   fixed_t x, y;
   // [crispy] remove slime trails
@@ -81,6 +81,11 @@ typedef struct
   // they are *only* used in rendering
   fixed_t px;
   fixed_t py;
+  // dsda/traverse.c
+  struct vertex_s* prev;
+  struct vertex_s* root;
+  unsigned short id;
+  dboolean visited;
 } vertex_t;
 
 // Each sector has a degenmobj_t in its center for sound origin purposes.
@@ -117,6 +122,7 @@ typedef struct
 #define SECF_LIGHTCEILINGABSOLUTE  0x00080000
 #define SECF_DAMAGEFLAGS (SECF_ENDGODMODE|SECF_ENDLEVEL|SECF_DMGTERRAINFX|SECF_HAZARD|SECF_DMGUNBLOCKABLE)
 #define SECF_TRANSFERMASK (SECF_SECRET|SECF_WASSECRET|SECF_DAMAGEFLAGS|SECF_FRICTION|SECF_PUSH)
+#define SECTOR_IS_REAL             0x00200000
 
 typedef struct
 {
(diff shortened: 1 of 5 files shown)


prboom2/src/r_defs.h
```diff
@@ -283,13 +289,20 @@ typedef enum
...
 } slopetype_t;
 
-typedef byte r_flags_t;
-#define RF_TOP_TILE 0x01 // Upper texture needs tiling
-#define RF_MID_TILE 0x02 // Mid texture needs tiling
-#define RF_BOT_TILE 0x04 // Lower texture needs tiling
-#define RF_IGNORE   0x08 // Renderer can skip this line
-#define RF_CLOSED   0x10 // Line blocks view
-#define RF_ISOLATED 0x20 // Isolated line
+typedef uint16_t r_flags_t;
```
kraflab: Please use a regular unsigned short here. Specifying 16 bits prevents the compiler from choosing on its own, which can only reduce performance. Of course, that's unlikely to make any difference here, but there's no need to add this restriction. Please also remove the use of the colon operator elsewhere for the same reason - unless you can demonstrate that choosing for the compiler creates a real impact, I don't think it's worth adding this pattern (which nothing else in the code base uses as far as I know). There _are_ places that do specify exact sizes, but usually this is a safety feature to deal with the binary formats (though there may be other uses to clean up at some point).

prboom2/src/r_defs.h
```diff
@@ -283,13 +289,20 @@ typedef enum
...
+#define RF_TOP_TILE 0x001 // Upper texture needs tiling
```
kraflab: Please use 4 characters for the hex to imply the full size for consistency (i.e., 0x0000)

kraflab:
I'm going to need some comprehensive performance comparisons. If it's in danger of creating a noticeable impact on load times then it won't be worth it to me. We can potentially just hard code exceptions like av01, since there aren't any insurmountable number of cases. But of course let's try this general solution.

Have you looked into how gzdoom handles this stuff?

prboom2/src/r_defs.h
bkoropoff: > Specifying 16 bits prevents the compiler from choosing on its own, which can only reduce performance

There are tens or hundreds of thousands of these structures in memory, so choosing the smallest type that fits the data is important, both for memory usage and cache efficiency.

> Please also remove the use of the colon operator elsewhere for the same reason

I also used bitfields to keep the size of `vertex_t` as low as possible.  Again, there can be hundreds of thousands of these in memory at once, and cache performance is going to be important.

kraflab: Supported maps are unimaginably small in the scope of modern gaming and those kinds of optimizations are completely irrelevant. If anything I've seen these attempts get in the way of optimization by blocking the compiler. Someone tried to optimize zone memory for the cache and low and behold gutting all those attempts actually made the zone memory faster. If you have definitive evidence let me know. Udmf alone added tons of integers to these data structures.

bkoropoff:
> We can potentially just hard code exceptions like av01

Do you have any preference for how to do this?  Store lumps in dsdadoom.wad indexed by map CRCs?

> Have you looked into how gzdoom handles this stuff?

GZDoom has some zscript that applies compatibility fixes to a bunch of maps.  It's solving the AV MAP01 some other way that I'll have to dig into.

kraflab:
We have a hash based compatibility autocorrect just like in gzdoom. It could be leveraged to apply autocorrects for this as well, if need be. But I would be happy to avoid that.

prboom2/src/r_defs.h
bkoropoff: I can remove bitfields.

bkoropoff:
Yeah, I'd like to find an analytic solution if possible.  I'll dig into GZDoom some more and see if there's something simpler to crib.

bkoropoff:
I take it back, something in the GZDoom compat script *is* modifying the level, which also forces it to rebuild GL nodes.  I found subsequent logic which uses GL node specific data to determine which sector should be used for rendering for each subsector.  One of the two things (the modification or the subsequent rendering sector pass) is fixing it.

bkoropoff:
Here: https://github.com/ZDoom/gzdoom/blob/master/wadsrc/static/zscript/level_compatibility.zs#L1232

bkoropoff:
Sunder MAP31 (extremely large map, takes over a second to load):

```
Traversal time: 77950 usec
```

HR2 MAP12 (has extreme cyclic complexity):

```
Traversal time: 174044 usec
```

The latter time could be reduced by tightening `UNDIR_DEPTH_LIMIT` to give up faster.  The relationship is exponential, e.g. reducing it to 7 halves the time:

```
Traversal time: 95415 usec
```

andrey-budko:
"traverse: Not enough memory for block map" message on hr2.wad map06 (state->bmapsize is 0)
https://prboom-plus.sourceforge.io/HR2.WAD.zip

hr2final.wad map19 takes more than 30 seconds to load

andrey-budko:
> The latter time could be reduced by tightening `UNDIR_DEPTH_LIMIT` to give up faster

Can't load hg.wad map07 even with UNDIR_DEPTH_LIMIT 3
https://www.doomworld.com/idgames/levels/doom2/Ports/g-i/hg

bkoropoff:
I pushed fixes for those problems, but I'm also exploring an alternate geometric approach.

andrey-budko:
my old test wad:
https://prboom-plus.sourceforge.io/stencil.wad.zip
[stencil.wad.zip](https://github.com/kraflab/dsda-doom/files/10906528/stencil.wad.zip)

<details>
<summary>master</summary>

image
</details>

<details>
<summary>pull 248</summary>

image
</details>

PR #255, opened 2023-03-11
BSP Analysis Infrastructure (and yet another AV MAP01 fix)

bkoropoff:
This introduces a set of analysis passes on the BSP tree:

- Find implicit segs within subsectors created by BSP partitions
- Figure out which subsectors are adjacent
- Detect or fix various anomalies (degenerate subsectors, fake segs and subsectors, overly narrow bounding boxes)

These are all things GL nodes give you, and that the rest of the GL renderer sort of works around here and there when it doesn't have them:

- The subsector triangulator computes the implicit outline of subsectors, although it doesn't actually record the segs in any useful way
- GLU tessellation of entire sectors, when it works, avoids the subsector triangulator choking on subsectors that are broken or not actually convex.  AV MAP01 has one of these that GLU hides.  Requiem MAP13 has a bunch that show up as HOM, because the sector in question has a single mistagged sidedef that prevents it from being GLU tessellated.
- Rendering all flats for a sector when `R_Subsector` touches one of its subsectors masks the fact that bounding boxes in the BSP tree are too small and actually cause some subsectors to be incorrectly culled, which I found out when trying to make the GL render path closer to software.  The software renderer skates by on the flood-fill like nature of its flat drawing, as usual.

BSP stands for "barely sufficient polygons".

This does fix AV MAP01, but it's intended to be a general starting point for improving the GL renderer.

It's basically impossible to make sense of what's going on by just poking around in a debugger, so I added a console command `dumpbsp <file.json>`, to dump the loaded/analyzed/fixed BSP data to a file.  The script `bsdplot.py` uses matplotlib to create navigable visualizations from dumps.  For example, here's E1M1:

```sh
# E1M1, BSP partition lines in magenta (clipped to local bounding box),
# implicit segs in blue, normal segs in black
python3 ./tools/bspplot.py e1m1.json \
  partitions --color 1,0,1 --width 0.7 isegs --color 0,0,1 --width 0.8 \
  segs
```
e1m1

Here's the problem child of AV MAP01:

```sh
# AV MAP01, restricted to sectors 34, 37, 38 and adjacent subsectors,
# partition lines in magenta, implicit segs in blue, normal segs in black,
# detected fake segs in red and annotated with text
python3 ./tools/bspplot.py av1.json --filter 'sector = 34,36,37 or adjacent.sector = 34,36,37' \
  partitions --color 1,0,1 --width 0.25 \
  isegs --color 0,0,1 --width 0.5 \
  segs \
  segs --color 1,0,0 --select 'flags.fake' --annotate
```

av1

Yeah, the annotation placement could be better.

I've tested this on the usual suspects that have come up in previous PRs.  It makes no change in how stencil.wad renders.  It doesn't seem to add to load times.


bkoropoff:
MSVC doesn't know the square root of 2 I guess.  Fixing.

bkoropoff:
> D:\a\dsda-doom\dsda-doom\prboom2\src\dsda\bsp.c(68,65): error C2099: initializer is not a constant [D:\a\dsda-doom\dsda-doom\build\src\dsda-doom.vcxproj]

> static const int epsilon_factor = (1 << 14);

> **const**

Love MSVC.

kraflab:
I'd suggest creating a separate repo for your python script and any other tools you end up creating, so that you're in charge of the source there.

bkoropoff:
Yeah, that would make it faster to improve the script alone, but it's also tightly tied to the output format of `dumpbsp`, so updates to the engine might require changing the script, and they can get out of sync.  Not sure which situation is better.

prboom2/src/dsda/bsp.h
```diff
@@ -0,0 +1,32 @@
...
+#define DSDA_BSP_H_
+
+#include "r_defs.h"
+
+extern iseg_t* isegs;
+extern unsigned int numisegs;
+extern int* adjacency;
+extern unsigned int numadjacency;
+
+void dsda_AnnotateBSP(void);
```
kraflab, line 34: If it's not too much agony, I'd prefer that you split out `bsp/dump` so that the annotation business logic is isolated and succinct.

bkoropoff: I broke geometry/bsp/bspdump stuff into separate files, should flow better now

prboom2/src/dsda/dgeom.h
```diff
@@ -0,0 +1,212 @@
...
+//
+//      Not intended for use outside of DSDA subdir, so these aren't
+//      prefixed with dsda_
+//
+
+#ifndef DSDA_INTERNAL_
+#error Internal header, not for use outside dsda subdir
+#endif
+
+#ifndef DSDA_DGEOM_H_
```
kraflab: Please use the existing header pattern, in this case `__DSDA_DGEOM__`

prboom2/src/dsda/dgeom.h
```diff
@@ -0,0 +1,212 @@
...
+// Uncertainty in a map x,y distance
+static const double epsilonmapr2 = (double) FRACUNIT * M_SQRT2 - 1;
+
+//
+// Inline functions
+// 
+
+// Conversion functions
+
+static inline fixed_t DoubleToFixed(double v)
```
kraflab: What does this function do exactly? If it's to fixed, it should multiply by FRACUNIT. If the double is already in fixed units, another name would be preferable.

prboom2/src/dsda/dgeom.h
```diff
@@ -0,0 +1,212 @@
...
+#ifndef DSDA_DGEOM_H_
+#define DSDA_DGEOM_H_
+
+#include <math.h>
+
+#include "m_fixed.h"
+#include "r_defs.h"
+#include "r_state.h"
+
+struct dpoint
```
kraflab: Please typedef all structs. I.e., set this to a `dpoint_t` and reference it as such when possible.

prboom2/src/dsda/dgeom.h
```diff
@@ -0,0 +1,212 @@
...
+//
+// DESCRIPTION:
+//	Internal double-precision geometry functions
+//
+//      Not intended for use outside of DSDA subdir, so these aren't
+//      prefixed with dsda_
+//
+
+#ifndef DSDA_INTERNAL_
+#error Internal header, not for use outside dsda subdir
```
kraflab: This protection doesn't look particularly necessary to me. I do think we need a prefix though, `dgeom_` maybe. Unprefixed simple names like `Len` don't capture their specificity otherwise.

prboom2/src/dsda/dgeom.h
bkoropoff: FYI, all identifiers starting with `__` are reserved by the C standard for system headers and shouldn't be used in user programs.  C++ is even more strict, see https://en.cppreference.com/w/c/language/identifier

prboom2/src/dsda/dgeom.h
bkoropoff: Yeah, all the dgeom functions use fixed units to avoid dividing/multiplying on conversion

prboom2/src/dsda/dgeom.h
bkoropoff: I was trying to avoid renaming everything, but I can add a prefix.

prboom2/src/dsda/dgeom.h
bkoropoff: I added a `dfixed_t` alias for double to make it clear that these functions are intended to work on fixed units.

prboom2/src/dsda/dgeom.h
kraflab: I appreciate that but every header uses the underscores like this, and I have no interest in changing that right now 😄

prboom2/src/dsda/dgeom.h
kraflab: 👍

prboom2/src/dsda/dgeom.h
kraflab: But, these aren't identifiers, so I'm not sure the doc you linked applies here. The defined strings are only seen by the compiler and don't have "meaning".

bkoropoff: Macro names are considered identifiers by the spec.  I changed this to follow the existing convention anyway.

kraflab: Oh I see. But why doesn't the compiler complain then?

bkoropoff: Clang will complain in some cases with `-Wreserved-identifier`, but compilers have generally not caught it because:

- System headers are allowed to use reserved identifiers, and the compiler can't easily tell what's a system header and what isn't.
- It's undefined behavior according to the spec, which means it's perfectly fine for your program to subtly stop working on the next version of Windows or MSVC or whatever because you used a reserved identifier.

kraflab: I see, thanks for the info 🙌

bkoropoff:
I have a more promising experimental branch based on this, so closing this

PR #260, opened 2023-03-16
Disable bleed-over logic for now

### LINKED ISSUE #256: gl rendering glitch on eternal.wad map02
https://www.doomworld.com/idgames/themes/TeamTNT/eternal/eternal

broken after https://github.com/kraflab/dsda-doom/commit/35b26aad4c14f45f5eeff6f1e9d66c4c764e81b3 (r_bsp: bleed through very low floors as well)

image

bkoropoff:
Closes #256

### DIFF
--- a/prboom2/src/r_bsp.c
+++ b/prboom2/src/r_bsp.c
@@ -42,6 +42,9 @@
 #include "v_video.h"
 #include "lprintf.h"
 
+// Turned off because it causes regressions on some maps (issue #256).  Fixing
+// this requires doing bleed with subsector granularity.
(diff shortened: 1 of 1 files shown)


kraflab:
Could you put something like `#define EXPERIMENTAL_BLEED_LOGIC 0` and use that for the ifdefs, and maybe add a comment for context on why it's off by default? I'm thinking about setting up a side branch for opengl stuff and planning for the long term with that coming in 0.27, so it will be good to leave a note in master.

bkoropoff:
Yeah, I can do that

kraflab:
Thanks 👍

PR #262, opened 2023-03-20
CMake: Improve dependency searching

FtZPetruska:
This PR brings a major rework of the dependency searching that aims to provide:
- better support for static linking
- interoperability with upstream CMake configs shipped through package managers or custom builds
- ~~a much smoother experience on Windows~~

---

### Static Linking Improvements

Find modules commonly do not capture transitive usage requirements of libraries, especially when they are not trivial, `Vorbis` always depends on `Ogg`, but `SDL2_mixer` may depend on up to nine libraries, each potentially have more transitive dependencies. This works fine for shared libraries, since they do not require the final executable to link to all their dependencies. But for static libraries, this is breaks in pieces.

To fix this issue, I use `PkgConfig` in the Find modules to the list of dependencies and flags that should be used with the package, as well as hints for where to find headers and libraries. It is commonly installed on Linux and macOS as a dependency for `-dev` packages.

### Interoperability with Upstream Configs

Since CMake 3.15, setting `CMAKE_FIND_PACKAGE_PREFER_CONFIG` to `ON` (through the CLI/cmake-gui/ccmake) makes CMake try searching with the `CONFIG` mode first, but if the upstream config does not provide the same target we expect (e.g. SDL2 provides `SDL2::SDL2` but we expect `SDL2::Core`), then the user gets an error. 
Similarly, if a user has SDL2 installed in a non-standard location and sets `SDL2_DIR` to the location of `SDL2Config.cmake`, the Find module may fail, but the config while it works could provide unexpected targets.

To address this, I checked upstream to see what nomenclature every dependency was using upstream (if any), and applied them to the Find modules, here's a breakdown of the findings:

<details>

<summary>Target names</summary>

SDL2 (has an upstream config):
  - SDL2::SDL2, core library
  - SDL2::SDL2main, main library, may not be available

SDL2_mixer (has an upstream config):
  - SDL2_mixer::SDL2_mixer, shared builds only
  - SDL2_mixer::SDL2_mixer-static, static builds only

SDL2_image (has an upstream config):
  - SDL2_image::SDL2_image, shared builds only
  - SDL2_image::SDL2_image-static, static builds only

FluidSynth (has an upstream config):
  - FluidSynth::libfluidsynth

Vorbis (has an upstream config):
  - Vorbis::vorbis, component "Vorbis", the base encoder/decoder library
  - Vorbis::vorbisenc, component "Enc", the high-level API encoder library
  - Vorbis::vorbisfile, component "File", the high-level API decoder library

Ogg (has an upstream config):
  - Ogg::ogg

PortMidi (has an upstream config):
  - PortMidi::portmidi

LibMad (no upstream config):
  - LibMad::libmad

DUMB (no upstream config):
  - DUMB::DUMB

</details>

<details>

<summary>Removed from this PR</summary>

### vcpkg integration

vcpkg is probably the easiest way to get C/C++ dependencies on Windows (at least in my experience), and it integrates nicely with CMake since it is designed around it.

The changes I have done are:
- Add a "manifest" (vcpkg.json) next to the top-level CMakeLists.txt, this lists the dependencies and the "features" (optional dependencies). I made it so all the `WITH_*` options in CMake have a matching feature in the manifest.
- In the CMakeLists, determine the features used based on the options. This has to be done **before** the call to `project` hence why the option declarations were moved up.
- If CMake is detected, set `CMAKE_FIND_PACKAGE_PREFER_CONFIG` to `ON`, since vcpkg always ships the upstream CMake config when available, this helps most notably static linking on Windows where pkg-config is not often available.

For packages that are not currently available on vcpkg (i.e. DUMB), or that have notable issues with a patch waiting to be released, I have included an `overlay port` in the `vcpkg_overlay/ports` directory:
- the `dumb` port is an updated version of what was in `.github/workflows`, the commands are updated replace the deprecated ones, and the `CONTROL` files is replaced with a manifest.
- the `sdl2-mixer` port contains a set patches that fixes features and static linking
- the `fluidsynth` port is an updated version based on the [2.3.2 commit](https://github.com/FluidSynth/fluidsynth/commit/0589db1a286a5fd70711a020bf5a3fd527e3fdb1), most notably, this version drops the hard dependency on pkg-config when consuming the library through CMake

With these changes, making a static build under Windows is as simple as running: 
```
cmake -Sprboom2 -Bbuild -DCMAKE_TOOLCHAIN_FILE=/path/to/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows-static-release
cmake --build build --config Release
```

Also, Woof has a similar vcpkg integration, and it seems to work well for them.

</details>

---

Additionally, there is a few miscellaneous changes:
- ~~Making a feature summary, at the end of the configure step, it prints a summary of all the optional/recommended/required packages that were found or not, as well as their purpose. I'm open to feedback if you think it's too noisy, if I misunderstood the purpose of a lib, or if something should be made recommended instead of optional and vice-versa.~~
- Setting the policy CMP0099 to NEW, this covers a *very* specific edge-case, where when you have libFoo installed in `/prefix1`, and it depends on libBar installed in `/prefix2`:
  - if `/prefix2` is not in your compiler default search path
  - if the dependencies of libFoo are captured with pkg-config
The `INTERFACE_LINK_DIRECTORIES` property is no transitively applied to the consumer of libFoo, which results in link failures. CMP0099 makes that property transitive, as it should, but it defaults to `OLD`.

Feedback is greatly appreciated!


rfomin:
Thanks for your effort, I can confirm it works for my MSVC + vcpkg build.

How about adding CPack support? With `X_VCPKG_APPLOCAL_DEPS_INSTALL` it automatically collects all dependencies:

<details>
  <summary>patch</summary>

```diff
diff --git a/prboom2/CMakeLists.txt b/prboom2/CMakeLists.txt
index 75b8817d7..541830383 100644
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -30,6 +30,10 @@ if(WITH_VORBISFILE)
     list(APPEND VCPKG_MANIFEST_FEATURES "libvorbis")
 endif()
 
+# X_VCPKG_APPLOCAL_DEPS_INSTALL automatically installs dependencies.
+set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+set(X_VCPKG_APPLOCAL_DEPS_INSTALL ON)
+
 project("dsda-doom" VERSION 0.25.6)
 
 # Set a default build type if none was specified
@@ -280,6 +284,12 @@ set(PRBOOM_OUTPUT_PATH ${CMAKE_BINARY_DIR})
 
 set(WAD_DATA_PATH "${PRBOOM_OUTPUT_PATH}/${WAD_DATA}")
 
+# Generate distribution packages with CPack.
+if(WIN32)
+    set(CPACK_GENERATOR ZIP)
+    include(CPack)
+endif()
+
 add_subdirectory(data)
 add_subdirectory(src)
 
diff --git a/prboom2/data/CMakeLists.txt b/prboom2/data/CMakeLists.txt
index cb6afabe1..282df4f93 100644
--- a/prboom2/data/CMakeLists.txt
+++ b/prboom2/data/CMakeLists.txt
@@ -411,4 +411,8 @@ add_custom_command(
     DEPENDS rdatawad ${WAD_SRC}
 )
 add_custom_target(dsda-doom-wad DEPENDS ${WAD_DATA_PATH})
-install(FILES ${WAD_DATA_PATH} DESTINATION ${DSDAPWADDIR} COMPONENT "DSDA-Doom internal WAD")
+if(WIN32)
+    install(FILES ${WAD_DATA_PATH} DESTINATION . COMPONENT "DSDA-Doom internal WAD")
+else()
+    install(FILES ${WAD_DATA_PATH} DESTINATION ${DSDAPWADDIR} COMPONENT "DSDA-Doom internal WAD")
+endif()
\ No newline at end of file
diff --git a/prboom2/src/CMakeLists.txt b/prboom2/src/CMakeLists.txt
index cebedb661..a2cb3d169 100644
--- a/prboom2/src/CMakeLists.txt
+++ b/prboom2/src/CMakeLists.txt
@@ -574,7 +574,11 @@ function(AddGameExecutable TARGET SOURCES)
         )
     endif()
 
-    install(TARGETS ${TARGET} COMPONENT "Game executable" RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}")
+    if(WIN32)
+        install(TARGETS ${TARGET} COMPONENT "Game executable" RUNTIME DESTINATION .)
+    else()
+        install(TARGETS ${TARGET} COMPONENT "Game executable" RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}")
+    endif()
 
     target_link_libraries(${TARGET} PRIVATE
         ${OPENGL_gl_LIBRARY}
```
</details>

Release MSVC build zip:
image

Static release MSVC build:
image

prboom2/CMakeLists.txt
```diff
@@ -22,6 +52,19 @@ TEST_BIG_ENDIAN(WORDS_BIGENDIAN)
 
 include(GNUInstallDirs)
 
+include(FeatureSummary)
+
+if(VCPKG_TOOLCHAIN)
+    set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
```
rfomin: What if we enable this option for all build types? It seems MSYS2 also provides configs.

kraflab:
Can any of these things be split apart? I would prefer to have separate PRs for separate improvements / features, as it will make it much easier for me to understand the changes.

FtZPetruska:
> Can any of these things be split apart? I would prefer to have separate PRs for separate improvements / features, as it will make it much easier for me to understand the changes.

Absolutely! I'll take away the vcpkg and feature summary parts out of this PR and just focus on the Find modules.

> Thanks for your effort, I can confirm it works for my MSVC + vcpkg build.
> 
> How about adding CPack support? With `X_VCPKG_APPLOCAL_DEPS_INSTALL` it automatically collects all dependencies:

Thank you for your testing! CPack support sounds like a great idea, since the plan is to split this PR into smaller one, I'll apply your patch along the one that adds vcpkg support.

FtZPetruska:
Okay, I've dropped the other commits and only kept the Find modules changes (and rebased on master while I was at it).

prboom2/src/CMakeLists.txt
```diff
@@ -517,50 +517,48 @@ function(AddGameExecutable TARGET SOURCES)
...
             winmm
             comctl32
         )
     endif()
+
     set_target_properties(${TARGET} PROPERTIES
         RUNTIME_OUTPUT_DIRECTORY ${PRBOOM_OUTPUT_PATH}
     )
-    if(SDL2_IMAGE_FOUND)
-        target_include_directories(${TARGET} PRIVATE ${SDL2_IMAGE_INCLUDE_DIRS})
```
kraflab, line 533: Could you tell me what the include_directories lines did and why they aren't needed anymore?

prboom2/src/CMakeLists.txt
```diff
@@ -517,50 +517,48 @@ function(AddGameExecutable TARGET SOURCES)
         ${CMAKE_BINARY_DIR}
         ${CMAKE_CURRENT_SOURCE_DIR}
     )
+
     target_link_libraries(${TARGET} PRIVATE
         ${SDL2_LIBRARIES}
+        SDL2_mixer::SDL2_mixer
+        ZLIB::ZLIB
```
kraflab, line 524: Is the change here because the old syntax indicated it was optional, but it should be here since it's required?

prboom2/CMakeLists.txt
```diff
@@ -106,8 +118,8 @@ endif()
 
 option(WITH_VORBISFILE "Use vorbisfile if available" ON)
 if(WITH_VORBISFILE)
-    find_package(Vorbis)
-    if(VORBIS_FOUND)
+    find_package(Vorbis COMPONENTS File)
```
kraflab, line 121: What does the difference here do?

kraflab (commented):
Just a few questions as I don't know much about cmake. I tested it on my end as well (msys2) and it all works.

prboom2/src/CMakeLists.txt
FtZPetruska: Previously, Find modules would declare variables `*_INCLUDE_DIRS` and `*_LIBRARIES` that contained literal paths to, respectively, the include directories and libraries.

Now, Find modules create imported CMake targets (`add_library(Foo::Foo UNKNOWN IMPORTED)`) which contain both the library itself, its dependencies and include directories:
```cmake
set_target_properties(Foo::Foo PROPERTIES
  IMPORTED_LOCATION "/path/to/the/library"
  INTERFACE_INCLUDE_DIRECTORIES "/path/to/include"
  INTERFACE_LINK_LIBRARIES "list;of;other;dependencies")
```

Linking to this target through `target_link_libraries` applies all these properties to the target, making `target_include_directories` redundant.

The only exception to that is with SDL2. On 2.0.10 and below, when SDL is built with autotools, the CMake config is really bare, only providing `SDL2_LIBRARIES` and `SDL2_INCLUDE_DIRS` and no target. So they are used to keep compatibility with 2.0.7.

prboom2/src/CMakeLists.txt
FtZPetruska: The old syntax guarded zlib behind `if(ZLIB_FOUND)`, which is no longer necessary since zlib is declared as required (`find_package(ZLIB REQUIRED)`).

Since I had to edit the calls to `target_link_libraries` anyway to change `*_LIBRARIES` into the target, I just moved all the required targets together.

prboom2/src/CMakeLists.txt
kraflab: Thanks, makes sense 👍

prboom2/CMakeLists.txt
FtZPetruska: The upstream config for Vorbis declares its 3 libraries as components: `Vorbis`, `Enc`, and `File`, for respectively vorbis, vorbisenc, and vorbisfile.

Currently, it seems like all components are included by default, but there are no guarantees that it will stay the same. CMake does not provide any guarantee for the default with components, leaving it to the package to decide. Since we specifically want vorbisfile, we ask for it explicitly.

In the Find module, I reflect that requirement with:
```cmake
find_package_handle_standard_args(
  Vorbis
  REQUIRED_VARS "Vorbis_File_LIBRARY" "Vorbis_Vorbis_LIBRARY"
                "Vorbis_INCLUDE_DIR"
  HANDLE_COMPONENTS)
```
Where I declare that only `File` and `Vorbis` are required, and that it should handle components.

prboom2/src/CMakeLists.txt
kraflab: 👍

prboom2/CMakeLists.txt
kraflab: Awesome, thanks 👍

kraflab:
Thanks a lot 👍

PR #263, opened 2023-03-20
Fix Warnings when building for the Win32 target

RamonUnch:
We cannot assume that a size_t == int64 nor can we assume that intptr_t == int64

So I just use some casts to avoid inconsistent format in the case of 32bit builds

### DIFF
--- a/prboom2/src/doomtype.h
+++ b/prboom2/src/doomtype.h
@@ -87,6 +87,12 @@ typedef unsigned char byte;
 #define NORETURN
 #endif
 
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
+  #define NORETURNC11 _Noreturn
+#else
+  #define NORETURNC11
+#endif
+
 // Definition of PACKEDATTR from Chocolate Doom
 #ifdef __GNUC__
   #if defined(_WIN32) && !defined(__clang__)
--- a/prboom2/src/doomtype.h
+++ b/prboom2/src/doomtype.h
@@ -87,12 +87,6 @@ typedef unsigned char byte;
 #define NORETURN
 #endif
 
-#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
-  #define NORETURNC11 _Noreturn
-#else
-  #define NORETURNC11
-#endif
-
 // Definition of PACKEDATTR from Chocolate Doom
 #ifdef __GNUC__
   #if defined(_WIN32) && !defined(__clang__)
--- a/prboom2/src/hexen/sv_save.c
+++ b/prboom2/src/hexen/sv_save.c
@@ -217,7 +217,7 @@ static uint64_t SV_ReadFlags(void)
 
 static void *SV_ReadPtr(void)
 {
-    return (void *) (intptr_t) SV_ReadLong();
+    return (void *) (uintptr_t) SV_ReadLong();
 }
 
 static void SV_Write(const void *buffer, size_t size)
--- a/prboom2/src/hexen/sv_save.c
+++ b/prboom2/src/hexen/sv_save.c
@@ -217,7 +217,7 @@ static uint64_t SV_ReadFlags(void)
 
 static void *SV_ReadPtr(void)
 {
-    return (void *) (uintptr_t) SV_ReadLong();
+    return (void *) (intptr_t) SV_ReadLong();
 }
 
 static void SV_Write(const void *buffer, size_t size)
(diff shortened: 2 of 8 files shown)


fabiangreffrath:
> The way I choosed to handle is to define format strings for each types:

An alternative approach would be to use the regular `%ld` format string and cast the `size_t` type variables to `(long)` in the `fprintf()` call.

> I am not sure on how to detect 32bit Linux so for now it is assumed Linux is 64bit only.

`ifdef __i386__`

https://wiki.debian.org/ArchitectureSpecificsMemo

RamonUnch:
> > The way I choosed to handle is to define format strings for each types:
> 
> An alternative approach would be to use the regular `%ld` format string and cast the `size_t` type variables to `(long)` in the `fprintf()` call.


I thought about that but I do not like it much, because you have to either extend to long long on 32bit platform or to get a truncated result on x64, this would have been slightly simpler but I thought it would be nicer to implement cistom `PRI*` macro to extend those that are defined in `inttypes.h`

> > I am not sure on how to detect 32bit Linux so for now it is assumed Linux is 64bit only.
> 
> `ifdef __i386__`
> 
> https://wiki.debian.org/ArchitectureSpecificsMemo

Thanks! How could I forget about that.

prboom2/src/hexen/sv_save.c
```diff
@@ -217,7 +217,7 @@ static uint64_t SV_ReadFlags(void)
 
 static void *SV_ReadPtr(void)
 {
-    return (void *) (intptr_t) SV_ReadLong();
```
kraflab: Why this?

fabiangreffrath:
> I thought about that but I do not like it much, because you have to either extend to long long on 32bit platform or to get a truncated result on x64, 

My take is that if a value has to get truncated in order to fit into a variable of type `long` on any platform, it is most probably not human readable anyway.

> this would have been slightly simpler but I thought it would be nicer to implement cistom `PRI*` macro to extend those that are defined in `inttypes.h`

Yes, but `inttypes.h` is maintained by your compiler chain, whereas you will have to maintain your own macros yourself across all platforms .

prboom2/src/doomtype.h
```diff
@@ -87,6 +87,12 @@ typedef unsigned char byte;
 #define NORETURN
 #endif
 
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
```
kraflab: Please split off the noreturn stuff to a separate pr 🙏

kraflab:
Whenever I added these format changes to fix warnings I looked into the PRI macros and decided not to use them. They ruin the readability of the format string and are impossible for me to maintain (because I'm neither going to compile a 32 bit binary nor am I going to verify potential warnings for 32 bit builds).

prboom2/src/hexen/sv_save.c
RamonUnch: this triggers the `-Wbad-function-cast` warning, because we are casting rom a signed value to a void*.
But now that you say it maybe the value needs to be sign-extended when going from a 32bit int to a 64bit void* ?

I should investigate first why there is this crazy int->void* cast in the first place.

kraflab: This code is actually "dead" and has no practical impact on execution I think. If you check the associated values they are read and then set to NULL anyway. I think all this code was autogenerated for chocolate hexen and then specific fields were patched to NULL. The read / write of pointers in sv_save can just be purged from the code. I guess chocolate hexen left it all in for save file compatibility.

RamonUnch:
Ok, I will do this:
1. Separated PR for noreturn stuff.
2. Revert the uintprt_t cast because this will be cleand up later anyway.
3. Use simple casts to fix printf for 32bit platforms.

kraflab:
Sounds great, thanks!

prboom2/src/hexen/sv_save.c
fabiangreffrath: > I should investigate first why there is this crazy int->void* cast in the first place.

That has some historic background. Back in the days, HEXEN.EXE used to save entire structs in its savegames. These structs contained pointers, which were of course invalid once the savegame was restored, thus they were read back into the struct but then invalidated. To improve portability to systems where `sizeof void* != sizeof int`, the writing and reading of the struct content was split into several functions, one for each struct member. So, in order to restore a pointer element of a struct from a savegame, you'd actually have to read an `int` and convert this into a pointer. However, some compilers complained if you cast an `int` into a pointer of different size and thus the intermediate cast had to be added. The return value is still invalidated, after all.

RamonUnch:
It is done,
I still have left the #include <math.h> that is needed for the pow() function.

prboom2/src/hexen/sv_save.c
RamonUnch: Thanks for this info, I love this kind of historic context.

kraflab:
Thanks 👌

PR #264, opened 2023-03-20
Make use of the C11 noreturn specifier or gcc's attribute

RamonUnch:
And the noreturn gcc specific attribute
for I_Error()
and I_SafeExit()


kraflab:
Thanks 👌

PR #265, opened 2023-03-21
CMake: Fix FindSDL2

FtZPetruska:
I noticed a few issues with the FindSDL2 modules from #262, namely:
- `manual-link` should be a `PATH_SUFFIXES` (vcpkg places SDL2main in `lib/manual-link`).
- SDL2main would set `SDL2_SDL2_FOUND` instead of `SDL2_SDL2main_FOUND`, that would lead to `SDL2::SDL2main` never being created, and `SDL2::SDL2` being created depending on SDL2main availability.


kraflab:
👍

PR #266, opened 2023-03-22
CMake: Add vcpkg integration

FtZPetruska:
This PR bring vcpkg integration to the project.

This is done by adding a `vcpkg.json` manifest file that lists out the dependencies of the project, optional dependencies are listed as "features".

When the vcpkg toolchain is provided to CMake (done automatically in Visual Studio), dependencies will be automatically built and made available.

For dependencies that are not available in vcpkg, custom ports are added in `vcpkg_overlay/ports`, namely:
- `dumb` is an updated version fo the port that was located in `.github/workflows` that uses the newer functions.
- `fluidsynth` is an updated version of the port to the [2.3.2](https://github.com/FluidSynth/fluidsynth/commit/0589db1a286a5fd70711a020bf5a3fd527e3fdb1) commit, this version no longer has a hard dependency on pkg-config when searching for its dependencies. Additionally, there is a minor patch to not install the `fluidsynth` executable.

Turning on/off the optional dependencies (SDL_image, DUMB, libmad, etc) in CMake automatically selects the features for vcpkg. For this to work, the options must be declared **before** the call to `project()`.

---

<details>

<summary> Removed from this PR </summary>

For completeness, the `windows_dependencies` workflow is updated since it relies on vcpkg:
- Build both x64 and x86 to generate an archive with the same layout as what's [provided upstream](https://github.com/coelckers/prboom-plus/releases/tag/windows_dependencies).
- Use the `-release` triplets to skip building debug, this drastically reduce runtime for CI (from 37 minutes down to  minutes).
- Make the workflow being able to be triggered manually (`workflow_dispatch`) instead of pushing a commit to a dummy branch.

You can see the result of a run [here](https://github.com/FtZPetruska/dsda-doom/actions/runs/4484625066), do note that the archive size displayed here is way off the actual size (shown 55MB, actually 27MB).

</details>

---

Lastly, as suggested by @rfomin, I made `CMAKE_FIND_PACKAGE_PREFER_CONFIG` an option defaulting to `ON`.
Setting this makes static linking much more consistent when the CMake config of dependencies is available.


rfomin:
Thank you for `windows_dependencies` workflow updates!

Previously, I didn't want to bother with vcpkg and just copied the `dependencies_x64` folder from `windows_dependencies.zip` to the root of the project. Perhaps we should document this and make a new dependencies "release", like here: https://github.com/coelckers/prboom-plus/releases/tag/windows_dependencies

FtZPetruska:
> Previously, I didn't want to bother with vcpkg and just copied the `dependencies_x64` folder from `windows_dependencies.zip` to the root of the project. Perhaps we should document this and make a new dependencies "release", like here: https://github.com/coelckers/prboom-plus/releases/tag/windows_dependencies

Instead of manually downloading it, there's also the possibility to use `FetchContent` (available from 3.11) from CMake directly, so instead of having the users manually download the archive and update it from time to time, we could put in the CMakeLists something like:
```cmake
include(CMakeDependentOption)
cmake_dependent_option(USE_PREBUILT_WINDOWS_DEPENDENCIES
  "Download prebuilt windows dependencies using FetchContent" 
  ON "MSVC;NOT VCPKG_TOOLCHAIN" OFF)

if(USE_PREBUILT_WINDOWS_DEPENDENCIES AND MSVC AND NOT VCPKG_TOOLCHAIN)
  include(FetchContent)
  FetchContent_Declare(
    WindowsDependencies
    URL https://github.com/..../windows_dependencies.zip
  )
  FetchContent_MakeAvailable(WindowsDependencies)
  list(APPEND CMAKE_PREFIX_PATH "${windowsdependencies_SOURCE_DIR}/dependencies_${DEPENDENCY_SUFFIX}")
endif()
```

This would download the prebuilt binaries and add them to the search path if vcpkg is not found.
`cmake_dependent_option` is used so:
- the user can simply turn it off
- it's automatically hidden if the compiler is not MSVC or if the vcpkg toolchain is provided.

I do think a new release of prebuilt Windows dependencies could be nice considering the latest archive from prboom+ is already three years old.

But all of this should probably be follow-up PRs, similar to how I held off the CPack changes you suggested yesterday to avoid making this PR bigger than it already is.

rfomin:
> Instead of manually downloading it, there's also the possibility to use `FetchContent` ...

Sounds good.

> Make the workflow being able to be triggered manually (`workflow_dispatch`) instead of pushing a commit to a dummy branch.

How about `schedule` event? https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule @kraflab doesn't use MSVC and will most likely never update `windows_dependencies` 😄

FtZPetruska:
> How about `schedule` event? ...

Making it a scheduled event is not so much of an issue, the issue remains to get the archive from the output to the "fake" release. There are some 3rd party actions to handle updating releases, but I am concerned about forks running it and each creating their own release, it does not seem like you can tell GitHub something like "do not include/run this workflow in forks".

Regardless, it would be preferable to have an existing release before thinking about setting up an automatic update mechanism.

kraflab:
Why exactly do we need this `windows_dependencies` thing in the first place?

FtZPetruska:
> Why exactly do we need this `windows_dependencies` thing in the first place?

It seems to be a leftover from coelckers/prboom-plus. It builds the "windows_dependencies" archive that were published [here](https://github.com/coelckers/prboom-plus/releases/tag/windows_dependencies) since for MSVC specifically, the only ways to get the dependencies is to acquire them manually or to install a package manager such as vcpkg. Having it a separate workflow to the main CI helps saving compilation time so we don't spend 20 minutes just building the dependencies every run and can just download it as such:

```bash
if [[ "${{ runner.os }}" == 'Windows' ]]; then
  declare -x DEPS_FILENAME=windows_dependencies_v4.zip
  curl -LO https://github.com/coelckers/prboom-plus/releases/download/windows_dependencies/${DEPS_FILENAME}
  7z x ${DEPS_FILENAME} -o..
```

rfomin:
We can get rid of `windows_dependencies` using vcpkg caching, we did it in Woof: https://github.com/fabiangreffrath/woof/blob/3bc2769f0acf3b2a07ef99736a47183a121a4a29/.github/workflows/win_msvc.yml#L26

But I think in some cases `windows_dependencies` is better.

kraflab:
I'd like to get rid of it if possible. It feels wrong for a build to rely on an artifact from a separate build. Ideally the checks should be independent and self-contained.

FtZPetruska:
> I'd like to get rid of it if possible. It feels wrong for a build to rely on an artifact from a separate build. Ideally the checks should be independent and self-contained.

Noted, I'll revert this commit for now, would you like the CI update (full removal of the windows_dependencies workflow, and making the continuous_integration one independent from upstream) to be added to this PR or to be a separate one ?

kraflab:
> Noted, I'll revert this commit for now, would you like the CI update (full removal of the windows_dependencies workflow, and making the continuous_integration one independent from upstream) to be added to this PR or to be a separate one ?

Let's do that one separately after this one is merged. That's assuming merging this on its own won't introduce any issue for it?

FtZPetruska:
> Let's do that one separately after this one is merged. That's assuming merging this on its own won't introduce any issue for it?

The current changes don't impact CI, since it does not use vcpkg. You can review it as-is 👍

kraflab:
One last thing to double check: for anyone not using vcpkg, this PR will not change anything for their build environment, correct?

FtZPetruska:
> One last thing to double check: for anyone not using vcpkg, this PR will not change anything for their build environment, correct?

The environment remains unchanged. Even if you have vcpkg installed, as long as you don't pass the toolchain file the behaviour will remain the same.

kraflab:
👍 thanks again

PR #273, opened 2023-03-23
CI: Remove unused workflow, modernise build workflow

FtZPetruska:
This PR drops the windows_dependencies workflow and all its associated files.

Moreover, it updates the continuous_integration workflow in the following ways:
- Use the latest version of actions (`@v3`)
- Run build on Msys2, so mingw issues can be caught early
- For MSVC builds, use vcpkg instead of depending on [prboom+](https://github.com/coelckers/prboom-plus/releases/download/windows_dependencies/). The binaries are cached, so while the first run takes about half an hour, subsequent runs take about four minutes ([see example run with cached binaries](https://github.com/FtZPetruska/dsda-doom/actions/runs/4501550615/jobs/7922139474)), thanks @rfomin for the idea!
- Use Ninja on all platforms (but MSVC), it's generally much faster than Makefiles.
- For apt and Homebrew, avoid partial upgrades by running `update` and `upgrade` before install.
- Use `apt-get` instead of `apt`, the latter warns about being unsuited for use in scripts.
- Switch to the `-latest` images, it is strongly recommended for Msys2, and avoids having to manually switch to newer images when the current ones get deprecated.


kraflab:
Nice and clean, thanks 💪

PR #274, opened 2023-03-24
Fix GitHub language detection 

FtZPetruska:
GitHub shows dsda-doom as a C++ project despite only using C. The reason for this is that linguist, the program used by GitHub to guess the language used in a source file, picks up the emacs style comments at the top of source files:

```c
/* Emacs style mode select -*- C++ -*-
```

Switching out `C++` for `C` properly updates the GitHub language section to "97.7% C and 2.3% Other", while still selecting the correct mode for people using Emacs.


kraflab:
Nice 😄

PR #275, opened 2023-03-25
CMake: Minor cleanups

FtZPetruska:
This PR brings a handful of minor cleanups in the CMake script that aim to modernise how flags and definitions are set, make the behaviour with vendored dependencies more explicit, and improve version compatibility.

### Flags and Defines

Generally, modern CMake guidelines strongly recommend against the use of `CMAKE_<LANG>_FLAGS` and the `add_definitions`/`add_compile_options`/`include_directories`/`link_directories`/`link_libraries` functions, in favour of using the `target_*` functions.

- Instead of setting flags globally through `CMAKE_C_FLAGS`, use `target_compile_options`.
- Instead of setting defines globally through `add_definitions`, use `target_compile_definitions`.
- Instead of using all the flags unconditionally, check for every flag if the compiler supports it.
- For fast-math, also check the MSVC-equivalent.

### Vendored Dependencies

We inherited from prboom-plus a piece of code that checks for a magic `dependencies_<arch>` directory at the root of git repo. This was mainly added as a means to consume the `windows_dependencies` archive provided upstream since getting dependencies for MSVC is difficult. Since we cannot just yank this out as people may rely on it and prefer it over using vcpkg, I've made it more explicit and hopefully user-friendly:

- Add checks, instead of just blindly adding the directory to `CMAKE_PREFIX_PATH`, check if it is an existing directory.
- Inform the user about it, with messages such as
  ```
  -- Checking for vendored dependencies in: "D:/git/dsda-doom/prboom2/../dependencies_x64".
  -- Found vendored dependencies.
  ```
- Make it controllable through the option `CHECK_FOR_VENDORED_DEPS`
- Turn this off outside of MSVC, or when using vcpkg.

### CMake Version Compatibility

With all the recent CMake changes, I wanted to ensure that the `cmake_minimum_required` still reflected the reality since CMake will not warn you about using features newer than the version you entered.

The only change I made regarding that was swapping `POP_FRONT` to `REMOVE_AT <list> 0` since it's functionally the same but drops the requirement from 3.15 down to 3.4.

On the topic of compatibility, since `TestBigEndian` module has been deprecated, I made a wrapper module that uses the new method for determining endianess available in CMake >= 3.20 and falls back to the deprecated module for older CMake versions. It's functionally the same, just more future-proof if CMake ever decides to drop the module on newer installs.

### Miscellaneous

- Rename `PRBOOM_` variables to `DSDA_`.
- Remove this block of dead code, since the check for `HAVE_SDL_JOYSTICKGETAXIS` has been removed, it is just a no-op.
  ```cmake
  set(CMAKE_REQUIRED_INCLUDES_PREV ${CMAKE_REQUIRED_INCLUDES})
  set(CMAKE_REQUIRED_LIBRARIES_PREV ${CMAKE_REQUIRED_LIBRARIES})
  set(CMAKE_REQUIRED_INCLUDES ${CMAKE_REQUIRED_INCLUDES} ${SDL2_INCLUDE_DIRS})
  set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} ${SDL2_LIBRARIES})
  set(CMAKE_REQUIRED_INCLUDES ${CMAKE_REQUIRED_INCLUDES_PREV})
  set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES_PREV})
  ```


kraflab:
> We inherited from prboom-plus a piece of code that checks for a magic dependencies_<arch> directory at the root of git repo. This was mainly added as a means to consume the windows_dependencies archive provided upstream since getting dependencies for MSVC is difficult. Since we cannot just yank this out as people may rely on it and prefer it over using vcpkg, I've made it more explicit and hopefully user-friendly

Actually, since we have moved away from the windows_dependencies artifact, I would suggest to clean all of that up as well. Maybe we need a small guide on how to set up the project for msvc after the changes. I think if anyone has been relying on a prebuilt batch of dependencies that haven't been updated in a long time, this is as good an excuse as any to migrate 😄

FtZPetruska:
Rebased on main to address the conflicts from https://github.com/kraflab/dsda-doom/commit/63c901d89e99cd6a4298a2643c1923beb5ab6e36 and took that opportunity to update the flag in this PR. Thank you for the heads-up!

> Actually, since we have moved away from the windows_dependencies artifact, I would suggest to clean all of that up as well. Maybe we need a small guide on how to set up the project for msvc after the changes. I think if anyone has been relying on a prebuilt batch of dependencies that haven't been updated in a long time, this is as good an excuse as any to migrate 😄

Got it, for now I just dropped the related commit. I will open a separate PR so that behaviour change comes with a full build guide for MSVC users.

kraflab:
What decides when the msvc action uses the cache? I noticed the current build is running the long one, but other commits from today used the cache.

kraflab:
And thanks again 🥳

kraflab:
Hm, I get `cc1plus.exe: warning: command-line option '-Wbad-function-cast' is valid for C/ObjC but not for C++` now when compiling, which I didn't get before (and similarly for a few other warnings). 🤔

kraflab:
Looking at the warnings popping up in cpp files, this is making me think possibly no flags were sent to cpp compilation before? 😅

FtZPetruska:
> Looking at the warnings popping up in cpp files, this is making me think possibly no flags were sent to cpp compilation before? 😅

Since `CMAKE_C_FLAGS` were set previously, it would make sense that the would not get the flags set, yeah 😅 I did not even notice that there were cpp files in the source tree.

> What decides when the msvc action uses the cache? I noticed the current build is running the long one, but other commits from today used the cache.

I'm honestly not sure about why these runs didn't get to read the cache, especially when previous runs on that PR did.

kraflab:
Here are the flags that don't work in cpp:

-Wbad-function-cast
-Wdeclaration-after-statement
-Wno-pointer-sign
-Wabsolute-value

I assume there's a way to have a c list and a cpp list? 😄

FtZPetruska:
> Here are the flags that don't work in cpp:
> 
> -Wbad-function-cast -Wdeclaration-after-statement -Wno-pointer-sign -Wabsolute-value
> 
> I assume there's a way to have a c list and a cpp list? 😄

There is! I was surprised to find there is no way of setting compile options per-language on a target, but there's a generator expression just for that!

I'll open a PR once I figure out a good way to go about it.

PR #276, opened 2023-03-25
Fix various incorrect format specifiers

bkoropoff:
This fixes various GCC warnings.  `%I64` is an MSVC-specific extension.  `long long int` and `long int` might have the same size on x86-64, but the compiler considers them distinct types and warns that you used the wrong specifier.  There is a standard `printf` specifier for `size_t` (`%zu`), but MSVC doesn't support it, so the cast in `oplplayer.c` is still required.

### DIFF
--- a/prboom2/src/dsda/exdemo.c
+++ b/prboom2/src/dsda/exdemo.c
@@ -410,7 +410,7 @@ static void DemoEx_GetFeatures(const wadinfo_t* header) {
   if (!str)
     return;
 
-  if (sscanf(str, "%*[^\n]\n%llx-%32s", &exdemo.features, signature) == 2) {
+  if (sscanf(str, "%*[^\n]\n%" PRIx64 "-%32s", &exdemo.features, signature) == 2) {
     byte features[FEATURE_SIZE];
     dsda_cksum_t cksum;
 
@@ -444,7 +444,7 @@ static void DemoEx_AddFeatures(wadtbl_t* wadtbl) {
   buffer_length = strlen(cksum.string) + strlen(description) + 24;
   buffer = Z_Calloc(buffer_length, 1);
 
-  snprintf(buffer, buffer_length, "%s\n0x%016llx-%s", description, features, cksum.string);
+  snprintf(buffer, buffer_length, "%s\n0x%016" PRIx64 "-%s", description, features, cksum.string);
 
   AddPWADTableLump(wadtbl, DEMOEX_FEATURE_LUMPNAME, buffer, buffer_length);
 
--- a/prboom2/src/MUSIC/oplplayer.c
+++ b/prboom2/src/MUSIC/oplplayer.c
@@ -1362,7 +1362,7 @@ static const void *I_OPL_RegisterSong(const void *data, unsigned len)
     // time numbers we have to traverse the tracks and everything
     if (mf.len < 100)
     {
-        lprintf (LO_WARN, "I_OPL_RegisterSong: Very short MIDI (%I64i bytes)\n", (long long)mf.len);
+        lprintf (LO_WARN, "I_OPL_RegisterSong: Very short MIDI (%llu bytes)\n", (unsigned long long)mf.len);
         return NULL;
     }
 
--- a/prboom2/src/info.h
+++ b/prboom2/src/info.h
@@ -5761,7 +5761,7 @@ typedef enum
  * Definition of the state (frames) structure                       *
  ********************************************************************/
 
-typedef int64_t statearg_t;
+typedef long long int statearg_t;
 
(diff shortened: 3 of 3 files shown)


kraflab:
There are no gcc warnings in msys2, so I guess it's a linux issue?
I decided against using those format defines at the time because they hurt readability, and my opinion unfortunately hasn't changed there.
The oplplayer change is good.

bkoropoff:
Yes, it's a Linux issue.  The defines are annoying, but the only portable way to write the code correctly.  Correctness is more important than readability.

kraflab:
Readability is more important than "correctness" unless the correctness actually means something. If you can show me how the code is actually not portable and doesn't work on Linux, I'll be inclined to figure out a solution. But otherwise the description tells me this is essentially an overzealous warning, and I'm not going to make the code harder to maintain just for that.

bkoropoff:
Being "-Wall" clean is important for maintainability.  If you let warnings pile up without fixing them, actual bugs get lost in the noise.  I can try to rework the code to not need the `PRI*` defines, but unfixed/disabled compiler warnings are generally a greater liability than making the compiler happy in the few cases where it's overzealous.

kraflab:
I agree, which is why it IS clean of warnings in my own dev environment. Having no warnings on any system in any compiler is out of scope to me. But if there are clean improvements that don't hurt my own development experience, then I won't block them either.

bkoropoff:
> I agree, which is why it IS clean of warnings in my own dev environment. Having no warnings on any system in any compiler is out of scope to me.

Sorting through more stringent warnings on other platforms has found several real bugs at this point.  Placating the compiler when a warning is harmless is annoying, but if you pay the cost and turn on `-Werror` you get to have these sorts of bugs caught for free in the future, either locally or when github runs a CI workflow.  It's a pretty good bargain.

kraflab:
I understand the tradeoffs. That's why there are certain warnings on and certain warnings disabled. We just disagree on where the line is 😉 

I appreciate you looking into this stuff regardless.

bkoropoff:
I reworked this to avoid using the `PRI*` macros, by the way

prboom2/src/dsda/console.c
```diff
@@ -1654,7 +1654,7 @@ static dboolean console_StateSetMisc2(const char* command, const char* args) {
 
 static dboolean console_StateSetArgs1(const char* command, const char* args) {
   int id;
-  statearg_t value;
```
kraflab: The whole point of having a `statearg_t` is to use it when we're dealing with state arguments, for consistency and clarity. Let's rather change the typedef for `statearg_t` to be a `long long`.

prboom2/src/dsda/exdemo.c
```diff
@@ -402,6 +402,7 @@ void dsda_MergeExDemoFeatures(void) {
 static void DemoEx_GetFeatures(const wadinfo_t* header) {
   char* str;
   char signature[33];
+  long long value;
```
kraflab: For the features, given this is a very explicit set of 64 bit flags being written and read, I'm not as bothered by the prospect of using the rigorous PRI defines here. Let's actually try them out in this file and if I hate it I'll change it later 😄

prboom2/src/dsda/console.c
bkoropoff: `statearg_t` is  `int64_t` because it's read from a file, right?  `long long` is 64 bits on most platforms, but it's not guaranteed, which is why this generates a warning in the first place.  If that's still OK, I can go ahead and change it.

kraflab: `statearg_t` are dehacked parameters, which are parsed from plain text and set with lines like this: `deh_state.state->args[0] = (statearg_t)value;`. Incidentally dehacked parameters are all 32 bit integers. The reason this is 64 bits is because some arguments are translated from dehacked values into port-specific flags, which can use higher bits. E.g., `states[i].args[1] = deh_translate_bits(states[i].args[1], deh_mobjflags_mbf21);`. So as long as this field has at least 64 bits it should all be fine.

That's assuming it won't introduce extra warnings when converting types here:
```c
void A_AddFlags(mobj_t* actor)
{
  uint64_t flags, flags2;

  if (!mbf21 || !actor)
    return;

  flags  = actor->state->args[0];
  flags2 = actor->state->args[1];

  actor->flags  |= flags;
  actor->flags2 |= flags2;
}
```

kraflab:
👍

prboom2/src/dsda/console.c
bkoropoff: It should be fine, I don't think anyone turns on warnings about implicit integer conversions in C because it's so prolific in practice.

PR #278, opened 2023-03-25
Fix GLU callback prototype warnings.

bkoropoff:
Explicitly cast function pointers to type specified by GLU.


kraflab:
Thanks

andrey-budko:
MSVC x86 build compilation error

> 'function': cannot convert from 'void (__cdecl *)()' to 'void (__stdcall *)()'

bkoropoff:
Windows and its 20 calling conventions :roll_eyes:

PR #279, opened 2023-03-25
Fix abs integer truncation warning

bkoropoff:
Alternatively, these types need to be changed to `int`/`int32_t`.  `long` is 64 bit on x86-64 Linux!

-----------------------

Use labs for long arguments.

### DIFF
--- a/prboom2/src/e6y.h
+++ b/prboom2/src/e6y.h
@@ -46,12 +46,12 @@
 
 typedef struct camera_s
 {
-  long x;
-  long y;
-  long z;
-  long PrevX;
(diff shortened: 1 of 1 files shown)


kraflab:
Could you rather convert the `camera_t` long fields into `fixed_t` for consistency?

bkoropoff:
Yeah, that was probably what was intended.

bkoropoff:
> Could you rather convert the `camera_t` long fields into `fixed_t` for consistency?

Done

kraflab:
Thanks 👌

PR #281, opened 2023-03-25
heretic:sb: fix misplaced paren

bkoropoff:
Not sure it matters since it only affects netgames, but it's a real logic bug that clang complains about.


kraflab:
Nice 👌

PR #282, opened 2023-03-25
gl_sky: fix precision loss detected by clang warning

bkoropoff:
Note sure how much of a difference this makes in practice, but some of these integer values were getting rounded off when cast to float (especially `ANGLE_MAX`).

-------------

Float doesn't have enough significand bits to represent large 32-bit integers, so use double instead.

### DIFF
--- a/prboom2/src/gl_sky.c
+++ b/prboom2/src/gl_sky.c
@@ -437,7 +437,7 @@ void gld_DrawScreenSkybox(void)
   {
     #define WRAPANGLE (ANGLE_MAX/4)
 
-    float fU1, fU2, fV1, fV2;
+    double fU1, fU2, fV1, fV2;
     GLWall *wall = &SkyBox.wall;
     angle_t angle;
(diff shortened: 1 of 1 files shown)


prboom2/src/gl_sky.c
```diff
@@ -437,7 +437,7 @@ void gld_DrawScreenSkybox(void)
   {
     #define WRAPANGLE (ANGLE_MAX/4)
 
-    float fU1, fU2, fV1, fV2;
+    double fU1, fU2, fV1, fV2;
```
kraflab, line 440: These variables are consumed as floats by the gl functions below, right? Won't they just lose the precision anyway?

bkoropoff: It's less loss overall to round at the end when calling gl than to use float for each intermediate result.

kraflab: 👍

PR #284, opened 2023-03-26
CMake: Avoid using C-only warnings on C++ files.

FtZPetruska:
Following #275, warnings were applied to C++ files. Unfortunately some of them do not apply to C++ with GCC providing the message: `warning: command-line option '-Wbad-function-cast' is valid for C/ObjC but not for C++`.

To address this, wrap the C-only warnings in the `$<$<COMPILE_LANGUAGE:C>:flag>` expression, so they only apply to C files.

Thanks @kraflab for pointing it out and providing the list of warnings affected!


kraflab:
Thanks for the quick action

PR #286, opened 2023-03-26
CMake: Better install on Windows

FtZPetruska:
On Windows, using the `GNUInstallDirs` does not play well with dsda. It is typical to have a flat layout instead such as:
```
dsda-doom/
|- dsda-doom.exe
|- dsda-doom.wad
```

Setting `X_VCPKG_APPLOCAL_DEPS_INSTALL` lets vcpkg copy the needed dlls when installing, this also applies to CPack since it uses the install command under the hood. For this to work properly, `CMP0077` needs to be set to NEW, and setting the variable must happen before the call to `project()`

Additionally, for CPack to work properly, `dsda-doom.wad` needs to be installed using a path relative to the install prefix instead of an absolute path. For this to work, it is preferable to limit `DSDAPWADDIR` to be a path under the install prefix (you probably would not want to install it outside of the prefix anyway).

This latter change **will** break buildscripts that set `DSDAPWADDIR` to a full path. But it's a trivial change (e.g. Homebrew's recipe will just need to change `#{libexec}` to `libexec`), and it is an option only really meant to be used by package managers.

### DIFF
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -162,8 +162,16 @@ if(WITH_PORTMIDI)
     endif()
 endif()
 
-set(DOOMWADDIR "${CMAKE_INSTALL_PREFIX}/share/games/doom" CACHE PATH "Path to look for WAD files")
-set(DSDAPWADDIR "${DOOMWADDIR}" CACHE PATH "Path to install DSDA-Doom internal WAD")
+if(WIN32)
+    set(DEFAULT_WAD_DIR ".")
+else()
+    set(DEFAULT_WAD_DIR "${CMAKE_INSTALL_DATADIR}/games/doom")
+endif()
+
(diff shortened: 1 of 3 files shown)


kraflab:
👍

kraflab:
By the way, the windows CI build seems to fail randomly now 😅 
The latest one:
```
fatal: unable to access 'https://github.com/Microsoft/vcpkg.git/': The requested URL returned error: 429
Error: Process completed with exit code 1.
```
Any idea what we can do about that?

rfomin:
> By the way, the windows CI build seems to fail randomly now 😅

From my experience, sometimes the vcpkg maser branch is broken, so I suggest removing this line: https://github.com/kraflab/dsda-doom/blob/cfb95f68614ef62c1e982612364776b373796238/.github/workflows/continuous_integration.yml#L83
vcpkg is updated every week in the windows-2022 image.

PR #287, opened 2023-03-26
CI: Do not manually update vcpkg.

FtZPetruska:
The `-latest` images already update vcpkg on a weekly basis, so no need to manually update it.

Thanks @rfomin for the tip.


kraflab:
Thank you both 😄

PR #288, opened 2023-03-26
Complete VBO implementation

bkoropoff:
What was there seemed complete except for textured automap, which needed some fixes.  I made using it a setting which is turned off by default for now.

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -448,25 +448,37 @@ void gld_MapDrawSubsectors(player_t *plr, int fx, int fy, fixed_t mx, fixed_t my
   dsda_GLSetScreenSpaceScissor(fx, fy, fw, fh);
   glEnable(GL_SCISSOR_TEST);
 
+  // Look "down" at the floor, since subsector flats extend along x and z coordinates
+  glRotatef(-90.0, 1.0f, 0.0f, 0.0f);
+
   if (automap_rotate)
   {
     float pivotx = (float)(fx + fw / 2);
     float pivoty = (float)(fy + fh / 2);
 
-    float rot = -(float)(ANG90 - viewangle) / (float)(1u << 31) * 180.0f;
+    float rot = (float)(ANG90 - viewangle) / (float)(1u << 31) * 180.0f;
 
     // Apply the automap's rotation to the vertexes.
-    glTranslatef(pivotx, pivoty, 0.0f);
-    glRotatef(rot, 0.0f, 0.0f, 1.0f);
-    glTranslatef(-pivotx, -pivoty, 0.0f);
+    glTranslatef(pivotx, 0.0f, pivoty);
+    glRotatef(rot, 0.0f, 1.0f, 0.0f);
+    glTranslatef(-pivotx, 0.0f, -pivoty);
   }
 
   glTranslatef(
     (float)fx - (float)mx / (float)FRACUNIT * (float)scale / (float)FRACUNIT,
-    (float)fy + (float)fh + (float)my / (float)FRACUNIT * (float)scale / (float)FRACUNIT,
-    0);
+    0,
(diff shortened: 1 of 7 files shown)


kraflab:
Is there a noticeable difference with this?

bkoropoff:
I haven't measured it.  It's technically free performance (no vertex copying to video memory every frame), but only for flats.  Doom doesn't exactly stress remotely recent video hardware, it's mostly CPU-bound.  Maybe maps with tons of sector detailing like sunder might benefit on older hardware.

andrey-budko:
180 vs 250 fps for me on Deus Vult map 05 with "indexed" light mode, and 240 vs 330 with "glboom"

kraflab:
I tested this out on the eviternity demo I use as a "normal gameplay" benchmark and went from 1500 to 1600 fps 🙂 
Do you have any concerns with making it always on and dropping the option?

kraflab:
Also went from 124 to 142 on eviternity map 26 at the beginning (this is an area where a lot of people struggle with getting good fps), very nice

bkoropoff:
I knew it would help in theory, but I guess Doom stresses GPU memory bandwidth more than you'd think.

I made it an option out of caution in case it breaks someone's weird GPU/driver combination somewhere, but there's no particular reason not to always use it otherwise.

andrey-budko:
> I tested this out on the eviternity demo

Should't this affect only automap?  I tested with DV map 5 \<tab\> IDDT

bkoropoff:
> Should't this affect only automap? I tested with DV map 5 <tab> IDDT

It was compiled out before, so it affects all flat drawing.  The automap just needed to be tweaked to work with it.

andrey-budko:
btw
> 2.5.0.3  @ 2009-Jul-11
> [*] Automap drawing optimisation in GL mode: 220 fps instead of only 30 on map05 @ dv.wad after IDDT IDDT.

bkoropoff:
I changed the setting to be on by default, but it's still there as a chicken switch.

prboom2/src/gl_main.c
```diff
@@ -448,25 +448,37 @@ void gld_MapDrawSubsectors(player_t *plr, int fx, int fy, fixed_t mx, fixed_t my
...
     float pivoty = (float)(fy + fh / 2);
 
-    float rot = -(float)(ANG90 - viewangle) / (float)(1u << 31) * 180.0f;
+    float rot = (float)(ANG90 - viewangle) / (float)(1u << 31) * 180.0f;
 
     // Apply the automap's rotation to the vertexes.
-    glTranslatef(pivotx, pivoty, 0.0f);
-    glRotatef(rot, 0.0f, 0.0f, 1.0f);
-    glTranslatef(-pivotx, -pivoty, 0.0f);
+    glTranslatef(pivotx, 0.0f, pivoty);
```
kraflab, line 462: Just want to double check, do the changes to the y/z axes here require any changes in the transforms below? (https://github.com/kraflab/dsda-doom/pull/288/files#diff-9467dec909417e0c3a69bffbdd2b00caac451893686c9f078d619bcd054e2fe4R535-R537)

kraflab: Double checked this stuff in my test map with scaled, translated, and rotated flats, all looks good.

kraflab:
We'll have a bunch of beta testers for 0.26 like we had in 0.25, so hopefully if there's an issue we'll catch it then, thanks again 🥳 

Is there a similar improvement that can be made for textures? 😄

andrey-budko:
> Is there a similar improvement that can be made for textures? 😄

This is what JC did for iPhone:
https://github.com/id-Software/DOOM-iOS/blob/master/code/iphone/iphone_render.c#L1327

andrey-budko:
When I tried something similar, I didn't get any noticeable difference on PC iirc. Shouldn't be hard to try again and recheck.

PR #289, opened 2023-03-26
Remove SDL_windows_main.

FtZPetruska:
When building with UCRT64 on MSYS2, I was getting errors about multiple definitions of WinMain.

Judging by the content of `src/SDL/SDL_windows_main.c`, it seems to be a carbon copy of [this file from the SDL project](https://github.com/libsdl-org/SDL/blob/SDL2/src/main/windows/SDL_windows_main.c). But since we're already linking to SDL2main, it is redundant at best.

Removing it also seems to fix the "ambiguous entry point" warning on MSVC.


kraflab:
Was this code ever reached?

FtZPetruska:
Not sure why CI failed, it reportedly failed at building libogg, while the run on my fork [ran fine](https://github.com/FtZPetruska/dsda-doom/actions/runs/4527012516).

> Was this code ever reached?

It was added to the sources on `WIN32` builds and effectively acting as a wrapper around the main function in `i_main.c`.

kraflab:
How confident are you that none of those defines are relevant? 😅

FtZPetruska:
> How confident are you that none of those defines are relevant? 😅

I am not sure I understand the question.

The `SDL_windows.h` was only included by the `SDL_windows_main.c` file. And the latter only provides a wrapper for the different entry points on Windows. So effectively, the preprocessor definitions in these two files are never used anywhere else.

From what I can see, the `SDL_windows.h` is also a copy of an SDL header, specifically `SDL/src/core/windows/SDL_windows.h`, an internal header `SDL_windows_main.c` depends on.

kraflab:
My question didn't make sense tbh 🙃 

My concern is if any of that wrapping functionality is going to be missed. What makes it redundant? Is it implicitly happening thanks to sdl somewhere?

FtZPetruska:
This piece of code is pretty much the content of the `SDL2main` library on Windows. Since both the Find module and the upstream config already link to said `SDL2main`, that is why I would call it redundant.

I should probably have done that earlier, but here is the specific error I'm getting (stripped paths for readability):
```
x86_64-w64-mingw32/bin/ld.exe: lib/SDL2main.lib(SDL_windows_main.obj): multiple definition of `WinMain';
dsda-doom/prboom2/src/SDL/SDL_windows_main.c:104: first defined here
```

kraflab:
Ok got it, thanks 👍

FtZPetruska:
No worries, sorry for the confusion 😅

PR #290, opened 2023-03-27
Add Windows build docs and settings.

FtZPetruska:
This PR adds documentation for building on Windows under the following configuration:
- MSVC
  - Using Visual Studio
  - Using Visual Studio Code
  - Manually from the terminal
- MSYS2 from the terminal

In order for Visual Studio to work out of the box, I committed its `CMakeWorspaceSettings.json` so VS knows it's a CMake project and where to look for the CMakeLists. Since it seems to be the only purpose of this file, it does not risk clashing with users' local config.

For Visual Studio Code, the same could not be done since all workspace-wide settings are done in the same file. I did add extension recommendations to get syntax highlighting for C/C++ and CMake integration.

Lastly, I took out the CMake code that added the `dependencies_<arch>` to the `CMAKE_PREFIX_PATH`. Documenting why it used to be there, why it was removed, and what are the alternatives.

Let me know what you think!


docs/guides/building_on_windows.md
```diff
@@ -0,0 +1,202 @@
...
+
+```
+git clone https://github.com/kraflab/dsda-doom.git
+```
+
+Run the CMake configuration:
+
+```
+cd dsda-doom
+cmake -Sprboom2 -Bbuild -DCMAKE_TOOLCHAIN_FILE="C:\vcpkg\scripts\buildsystems\vcpkg.cmake" -DCMAKE_BUILD_TYPE=Release
```
rfomin: Why do you recommend the `Release` configuration? Developers probably want `Debug`. Also you are using MS Visual Studio generator here, it will configure both `Debug` and `Release` anyway.

docs/guides/building_on_windows.md
```diff
@@ -0,0 +1,202 @@
...
+
+```
+cd dsda-doom
+cmake -Sprboom2 -Bbuild -GNinja -DCMAKE_BUILD_TYPE=Release
+```
+
+And finally, build the project:
+
+```
+cmake --build build --config Release
```
rfomin, line 168: I think it worth noting that packaging won't work in MSYS2.

FtZPetruska: You're correct, I've looked into using the CMake facilities to get and install the runtime dependencies, but for libraries such as DUMB and LibMad which don't have a CMake config, or Vorbis which CMake config is not shipped by MSYS2 this mechanism fails.

I will just move that section up to the vcpkg one and add a note about copying DLLs for MSYS2.

docs/guides/building_on_windows.md
FtZPetruska: I used `Release` to match what the macOS build guide uses. I will go ahead and add a note about replacing Release with the other configs if needed.

kraflab:
Awesome stuff, thanks 🙂

PR #292, opened 2023-03-29
Add zip support for `-file`

### LINKED ISSUE #133: feature request: zip support
I check the patch notes from time to time and 0.25 is shaping up to be a pretty major release... 
Would you consider adding zip archive support, too?

I think it would be a pretty handy feature: Woof has it and being able to load the archive downloaded from dsda/idgames directly has pampered me a little bit lately..

Thank you for all the work on this port!

FtZPetruska:
Fixes #133

---

This PR adds zip support using [libzip](https://libzip.org/).

The logic used is as follow:
- When processing the files passed to `-file`, consider files ending with `.zip`
- Use `I_FindZip` (forwarded to `I_FindFile`) to ensure the file exists.
- Get a system-specific temporary directory
- Unpack the archive to that directory
- Pass the directory to `AutoLoadWADs` and `AutoLoadPatches` to discover WADs and dehacked files.

---

The implementation consists in two modules:
- tempdir, to get OS-specific temporary directories.
- zipfile, to unpack zip archives to a directory.

The `tempdir` module will use `mkdtemp` on *nix systems that provide it, `GetTempPath(2)` on Windows, or fallback to the current directory. You may notice the large block of checks for `mkdtemp` in the CMakeLists, this is mainly due to glibc having different needs to make the function available, and macOS having it in a completely different header.

I preferred adding this modules over using `I_GetTempDir` as the latter makes the assumption that all non-Windows platform use `/tmp` which is not necessarily correct.

The `zipfile` module uses `libzip` to access and decompress the zip file to the given directory. I have added all the necessary stuff for CMake to find and consume libzip and added it to the CI.

There is an odd packaging issue on Debian/Ubuntu where the CMake config shipped with `libzip-dev` contains executable targets that are not automatically installed, hence why `zipcmp`, `zipmerge`, and `ziptool` are also manually installed.

The vcpkg build of libzip has no encryption support since you presumably would not want to share your WADs in password-protected archives. But it does build additional support for the gzip and lzma algorithm. I did not enable zstd support due to issues with consuming the libzip in CMake.

There were several alternatives I considered:
- miniz, used by Woof, provides Zlib's inflate/deflate, zip read/write, and PNG writing. While it has no dependency, it requires either its source to be committed to this repo or using vcpkg.
- minizip(-ng), seems to be on-par feature-wise with libzip, the main caveat with this one is that there are two "version", the original `minizip` and its continuation `minizip-ng`, but distros like Debian are still stuck on the old `minizip` version.

---

As for testing, I tried to following use cases:
- pass an inexistant zip file -> ignored
- pass an unrelated zip file -> decompressed, but nothing loaded
- pass a valid zip file in the current directory with `-file ./foo.zip` -> successfully loaded
- pass a valid zip file in the current directory with `-file foo.zip` -> successfully loaded
- pass a valid zip file in `DOOMWADDIR` with `-file foo.zip` -> successfully loaded
- pass a valid zip file in parent directory with `-file ../foo.zip` -> successfully loaded


rfomin:
> I preferred adding this modules over using `I_GetTempDir` as the latter makes the assumption that all non-Windows platform use `/tmp` which is not necessarily correct.

Just curious, what OS besides Windows doesn't have `/tmp`? `I_GetTempDir` has been used by Choco/Crispy and PrBoom+ for many years. I think such an OS will require some porting work anyway.

FtZPetruska:
> Just curious, what OS besides Windows doesn't have `/tmp`? `I_GetTempDir` has been used by Choco/Crispy and PrBoom+ for many years. I think such an OS will require some porting work anyway.

It's not so much that `/tmp` does not exists, but canonically, you're supposed to look up the value of `TMPDIR`. For example on my machine `TMPDIR` is some directory under `/var`. Plus in the implementation, if no environment is found it will default to `/tmp`.

prboom2/src/dsda/tempdir.c
```diff
@@ -0,0 +1,185 @@
...
+/* Mingw requires these two headers first */
+#include <tchar.h>
+#include <windows.h>
+
+#include <inttypes.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strsafe.h>
+#if defined(HAVE_GET_TEMP_PATH_2)
+#define winapi_GetTempDir GetTempPath2
```
rfomin: We don't define `UNICODE`, so we must use the `GetTempPath2W`/`GetTempPathW` function to support non-Latin paths. Typically `%TEMP%` is found in the user's directory on Windows, with an often non-Latin name. I recommend just using `tempdir = M_getenv("TEMP");` it works for us in Woof/Crispy. See: https://github.com/chocolate-doom/chocolate-doom/blob/981da81f2adabd6eb82f1c3fc4ea27baecfccd82/src/m_misc.c#L572

rfomin:
We also have the `I_GetTempDir` function declared in `i_system.h`, but it is not used anywhere.

prboom2/src/dsda/tempdir.c
FtZPetruska: Heh, you're probably right. I'll just go ahead and use the existing constructs instead of making a whole new module.

prboom2/src/SDL/i_system.c
```diff
@@ -513,3 +513,8 @@ char* I_FindDeh(const char* wfname)
 
   return I_FindFile(wfname, ".deh");
 }
+
+char* I_FindZip(const char* wfname)
```
kraflab, line 522: Please add an `I_RequireZip` as well so we have consistent interface options.

prboom2/src/d_main.c
```diff
@@ -1510,6 +1512,29 @@ static void EvaluateDoomVerStr(void)
   lprintf(LO_INFO, "Playing: %s\n", doomverstr);
 }
 
+static void D_AddZip(const char* zipped_file_name)
+{
+  dsda_string_t temporary_directory = { NULL, 0 };
+  char* full_zip_path = I_FindZip(zipped_file_name);
```
kraflab: Please use the aforementioned `I_RequireZip` here. It's dsda-doom policy to abort if a requested file doesn't exist (and then you won't need the null check).

prboom2/src/d_main.c
```diff
@@ -1510,6 +1512,29 @@ static void EvaluateDoomVerStr(void)
   lprintf(LO_INFO, "Playing: %s\n", doomverstr);
 }
 
+static void D_AddZip(const char* zipped_file_name)
+{
+  dsda_string_t temporary_directory = { NULL, 0 };
```
kraflab: General comment for the whole PR. I generally prefer code to be written like this:
```c
char* full_zip_path;

full_zip_path = I_FindZip(zipped_file_name);
```
I.e., split the list of variables from any code. The code base isn't consistent on this point but I try to do this for new code.

kraflab: Also, please leave out the `= { NULL, 0 }` here and leave the reliance on the init function. I'm not sure what I will do with this dsda_string_t in the future, but the goal is to not have any fragile code related to it (e.g., this constant struct might end up being wrong in the future).

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,135 @@
...
+// DESCRIPTION:
+//	DSDA zipfile support using libzip
+//
+
+#include "config.h"
+
+#include "../lprintf.h"
+
+#ifdef HAVE_LIBZIP
+#include "../m_file.h"
```
kraflab: Please organize headers and paths like so for consistency:
```c
#include <stdio.h>
#include <zip.h>

#include "m_file.h"
#include "z_zone.h"

#include "dsda/utility.h"
```

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,135 @@
...
+//
+// DESCRIPTION:
+//	DSDA zipfile support using libzip
+//
+
+#include "config.h"
+
+#include "../lprintf.h"
+
+#ifdef HAVE_LIBZIP
```
kraflab: Unless there is a very good reason, I would prefer to not add _optional_ dependencies but rather require them. I want people to have a consistent experience when using the port and something like loading zip files I would see as a basic element of the port. You can try to convince me otherwise though if you feel strongly about making it optional.

prboom2/src/SDL/i_system.c
FtZPetruska: Added in [640f573](https://github.com/kraflab/dsda-doom/pull/292/commits/640f573eb115a4b37863151e1b45dfb7cec4d62f).

prboom2/src/d_main.c
FtZPetruska: Done in [607c428](https://github.com/kraflab/dsda-doom/pull/292/commits/607c42895cd79d3bb4f9dc66b98c79a357c0e835).

prboom2/src/dsda/zipfile.c
FtZPetruska: Oh I have nothing against making it a requirement, the package is widely available anyway 😁

Changed in [285a7eb](https://github.com/kraflab/dsda-doom/pull/292/commits/285a7ebc8638a6ff5fce2442fc536988134ef11f).

prboom2/src/d_main.c
FtZPetruska: Got it 👍 

I made the style changes in [9fb2b1f](https://github.com/kraflab/dsda-doom/pull/292/commits/9fb2b1f4ac196728626ebe9b0780f85d67da66bb) and fixed the `{ NULL, 0 }` issue in [607c428](https://github.com/kraflab/dsda-doom/pull/292/commits/607c42895cd79d3bb4f9dc66b98c79a357c0e835).

prboom2/src/dsda/zipfile.c
FtZPetruska: Done in [9fb2b1f](https://github.com/kraflab/dsda-doom/pull/292/commits/9fb2b1f4ac196728626ebe9b0780f85d67da66bb).

prboom2/src/d_main.c
```diff
@@ -1510,6 +1512,27 @@ static void EvaluateDoomVerStr(void)
...
+  dsda_InitString(&temporary_directory, I_GetTempDir());
+  dsda_StringCatF(&temporary_directory, "%c%s%c",
+                  PATHNAME_SEPARATOR,
+                  dsda_BaseName(zipped_file_name),
+                  PATHNAME_SEPARATOR);
+  M_MakeDir(temporary_directory.string, false);
+
+  dsda_UnzipFile(full_zip_path, temporary_directory.string);
+
+  AutoLoadWADs(temporary_directory.string);
```
kraflab: When someone records a demo, it adds wad and deh files used to the footer of the demo file so that the launcher can autodetect the needed files to play back the demo and load them without prompting the user to choose a file.

It does not add autoload files (because these tend to be things NOT required to play back a demo, like cosmetic changes).

But the files inside a zip need to be treated as "real files" so that they 1) get added to the footer of recordings and 2) trigger loading of autoload files for the wads within.

You'll need to do two things to sort this out:

1) The AutoLoadWADs function needs to take an argument that chooses the source it assigns to files (currently it's `source_auto_load`, but for this use case we need `source_pwad`). Most likely this will need a rename like `LoadWADsAtPath`.

2) Instead of auto loading deh patches, we need to add them to the `-deh` command so they are registered as "real files" as well. So instead of `ProcessDehFile` in `AutoLoadPatches` we would need to do `dsda_AppendStringArg(dsda_arg_deh, filename)` (this is the same thing that happens when it detects a deh file in the `-file` command). I guess an option is to add a flag to `AutoLoadPatches` for whether the dehs should be processed immediately or not, and then give it a corresponding name like `LoadDehackedFilesAtPath`.

prboom2/src/doomtype.h
```diff
@@ -116,6 +116,12 @@ typedef unsigned char byte;
   #define INLINE inline        /* use standard inline */
 #endif
 
+#ifdef WIN32
+#define PATHNAME_SEPARATOR '\\'
```
kraflab: Does this actually matter? I ask because there's a lot of code in this code base that builds paths with `/` and it seems to work fine 😄 

But maybe it doesn't work on every system, I don't know.

bkoropoff: DOS and Windows have always accepted `/` as a separator at the API level, but individual programs can choke on them due to using `/` to start command line switches, etc.  It's probably fine if it goes straight into a Win32 API call, but you could run into trouble if you pass it to a 3rd-party library which does something unexpected like pass it to an external program as a parameter.

FtZPetruska: Seems like libzip does not make use of external programs, so I'll just revert that change. Thanks for the insight 👍 

---

EDIT: Reverted in [d295b81](https://github.com/kraflab/dsda-doom/pull/292/commits/d295b81125fef9ce0969213e94f04fe49cb65c59).

prboom2/src/d_main.c
FtZPetruska: Thank you for the suggestions, this seems to work as expected (record a demo by passing only a zip file and dsda automatically adding the pwads/deh patches when playing it back).

1. Changed in [ddc6b4b](https://github.com/kraflab/dsda-doom/pull/292/commits/ddc6b4bd24dd62198c1ee8fe896fdc3674d4e510). Renamed to `LoadWADsAtPath` as you suggested. Made the function take the `wad_source_t` as an argument. Replaced all the existing calls with `source_auto_load` and made the one in `D_AddZip` use `source_pwad`.
2. Changed in [8e7b5a9](https://github.com/kraflab/dsda-doom/pull/292/commits/8e7b5a94547d8f777b76ff76b74619ab86cf6629). Renamed to `LoadDehackedFilesAtPath` as you suggested. Similar idea as above, I added a bool argument indicating whether loading should be deferred or processed immediately.

prboom2/src/dsda/zipfile.h
```diff
@@ -0,0 +1,29 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// DESCRIPTION:
+//	DSDA zipfile support using libzip
+//
+
+#ifndef DSDA_ZIPFILE
```
kraflab: Please use `__DSDA_ZIPFILE__` for consistency with the other header files. I'm aware that using underscores at the front of these macros is "wrong" but for now I'd like to keep the standard.

prboom2/src/dsda/zipfile.h
```diff
@@ -0,0 +1,29 @@
...
+#define DSDA_ZIPFILE
+
+/**
+ * Unzips the file into the destination
```
kraflab: You can remove this block, it's not that kind of code base :^)

prboom2/src/dsda/zipfile.h
```diff
@@ -0,0 +1,29 @@
+//
+// Copyright(C) 2022 by Pierre Wendling
```
kraflab: 2023 :^)

prboom2/src/d_main.c
```diff
@@ -1510,6 +1519,24 @@ static void EvaluateDoomVerStr(void)
...
 
+static void D_AddZip(const char* zipped_file_name)
+{
+  dsda_string_t temporary_directory;
+  char* full_zip_path;
+
+  full_zip_path = I_RequireZip(zipped_file_name);
+  dsda_InitString(&temporary_directory, I_GetTempDir());
+  dsda_StringCatF(&temporary_directory, "/%s/", dsda_BaseName(zipped_file_name));
+  M_MakeDir(temporary_directory.string, false);
```
kraflab: The second argument of `M_MakeDir` says whether or not this code _requires_ the directory to be successfully created, so we had better send `true`. Otherwise this will return the error without handling it.

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+
+static void dsda_WriteZippedFilesToDest(zip_t *archive, const char *destination_directory) {
+  for (zip_int64_t i = 0; i < zip_get_num_entries(archive, ZIP_FL_UNCHANGED); i++) {
+    zip_file_t *zipped_file;
+    zip_stat_t stat;
+    FILE *dest_file_handle;
+    const char *file_name = zip_get_name(archive, i, ZIP_FL_UNCHANGED);
+
+    /* Intermediate directories show up as file */
+    if (dsda_HasFileExt(file_name, "/")) {
```
kraflab: Do zips always use `/` when read using this library or might it be `\` as well? There is a function `dsda_BaseName` that returns the file name without leading directories and accounts for both slashes. That would return a pointer to the null terminator of the file name if it ends in a slash.

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+  fwrite(buffer, sizeof(char), data_size, dest_file);
+  Z_Free(buffer);
+}
+
+static void dsda_WriteZippedFilesToDest(zip_t *archive, const char *destination_directory) {
+  for (zip_int64_t i = 0; i < zip_get_num_entries(archive, ZIP_FL_UNCHANGED); i++) {
+    zip_file_t *zipped_file;
+    zip_stat_t stat;
+    FILE *dest_file_handle;
+    const char *file_name = zip_get_name(archive, i, ZIP_FL_UNCHANGED);
```
kraflab, line 64: Probably it will save some room if we calculate the path and then pass it to functions below, rather than calculating it in each function:
```c
dsda_string_t full_path;

dsda_StringPrintF(&full_path, "%s%s", destination_directory, file_name);

...

dsda_CreateSubDir(full_path);

...

dsda_CreateFile(full_path);

...

dsda_FreeString(&full_path);
```

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+  full_path = dsda_JoinPath(directory, filename);
+  if (full_path == NULL) {
+    return NULL;
+  }
+  file_handle = M_OpenFile(full_path, "wb");
+  Z_Free(full_path);
+  return file_handle;
+}
+
+static void dsda_WriteContentToFile(zip_file_t *input_file, FILE *dest_file, zip_uint64_t data_size) {
```
kraflab: Take a look at `dboolean M_WriteFile(char const *name, const void *source, size_t length)` in `m_file`. You can allocate the buffer, call `zip_fread`, and then call `M_WriteFile` using the path, rather than creating a file separately. `M_WriteFile` also manages removing the file if there's an error writing it. You should probably also check if the return value matches the expected length and raise an error if not.

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+    /* Intermediate directories show up as file */
+    if (dsda_HasFileExt(file_name, "/")) {
+      dsda_CreateSubDir(destination_directory, file_name);
+      continue;
+    }
+
+    zip_stat_index(archive, i, ZIP_FL_UNCHANGED, &stat);
+    if ((stat.valid & ZIP_STAT_SIZE) == 0) {
+      lprintf(LO_ERROR, "dsda_WriteZippedFilesToDest: Failed to read size of zipped file %s.\n", file_name);
+      return;
```
kraflab: If anything goes wrong in this process, we can call `I_Error` for issues instead of using `lprintf`, since it's a requirement that files coming through `-file` do exist. The `I_Error` function shows an error popup on windows and prints to the terminal otherwise, and does not return. Maybe at some point there will be a case for zip files where it's acceptable that the process fails, and we would need to add an option to trigger or not trigger `I_Error`, but I don't think we need to manage that right now.

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+  file_handle = M_OpenFile(full_path, "wb");
+  Z_Free(full_path);
+  return file_handle;
+}
+
+static void dsda_WriteContentToFile(zip_file_t *input_file, FILE *dest_file, zip_uint64_t data_size) {
+  char *buffer;
+
+  buffer = (char *)Z_Malloc(data_size);
+  if (buffer == NULL) {
```
kraflab: Zone memory functions handle errors themselves. It is impossible to unsuccessfully return from them.

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+  }
+  file_handle = M_OpenFile(full_path, "wb");
+  Z_Free(full_path);
+  return file_handle;
+}
+
+static void dsda_WriteContentToFile(zip_file_t *input_file, FILE *dest_file, zip_uint64_t data_size) {
+  char *buffer;
+
+  buffer = (char *)Z_Malloc(data_size);
```
bkoropoff: Typically you would use a fixed-size buffer and read/write in a loop to avoid loading the entire contents into memory at once, which could be a substantial amount of data for a wad with lots of texture replacements, ogg/mp3 music, etc.

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    dest_file_handle = dsda_CreateFile(destination_directory, file_name);
+    if (dest_file_handle == NULL) {
+      lprintf(LO_ERROR, "dsda_WriteZippedFilesToDest: Failed to create destination file %s%s.\n", destination_directory,
+              file_name);
+      zip_fclose(zipped_file);
+      return;
+    }
+    dsda_WriteContentToFile(zipped_file, dest_file_handle, stat.size);
```
bkoropoff: You should refuse to load content over a certain cumulative size, e.g. more than 1GB of data from the same zip, to defend against [zip bombs](https://en.wikipedia.org/wiki/Zip_bomb).

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+
+static void dsda_WriteContentToFile(zip_file_t *input_file, FILE *dest_file, zip_uint64_t data_size) {
+  char *buffer;
+
+  buffer = (char *)Z_Malloc(data_size);
+  if (buffer == NULL) {
+    lprintf(LO_ERROR, "dsda_WriteContentToFile: Could not allocate temporary buffer.\n");
+    return;
+  }
+  zip_fread(input_file, buffer, data_size);
```
bkoropoff: Need to check for error (`-1`) here

prboom2/src/dsda/zipfile.c
FtZPetruska: Yes, after testing with various archives made on Windows and macOS, I observed that:
- The path separator is always `/`
- The name of a directory always ends with a `/`

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,134 @@
...
+  fwrite(buffer, sizeof(char), data_size, dest_file);
```
bkoropoff: Technically need to check for error here as well

prboom2/src/dsda/zipfile.h
FtZPetruska: My editor sort of freaks out with the leading `__`, but I can live with it 😛

prboom2/src/dsda/zipfile.c
FtZPetruska: I went ahead and used a fixed-sized buffer instead as suggested by @bkoropoff.

prboom2/src/dsda/zipfile.c
```diff
@@ -0,0 +1,109 @@
...
+
+    dsda_WriteContentToFile(zipped_file, dest_file, stat.size);
+
+    zip_fclose(zipped_file);
+    fclose(dest_file);
+    dsda_FreeString(&full_path);
+  }
+}
+
+void dsda_UnzipFile(const char *zipped_file_name, const char *destination_directory) {
```
kraflab, line 95: We better set `total_bytes_read` to zero each time this function is called, right? In case we are loading multiple zip files.

FtZPetruska: I wasn't sure whether to make `total_bytes_read` a global limit for all archives, or a per-archive limit. But I figured it makes more sense to reset it for every archive.

kraflab (approved):
Awesome stuff, thanks 👍

PR #293, opened 2023-03-29
[EXPERIMENTAL] GL renderer overhaul

### LINKED ISSUE #247: GL glitch on requiem MAP13
This occurs on 0.25.6, so it's not new:

doom00

### LINKED ISSUE #254: [GL] Seam between floor polygons in kdikdizd MAP15
Probably something to do with precision loss when triangulating.

seam

bkoropoff:
Summary:

- Import ajbsp and use it to build and cache GL verts/segs/ssects/nodes (for rendering only)
- Do some basic analysis on top of GL nodes
    - Coalesce subsectors into connected chunks (for rendering efficiency)
    - Detect rendering hacks
    - Compute potential flat bleeds
    - Generate cache-efficient structures for rendering
    - Generate portals (view apertures) between chunks
- Rework GL render path
    - Separate r_bsp/gl_bsp entirely
    - Triangulate and render flats based on *chunks*, transitively connected subsectors in a sector
    - Emulate flat bleeding
    - Render using chunks and portals instead of BSP nodes
        - BSP nodes are still used to locate initial chunk
        - A chunk is now the smallest piece of texture that becomes visible on the textured automap at a time.  This is better in that the shapes are more sensible (node builders often spit out weird subsectors), but worse in that they are less granular (large patches of texture can be revealed at once depending on how the map is designed, although this applies to subsectors to some extent as well).
    - Aggressively optimize CPU cache usage
- BSP internals dumping and plotting features to figure out what's going on

The following problematic maps seem to render correctly so far:
- AV MAP01
- HR2 MAP01 (actually beats gzdoom, which lacks compat hacks for HR2)
- Requiem MAP13 (fixes https://github.com/kraflab/dsda-doom/issues/247)
- KDiKDiZD MAP15 (doesn't fix https://github.com/kraflab/dsda-doom/issues/254 though :upside_down_face:)
- Eternal MAP02 (granular flat bleeding avoids https://github.com/kraflab/dsda-doom/issues/256)

Known problems:
- [x] Bleeds need to be dynamically rechecked for height (and maybe texture) changes
- [x] Bleeds not present at map start that might emerge due to a height change need to be considered
- [x] Poly objects are not handled at all
- [x] ajbsp leaks memory
- [ ] Performance problems on Sunder MAP14
    * Too many walls are being rendered.  Not sure if this is a bug or just a pathological case for the portal rendering algorithm.  Without #324 implemented, failing to occlude walls can be very costly.  With it, it would probably be better to take the overdraw rather than spend CPU time trying to eliminate it.

TODO:

- [x] Split `subsector_t`
- [x] Split `component_t`
- [x] Find a new name for "components"
- Optimize analysis passes on large maps
    - [x] Only process tagged sectors if a linedef could cause ceiling/floor height changes
    - [x] Multi-threaded analysis
- [x] Harmonize with stencil bleeding (some edge case regressions may remain)
- [x] Break up code more


andrey-budko:
17 fps instead of 90 in open area of FrozenT.wad

https://i.imgur.com/n79MJDY.png

image

https://www.doomworld.com/idgames/levels/doom2/Ports/d-f/frozent

bkoropoff:
@andrey-budko 

The number of planes is actually a red herring caused by flats being rendered a subsector at a time.  With VBO enabled the GL call overhead is not that bad, and the number of triangles involved isn't going to make a modern GPU break a sweat.  I reduced the count by allowing more subsectors to be coalesced, but the real bottleneck is computing the bleeds on the fly.  I thought I could get away with it since GZDoom does it, but I guess not.

I'm not sure about your system, but that map is badly CPU cache throttled on mine.  There appears to be a lot of room for improvement in general just by optimizing rendering structures.

andrey-budko:
> but the real bottleneck is computing the bleeds on the fly

yeah, with empty gld_DrawScene fps icreased from 17 to 22

bkoropoff:
I moved as much of the bleed calculation as possible out of the rendering path.  If someone actually depends on bleeds changing as sectors change floor/ceiling height, that will have to be addressed in the future.  I also optimized some other hot spots to try to gain some more CPU headroom:

- Use an array instead of linked list for the deferral queue
- Bypass the visplane hash entirely for GL

andrey-budko:
much better, only 10% slowdown on FrozenT

andrey-budko:
I tried to restore display lists stuff from the earlier version of prboom+ (single glCallList for floor) - no fps improvement.

Merging floors with the same gltexture and z into single glDrawElements call also without a luck.

bkoropoff:
Noclipped onto a vista in FrozenT to measure perf:

Screenshot from 2023-03-30 03-34-21

I actually get 60 FPS (capped) here normally, taking the screenshot tanked it temporarily.

Here's what CPU hotspots looks like on my system from this position:

Screenshot from 2023-03-30 03-15-53

Drilling down into one of them:

Screenshot from 2023-03-30 03-16-37

I think the cost is being attributed to the instruction after the actual culprit for some reason.  Some sort of quirk with the `perf` tool on Linux, or HW perf counters on my CPU.  Not sure.  Switching to cache misses suggests what the problem is:

Screenshot from 2023-03-30 03-17-52

So there's tons of cache misses on reading each `seg_t` in the subsector for the first time.  Simply *using* GL nodes makes this worse because we'll take misses and cache evictions on minisegs even though they don't get rendered.  In fact, most segs are minisegs in GL subsectors.  A preprocessing step to filter minisegs out of the rendering arrays might help.

Aside from `R_PointToPseudoAngle`, which has a hot spot in a floating point division, the other worst offenders are also cache misses chasing pointers to segs/lines/vertices, e.g.:

Screenshot from 2023-03-30 03-28-46

(Again, pretend the hot spot shows up on the instruction one above).  Cache misses on touching each `line_t` for the first time.  Another one:

Screenshot from 2023-03-30 03-32-19

This is taking a double whammy chasing the `line_t` and then each `vertex_t`.  Inlining the vertexes into the `line_t` might help.

Are you also CPU limited running this map?  Do you see a single CPU core being pegged?

andrey-budko:
I got random errors during loading -file eternal.wad -warp xx

LoadGLSegs: seg 28 had invalid linedef 2929
GL subsector had miniseg as first seg
LoadGLSegs: invalid original vertex 3589

For some reason, these errors are not reproducible even after removing gl.eternal.wad

bkoropoff:
> For some reason, these errors are not reproducible even after removing gl.eternal.wad

There's probably still some issues to iron out in consistently writing gl nodes to a separate wad file correctly, it was a pretty quick hack of the ajbsp source to support that.  Let me know if you figure out how to reproduce it.

andrey-budko:
> Cache misses

Probably something like this can help. I mean copying only needed fields to separate array.
https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/r_things.c#L1536

bkoropoff:
Shaved off some more CPU overhead by moving added fields out of `seg_t` and annotating subsectors with a contiguous list of renderable (non-mini) segs.

prboom2/src/r_defs.h
```diff
@@ -638,4 +662,87 @@ extern fixed_t Sky1ColumnOffset;
...
+  // Component is degenerate (contains a single degenerate subsector)
+  COMPF_DEGENERATE = 0x1,
+  // Component is eligible for flat "bleed-through"
+  COMPF_BLEED_THROUGH = 0x2,
+  // Component is interior (has no segs on single-sided linedefs)
+  COMPF_INTERIOR = 0x4,
+  // Generic mark for recursive algorithms to avoid cycles or repeating work.
+  // Should always be unset when done.
+  COMPF_MARK = 0x8
+} compflags_t;
```
kraflab: The word "comp" is used throughout the code base for (demo) compatibility so this needs a different name. Component is also too generic, so we need a fancy name for this structure 😄

prboom2/src/r_state.h
```diff
@@ -119,4 +119,33 @@ extern int              rw_angle1;
...
+// Components, connected group of subsectors in a sector
+extern component_t* gl_components;
+extern unsigned int gl_numcomponents;
+
+// Bleed lists for components
+extern bleed_t* bleeds;
+extern unsigned int numbleeds;
+
+// Deferred components (used by gl_bsp.c)
+extern component_t** deferred;
```
kraflab: `deferred` is too generic a name for a global like this. I'd prefer to bundle related globals into structs to avoid more pollution. Maybe something like this:
```c
typedef struct {
  gl_component_t* components;
  unsigned int numcomponents;

  gl_bleed_t* bleeds;
  unsigned int numbleeds;

  gl_component_t** deferred;
  unsigned int numdeferred;
} gl_component_data_t;

extern gl_component_data_t gl_component_data;
```
There's probably a better name for this struct, you will have a better conceptual idea of what it represents.

prboom2/src/r_defs.h
bkoropoff: It's technically a [connected component](https://en.wikipedia.org/wiki/Component_\(graph_theory\)) in graph theory, which is where I got the name.  I can think of something else.

prboom2/src/r_state.h
bkoropoff: I could also put `deferred` as a static in `gl_bsp.c` (the only place it's really needed) and expose an init function to have it get allocated/freed when the map changes.

prboom2/src/dsda/bsp.c
```diff
@@ -0,0 +1,1265 @@
...
+    Z_Free(gl_vertexes);
+
+  if (gl_segs && gl_segs != segs)
+    Z_Free(gl_segs);
+
+  if (gl_subsectors && gl_subsectors != subsectors)
+    Z_Free(gl_subsectors);
+
+  if (gl_map_subsectors && gl_map_subsectors != map_subsectors)
+    Z_Free(gl_subsectors);
```
andrey-budko:
```suggestion
  if (gl_map_subsectors && gl_map_subsectors != map_subsectors)
    Z_Free(gl_map_subsectors);
```

andrey-budko:
> There's probably still some issues to iron out in consistently writing gl nodes to a separate wad file correctly, it was a pretty quick hack of the ajbsp source to support that. Let me know if you figure out how to reproduce it.

delete gl.sunder.wad
dsda-doom sunder -warp 9
quit
dsda-doom sunder -warp 1

gl nodes from map09 instead of map01 will be loaded

LoadGLSegs: invalid original vertex 10578

andrey-budko:
stack overflow on sunder.wad map14 with MSVC Debug build

https://i.imgur.com/R47wfOw.png

bkoropoff:
> > There's probably still some issues to iron out in consistently writing gl nodes to a separate wad file correctly, it was a pretty quick hack of the ajbsp source to support that. Let me know if you figure out how to reproduce it.
> 
> delete gl.sunder.wad dsda-doom sunder -warp 9 quit dsda-doom sunder -warp 1
> 
> gl nodes from map09 instead of map01 will be loaded
> 
> LoadGLSegs: invalid original vertex 10578

Thanks, I'll try to figure out what's going on

bkoropoff:
> stack overflow on sunder.wad map14 with MSVC Debug build
> 
> https://i.imgur.com/R47wfOw.png

Yeah, stack bloat with debug builds can be a problem.  I can probably make it use the `paths` array as a stack and avoid recursion.

bkoropoff:
> > There's probably still some issues to iron out in consistently writing gl nodes to a separate wad file correctly, it was a pretty quick hack of the ajbsp source to support that. Let me know if you figure out how to reproduce it.
> 
> delete gl.sunder.wad dsda-doom sunder -warp 9 quit dsda-doom sunder -warp 1
> 
> gl nodes from map09 instead of map01 will be loaded
> 
> LoadGLSegs: invalid original vertex 10578

This is now fixed

bkoropoff:
> stack overflow on sunder.wad map14 with MSVC Debug build
> 
> https://i.imgur.com/R47wfOw.png

This is fixed, though the path finding algorithm still doesn't feel completely baked.

prboom2/src/r_state.h
bkoropoff: It's renamed `gl_deferred` for now, but I'll come up a better way of grouping things once we're confident in the functionality/performance of the code.

andrey-budko:
eternal.wad map28 around sector 721

image

bkoropoff:
> eternal.wad map28 around sector 721

Hmm, this appears to be an ajbsp bug.  Rebuilding nodes with it and running the result in master causes missing wall textures all over the place.

bkoropoff:
Eternal MAP28 seems to have exactly overlapping linedefs in order to have multiple mid textures in the same location, but ajbsp produces segs for only one of the linedefs.  No segs, no wall rendered.

bkoropoff:
Uh, well: https://gitlab.com/andwj/ajbsp/-/blob/master/src/node.cc#L1270

bkoropoff:
It was nice that I just needed to comment out something in ajbsp to fix the missing segs.

ETERNAL MAP28 has a rising ceiling at the start that involves dynamic bleeds: when the map starts, some subsectors need under-bleed from the surrounding ceiling.  The rising ceiling eventually goes higher than the surrounding subsector (so the midtexture with text on it "scrolls" out of the room), at which point it turns into an over-bleed.

I added extremely simplistic logic that just looks for non-zero sector tags and bypasses the height checks during BSP annotation.  It doesn't apply recursively yet.  It would be nice to be more conservative and only allow it if a linedef would actually cause the sector floor/ceiling to change, but I suppose with ACS in the picture that could be equivalent to solving the halting problem...

Anyway, it seems to work, but I'm not sure what side effects it might have.

bkoropoff:
I think I got poly objs working, including when switching render mode.  I found #313 in the process, though.

andrey-budko:
delete gl.eternal.wad
dsda-doom eternal -warp 1
press PgDn (next level) 21 times

`FindOutputLevel(map)` returns -1 for some reasons, and game crashes

No errors with just dsda-doom eternal -warp 19

andrey-budko:
eternal.wad map28 still has some issues around sector 721

https://i.imgur.com/DRjKPV7.png

bkoropoff:
> delete gl.eternal.wad dsda-doom eternal -warp 1 press PgDn (next level) 21 times
> 
> `FindOutputLevel(map)` returns -1 for some reasons, and game crashes
> 
> No errors with just dsda-doom eternal -warp 19

Fixed

bkoropoff:
> eternal.wad map28 still has some issues around sector 721
> 
> https://i.imgur.com/DRjKPV7.png

This is fixed.  The bleed logic is getting a tad subtle, though I guess that's to be expected given the behavior it has to emulate.

prboom2/src/r_defs.h
bkoropoff: Any idea for a new name?  Element?  Constituent?

bkoropoff:
Vote on new name for `component_t`:

- `midsector_t`
- `supersubsector_t`
- `region_t`
- `zone_t`
- `constituent_t`
- `neighborhood_t`
- `connected_graph_component_t`
- `demense_t`

kraflab:
`gl_graph_component_t`? 🤔

bkoropoff:
`gl_chunk_t`?

bkoropoff:
This branch now includes an experimental portal-based renderer and a great deal of cache optimizations.  I get a ~30% FPS increase on FrozenT when overlooking the main area.

bkoropoff:
@andrey-budko any thoughts on the new rendering strategy?

PR #295, opened 2023-03-30
Pragma cleanup

bkoropoff:
This tries to remove a lot of ugly pragmas where possible:

- According to the manual, the GCC interface/implementation pragmas only apply to C++ and have been deprecated since GCC 2.7.2, which was released in *1995*.  Just remove them.
- scanner.cpp went out of its way to memcpy a non-trival class and disable the relevant warning.  Just use the copy assignment operator instead, which generates *identical* assembly on gcc/clang. :man_shrugging:
- Use union type punning to brute-force discard `const`.  This is somewhat less ugly than a bunch of pragmas, and works on all compilers.  Unions are basically the escape hatch for doing questionable casts that would otherwise trigger warnings or undefined behavior.

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -72,11 +72,6 @@
 #include "e6y.h"
 #include "i_system.h"
 
-#ifdef __GNUG__
-#pragma implementation "i_system.h"
-#endif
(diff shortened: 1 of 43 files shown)


kraflab:
Could you split the three ideas into three PRs?

PR #296, opened 2023-03-30
Enum cleanup

bkoropoff:
Fixes various warnings caused by enums being mixed up with each other or plain integers.

### DIFF
--- a/prboom2/src/p_spec.c
+++ b/prboom2/src/p_spec.c
@@ -6287,7 +6287,7 @@ dboolean P_ExecuteZDoomLineSpecial(int special, int * args, line_t * line, int s
         dboolean raise_or_lower;
         byte index;
 
-        static floor_e ceiling_type[2][7] = {
+        static ceiling_e ceiling_type[2][7] = {
           {
             ceilLowerByValue,
             ceilLowerToHighest,
@@ -6412,7 +6412,7 @@ dboolean P_ExecuteZDoomLineSpecial(int special, int * args, line_t * line, int s
       break;
     case zl_generic_stairs:
       {
-        stairs_e type;
+        stair_e type;
 
         type = (args[3] & 1) ? stairBuildUp : stairBuildDown;
         buttonSuccess = EV_BuildZDoomStairs(args[0], type, line,
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -206,16 +206,14 @@ void gld_MultisamplingSet(void)
 
 int gld_LoadGLDefs(const char * defsLump)
 {
-  enum
+  typedef enum
   {
     TAG_SKYBOX,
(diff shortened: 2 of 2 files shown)


prboom2/src/dsda/game_controller.c
```diff
@@ -115,41 +115,47 @@ static void dsda_PollRightStick(void) {
...
+               PollButton(DSDA_CONTROLLER_BUTTON_PADDLE1) |
+               PollButton(DSDA_CONTROLLER_BUTTON_PADDLE2) |
+               PollButton(DSDA_CONTROLLER_BUTTON_PADDLE3) |
+               PollButton(DSDA_CONTROLLER_BUTTON_PADDLE4) |
+               PollButton(DSDA_CONTROLLER_BUTTON_TOUCHPAD);
 
   trigger = dsda_AxisValue(&left_trigger);
   if (trigger)
-    ev.data1.i |= (1 << DSDA_CONTROLLER_BUTTON_TRIGGERLEFT);
+    ev.data1.i |= 1 << DSDA_CONTROLLER_BUTTON_TRIGGERLEFT;
```
kraflab: Are these changes related? If not, please remove them, so the PR remains focused.

prboom2/src/p_spec.c
```diff
@@ -6099,25 +6099,25 @@ dboolean P_ExecuteZDoomLineSpecial(int special, int * args, line_t * line, int s
       buttonSuccess = EV_DoZDoomCeiling(ceilLowerByValue, line, args[0],
                                         P_ArgToSpeed(args[1]), 0,
                                         args[2], P_ArgToCrush(args[4]), 0,
-                                        P_ArgToChange(args[3]), false);
+                                        P_ArgToChange(args[3]), crushDoom);
```
kraflab: `false` and `crushDoom` are very different things, even if they have the same result, so this is going to need a more substantial solution. I'll take a look at this one myself, so you can drop it from this PR.

kraflab:
Could you split this into controller, gl, and map element PRs? I realize that's a bit tedious, but it helps me to digest and merge things in smaller chunks.

prboom2/src/p_spec.c
bkoropoff: `EV_DoZDoomCeiling` explicitly takes a `crushmode_e`.  I dropped these changes from the PR.

bkoropoff:
This PR is very small now, so I'm not sure it needs to be split any more, but let me know.

kraflab:
👍

prboom2/src/p_spec.c
kraflab: I went into the zdoom source and it uses `crushDoom` here as well, so `false` is actually a mistake in the first place. Maybe this predates the enum. If you could open a PR adding the changes back, I would appreciate it 😸

PR #301, opened 2023-03-31
Use union to discard const

bkoropoff:
Cleaner than pragmas (sort of), works on all compilers


kraflab:
This is a funny trick 😄 
Kind of surprising the compiler doesn't complain about this.

PR #302, opened 2023-03-31
game_controller: clean up enums

bkoropoff:
- Use correct enums
- Explicitly cast enums when intentional
- Fix "magic" bitshifts

Fixes warnings, cleaner.


kraflab:
Thanks

PR #304, opened 2023-04-01
Support for non-Latin TEMP directory path on Windows

rfomin:
Taken from Chocolate Doom: https://github.com/chocolate-doom/chocolate-doom/blob/4bfbd986b2831945250930c2c7e80896254cab16/src/m_misc.c#L564

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -254,7 +254,7 @@ const char *I_DoomExeDir(void)
 
 const char* I_GetTempDir(void)
 {
-  static const char* tmp_path = NULL;
+  static const char* tmp_path;
 
   if (!tmp_path)
   {
@@ -264,14 +264,9 @@ const char* I_GetTempDir(void)
     result = GetTempPathW(PATH_MAX, wpath);
 
     if (result == 0 || result > MAX_PATH)
-    {
-      lprintf(LO_ERROR, "GetTempPath(2) failed\n");
-      tmp_path = strdup(".");
-    }
+      I_Error("I_GetTempDir: GetTempPathW failed");
     else
-    {
       tmp_path = ConvertWideToUtf8(wpath);
-    }
   }
 
   return tmp_path;
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -261,11 +261,8 @@ const char* I_GetTempDir(void)
     wchar_t wpath[PATH_MAX];
     DWORD result;
 
-#ifdef HAVE_GET_TEMP_PATH_2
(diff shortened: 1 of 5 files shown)


kraflab:
What's the problem exactly? The chocolate doom implementation has both reduced coverage and the shocking possibility of returning `.`

kraflab:
What about `GetTempPathW`?

prboom2/src/SDL/i_system.c
```diff
@@ -254,11 +254,27 @@ const char *I_DoomExeDir(void)
...
+
+#ifdef HAVE_GET_TEMP_PATH_2
+    result = GetTempPath2W(PATH_MAX, wpath);
+#else
+    result = GetTempPathW(PATH_MAX, wpath);
+#endif
+    if (result == 0 || result > MAX_PATH)
+    {
+      lprintf(LO_ERROR, "GetTempPath(2) failed\n");
+      tmp_path = strdup(".");
```
FtZPetruska: If the current directory is used as a temporary directory, unzipping a file in the same directory may fail, e.g.:

- `dsda-doom -iwad doom2 -file ./foo.zip`
- `GetTempPathW` fails and `I_GetTempDir` returns `.`
- `D_AddZip` uses `./foo.zip/` as the output directory:
```c
dsda_InitString(&temporary_directory, I_GetTempDir());
dsda_StringCatF(&temporary_directory, "/%s/", dsda_BaseName(zipped_file_name));
M_MakeDir(temporary_directory.string, true);
```

Maybe strip the `.zip` part of the name with `dsda_CutExtension`.

kraflab: We can't use `.` ever, this would be completely insane if files just get dumped into a directory without being requested by a user.

kraflab: Could you leave this branch as an `I_Error` for this PR? I will set up something in the dsda_doom_data directory as a fallback, but raising an error for now is fine with me.

prboom2/src/SDL/i_system.c
```diff
@@ -254,11 +254,27 @@ const char *I_DoomExeDir(void)
 
 const char* I_GetTempDir(void)
 {
-  static char tmp_path[PATH_MAX] = {0};
+  static const char* tmp_path = NULL;
```
kraflab: `static` values are always zero-initialized, so you can leave off the assignment.

prboom2/src/SDL/i_system.c
rfomin: Done in https://github.com/kraflab/dsda-doom/pull/304/commits/f4c81bc197c81bc7072445122ef5ce8dd27d7780

kraflab:
Awesome, thanks 👍

PR #305, opened 2023-04-01
Fix compilation with clang-cl

rfomin:
`clang-cl` allows using MS dev tools with the clang compiler. It defines `_MSC_VER` but still relies on `__attribute__` extensions in its headers. Perhaps we should move these macros into one header and remake them like in Choco: https://github.com/chocolate-doom/chocolate-doom/blob/4bfbd986b2831945250930c2c7e80896254cab16/src/doomtype.h#L68

### DIFF
--- a/prboom2/src/doomdef.h
+++ b/prboom2/src/doomdef.h
@@ -41,7 +41,7 @@
 #endif
 
 // killough 4/25/98: Make gcc extensions mean nothing on other compilers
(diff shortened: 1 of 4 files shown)


kraflab:
Thanks. About moving the macros, I don't necessarily feel like that's necessary at this point, but I'll keep it in mind.

PR #306, opened 2023-04-01
Remove redundant miniseg field

bkoropoff:
Measurements on my experimental branch showed that performance on large maps is sensitive to `seg_t` cache performance, so I'm trying to optimize its size a bit.  Removing this field is low-hanging fruit since it's completely redundant (minisegs have a `NULL` linedef, so check that instead).

### DIFF
--- a/prboom2/src/r_bsp.c
+++ b/prboom2/src/r_bsp.c
@@ -434,8 +434,7 @@ static void R_AddLine (seg_t *line)
       maxdrawsegs = newmax;
     }
 
-    if(curline->miniseg == false) // figgi -- skip minisegs
-      curline->linedef->flags |= ML_MAPPED;
+    curline->linedef->flags |= ML_MAPPED;
 
     // proff 11/99: the rest of the calculations is not needed for OpenGL
     ds_p++->curline = curline;
@@ -849,7 +848,7 @@ static void R_Subsector(int num)
   line = &segs[sub->firstline];
   while (count--)
   {
-    if (line->miniseg == false)
+    if (line->linedef)
       R_AddLine (line);
     line++;
     curline = NULL; /* cph 2001/11/18 - must clear curline now we're done with it, so R_ColourMap doesn't try using it for other things */
--- a/prboom2/src/p_setup.c
+++ b/prboom2/src/p_setup.c
@@ -557,8 +557,6 @@ static void P_LoadSegs (int lump)
       //li->v1 = &vertexes[v1];
       //li->v2 = &vertexes[v2];
 
-      li->miniseg = false; // figgi -- there are no minisegs in classic BSP nodes
-
       // e6y: moved down, see below
       //li->length  = GetDistance(li->v2->x - li->v1->x, li->v2->y - li->v1->y);
 
@@ -691,8 +689,6 @@ static void P_LoadSegs_V4(int lump)
     v1 = LittleLong(ml->v1);
     v2 = LittleLong(ml->v2);
 
-    li->miniseg = false; // figgi -- there are no minisegs in classic BSP nodes
-
     li->angle = (LittleShort(ml->angle))<<16;
     li->offset =(LittleShort(ml->offset))<<16;
     linedef = (unsigned short)LittleShort(ml->linedef);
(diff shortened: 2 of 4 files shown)


prboom2/src/r_bsp.c
```diff
@@ -434,8 +434,7 @@ static void R_AddLine (seg_t *line)
       maxdrawsegs = newmax;
     }
 
-    if(curline->miniseg == false) // figgi -- skip minisegs
-      curline->linedef->flags |= ML_MAPPED;
+    curline->linedef->flags |= ML_MAPPED;
```
kraflab, line 437: What happened to the `if`?

andrey-budko:
I tried to split `seg_t` into common and rare/software fields on top of this and got 2-3%% speed up in comparison with master. 224 fps vs 216 on Sunder map10, and 98 vs 95 on FrozenT.

`seg->angle` - not used at all in doom, can be removed in prboom-plus-um
`seg->pangle`, `seg->length` and `seg->offset` are used only once, and only for Software renderer

https://github.com/andrey-budko/dsda-doom/commit/54136a762158c3e861a7cbd4799751c8e752be92

prboom2/src/r_bsp.c
bkoropoff: It was redundant with the check in `R_Subsector` before calling `R_AddLine`

bkoropoff:
This and #309 together get `seg_t` to just fit in a cache line.  On my experimental branch it helped offset some of the performance loss (I think from GL nodes in general using more segs, because my bleed effect code wasn't showing up in the profile).  Doing what @andrey-budko did on top of it would probably help even more.

andrey-budko:
Another attempt.

https://github.com/andrey-budko/dsda-doom/commit/0481eec79d9fcc741a9f8b480aa56fa314250a93

Slightly slower, but cleaner.

bkoropoff:
@andrey-budko It's possible to go even smaller by replacing pointer fields with array offsets of an appropriate size (e.g. `uint16_t` for sector, `uint32_t` for vertex) and sorting the resulting struct fields by descending size to eliminate padding.  I experimented with doing this for `frontsector`/`backsector` but did not submit it because it makes the code uglier.

kraflab:
> @andrey-budko It's possible to go even smaller by replacing pointer fields with array offsets of an appropriate size (e.g. `uint16_t` for sector, `uint32_t` for vertex) and sorting the resulting struct fields by descending size to eliminate padding. I experimented with doing this for `frontsector`/`backsector` but did not submit it because it makes the code uglier.

Eventually I will for instance be uncapping the sector count limit (a consequence of udmf), so it's probably not worth pursuing that direction.

bkoropoff:
> > @andrey-budko It's possible to go even smaller by replacing pointer fields with array offsets of an appropriate size (e.g. `uint16_t` for sector, `uint32_t` for vertex) and sorting the resulting struct fields by descending size to eliminate padding. I experimented with doing this for `frontsector`/`backsector` but did not submit it because it makes the code uglier.
> 
> Eventually I will for instance be uncapping the sector count limit (a consequence of udmf), so it's probably not worth pursuing that direction.

`uint32_t` still saves 4 bytes over a pointer on x86_64, unless UDMF has maps with *a lot* of sectors. :)

I mostly avoided it because it clutters up code that uses the structure, but we can explore it if we need to eek out more performance somewhere.

andrey-budko:
I reduced seg_t to 16 bytes with this technique, but without noticeable fps improvement in comparison with segs_data separate array, which by itself doesn't add much

```c++
typedef struct
{
  unsigned int v1i, v2i;
  unsigned short side;
  unsigned short line;
  unsigned short front, back;
} seg_t;
```

```c++
inline sector_t* SEG_BACK(seg_t* seg) { return sectors + seg->back; };
inline sector_t* SEG_BACK_NULL(seg_t* seg) { return seg->back == NO_SECTOR ? 0 : sectors + seg->back; };
inline void SEG_SET_BACK(seg_t* seg, sector_t* sector) { seg->back = sector - sectors; };
inline void SEG_SET_NOBACK(seg_t* seg) { seg->back = NO_SECTOR; };
inline dboolean SEG_HAS_BACK(seg_t* seg) { return seg->back != NO_SECTOR; };
// etc
```

bkoropoff:
@andrey-budko Subsectors tend to have only one or two segs and aren't necessarily contiguous in memory even if they are adjacent in rendering order, so spatial locality isn't great.  This means returns diminish under 64 bytes (the size of a cache line).  Other changes to the renderer (#311) will be necessary to benefit from extremely compact segs.

PR #307, opened 2023-04-01
Docs: Update dependencies lists.

FtZPetruska:
Updates the build guides to account for changes in the dependencies:
- Add `libzip` to all guides.
- On the macOS guide specifically:
  - Remove `pcre` since it's no longer needed.
  - Add `pkgconf` to help searching dependencies without CMake configs.


kraflab:
Thanks 👍

PR #309, opened 2023-04-02
seg_t: shrink length field

bkoropoff:
Another easy `seg_t` optimization.  The `length` field can be reduce to a `uint32_t` based on the following observation:

- The maximum hypotenuse of a seg is `sqrt(2) * (INT32_MAX - INT32_MIN)` i.e. `sqrt(2) * UINT32_MAX`
- The hypotenuse is always shifted to the right by 1 before using it, so the result can be pre-divided by `2.0`
- `2.0` is greater than `sqrt(2)`, so the result always fits in `uint32_t`.

This also reorders the fields to ensure there is no struct padding on any platform.

4 bytes (and a bit shift :upside_down_face:) saved for free.

### DIFF
--- a/prboom2/src/p_setup.c
+++ b/prboom2/src/p_setup.c
@@ -3307,7 +3307,7 @@ static void R_CalcSegsLength(void)
     int64_t dx = (int64_t)li->v2->px - li->v1->px;
     int64_t dy = (int64_t)li->v2->py - li->v1->py;
     length = sqrt((double)dx*dx + (double)dy*dy);
-    li->length = (int64_t)length;
+    li->halflength = (uint32_t)(length / 2.0);
     // [crispy] re-calculate angle used for rendering
     li->pangle = R_PointToAngleEx2(li->v1->px, li->v1->py, li->v2->px, li->v2->py);
   }
--- a/prboom2/src/r_defs.h
+++ b/prboom2/src/r_defs.h
@@ -390,19 +390,17 @@ typedef struct msecnode_s
 typedef struct
 {
   vertex_t *v1, *v2;
-  fixed_t offset;
-  angle_t angle;
-  angle_t pangle; // re-calculated angle used for rendering
-  int64_t length; // fix long wall wobble
   side_t* sidedef;
   line_t* linedef;
-
   // Sector references.
   // Could be retrieved from linedef, too
   // (but that would be slower -- killough)
   // backsector is NULL for one sided lines
-
   sector_t *frontsector, *backsector;
+  fixed_t offset;
+  angle_t angle;
+  angle_t pangle; // re-calculated angle used for rendering
+  uint32_t halflength; // fix long wall wobble
 } seg_t;
 
 typedef struct ssline_s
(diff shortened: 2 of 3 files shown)


prboom2/src/r_defs.h
```diff
@@ -390,23 +390,20 @@ typedef struct msecnode_s
 typedef struct
 {
   vertex_t *v1, *v2;
-  fixed_t offset;
```
kraflab, line 393: There is at least one struct in the doom code base that has members that need to be in specific positions (as insane as that sounds). I don't think this is one, but I wanted to bring that up as a caution for you if you hadn't considered it.

bkoropoff: Do you have any idea which one and for what reason?

kraflab:
```c
typedef struct visplane
{
  struct visplane *next;        // Next visplane in hash chain -- killough
  int picnum, lightlevel, minx, maxx;
  int special; // heretic
  fixed_t height;
  fixed_t xoffs, yoffs;         // killough 2/28/98: Support scrolling flats
  angle_t rotation;
  fixed_t xscale;
  fixed_t yscale;
  // e6y: resolution limitation is removed
  // bottom and top arrays are dynamically
  // allocated immediately after the visplane
  unsigned short *bottom;
  unsigned short pad1;          // leave pads for [minx-1]/[maxx+1]
  unsigned short top[3];

  // NEW FIELDS MUST BE ADDED **ABOVE** `unsigned short *bottom;`

} visplane_t;
```
The comment at the bottom is from me, after I ran into rendering errors while adding udmf transforms to this struct. The visplane has a dynamic size based on resolution, which shows up here:
```c
  visplane_t *check = freetail;
  if (!check)
  {
    // e6y: resolution limitation is removed
    check = Z_Calloc(1, sizeof(*check) + sizeof(*check->top) * (SCREENWIDTH * 2));
```

bkoropoff: Ah, yeah.  This would typically be done with "flexible array members" in C99 or later, e.g.

```c
struct foo
{
    ...
    unsigned short array[];
}
```

If you try to put something after `array` the compiler will complain.

bkoropoff:
Rebased against master

prboom2/src/p_setup.c
```diff
@@ -3307,7 +3307,7 @@ static void R_CalcSegsLength(void)
     int64_t dx = (int64_t)li->v2->px - li->v1->px;
     int64_t dy = (int64_t)li->v2->py - li->v1->py;
     length = sqrt((double)dx*dx + (double)dy*dy);
-    li->length = (int64_t)length;
+    li->length = (uint32_t)(length / 2.0);
```
kraflab: Should we rather convert to 64 bits, shift right, and convert to 32 bits, so it's exactly the same as before?

bkoropoff: I can't think of a way the results could be different.  Floating point division by 2.0 should be exact for non-subnormal values (subtracting 1 from the exponent), and subnormal inputs should be impossible (the smallest non-zero length will be 1).  Truncating twice instead of once should also not change the result.

bkoropoff:
Any further problems with this?  Note that even if it changed results slightly (which I don't think it does), it would only slightly perturb wall rendering by a pixel, not game behavior.

kraflab:
I'm knee deep in the sky rendering code right now but I will get back to PRs soon :^)

Could you change the name of the field to "halflength" (and then you don't need the comment about it being pre-shifted I guess). Otherwise some future developer will be in for a surprise when they try to use the "length" for something 😄

bkoropoff:
Done

bkoropoff:
I didn't break macOS, it's failing to install dependencies.

kraflab:
Thanks 👍

PR #315, opened 2023-04-03
Fix MSVC x86 build compilation error

andrey-budko:
> cannot convert from 'void (__cdecl *)()' to 'void (__stdcall *)()'


kraflab:
Thanks 👍

PR #316, opened 2023-04-03
Make installs relocatable.

FtZPetruska:
Currently, the search directory for `dsda-doom.wad` is a hardcoded absolute path.

This works great when the install location is guaranteed to be the same everywhere, through a package manager for example. But this stops you from "making a single package and run it from anywhere" due to the initial prefix being hardcoded.

This PR addresses the issue as such:
- Set the `DSDAPWADDIR` in `config.h` to the relative path from the bin directory to the location of `DSDAPWADDIR`. With the current defaults, it would be `""` on Windows and `"../share/games/doom` on Unix.
- In `search0`, change the `DSDAPWADDIR` entry to start from the executable's directory, and use the relative path as a sub-dir. The executable directory is acquired with `SDL_GetBasePath`.

Here's the testing I've done that illustrates the change:
- Configure to install in `/opt/foo`, and with `DSDAPWADDIR` set to `data` (so it differs from the `DOOMWADDIR`)
- Move `/opt/foo` to `/opt/bar`
- Run `/opt/bar/bin/dsda-doom`

Previously:
```
Unable to find required file "dsda-doom.wad"
```
Now:
```
Playing: DOOM 2: Hell on Earth
 adding /usr/local/share/games/doom/doom2.wad
 adding /opt/bar/bin/../data/dsda-doom.wad
```

It also works on Windows if the pwad is installed in a different directory than the executable!

---

A few notes about `SDL_GetBasePath`:
- The documentation tags it as an "expensive call", so I wrapped it in a helper and store its result in a static variable.
- It is heap-allocated, but I'm having a hard time finding a good way to free it without using `I_AtExit`. Maybe it can be achieved through making `Init/FreeBasePath` functions and have them called in `D_DoomMainSetup` like:
  ```c
  static void D_DoomMainSetup(void)
  {
    InitBasePath();
    // Functions that use I_FindFileInternal
    FreeBasePath();
  }
  ```
Then again, it does not seem like the result of `I_DoomExeDir` is ever freed either.


prboom2/src/SDL/i_system.c
```diff
@@ -333,6 +333,15 @@ dboolean HasTrailingSlash(const char* dn)
           );
 }
 
+static const char *GetExecutableDir(void)
```
fabiangreffrath: What's wrong with `I_DoomExeDir()`?

prboom2/src/SDL/i_system.c
```diff
@@ -363,7 +372,7 @@ char* I_FindFileInternal(const char* wfname, const char* ext, dboolean isStatic)
     {NULL}, // current working directory
     {NULL, NULL, "DOOMWADDIR"}, // run-time $DOOMWADDIR
     {DOOMWADDIR}, // build-time configured DOOMWADDIR
-    {DSDAPWADDIR}, // build-time configured location of dsda-doom.wad
+    {NULL, DSDAPWADDIR, NULL, GetExecutableDir}, // build-time configured location of dsda-doom.wad
```
fabiangreffrath: I wouldn't have anything against *adding* a relative data directory, but please do not *remove* the ability to add an absolute one. I need this for the Debian package.

prboom2/src/SDL/i_system.c
FtZPetruska: Unlike what the name seems to imply, the `I_DoomExeDir()` function returns the path to `~/.dsda-doom` (on Unix at least), not the actual directory the executable is located in.

prboom2/src/SDL/i_system.c
FtZPetruska: Could you elaborate on your needs for the Debian package please ?

Would having a separate absolute version of that path be a solution ? For example:

In CMake:
```cmake
set(DSDAPWADDIR_ABSOLUTE "${CMAKE_INSTALL_PREFIX}/${DSDAPWADDIR}")
```

In config.h.cin:
```c
#cmakedefine DSDAPWADDIR "@DSDAPWADDIR_ABSOLUTE@"
#cmakedefine BIN_TO_PWAD_PATH "@DSDA_BIN_TO_PWAD_PATH@"
```

In SDL/i_system.c:
```c
{DSDAPWADDIR}, // absolute path to dsda-doom.wad's directory
{NULL, BIN_TO_PWAD_PATH, NULL, GetExecutableDir}, // relative path from the executable to dsda-doom.wad's directory
```

One thing that cannot be changed is having the CMake `DSDAPWADDIR` be a path relative to the install prefix since that is necessary for CPack to work properly.

fabiangreffrath: > Would having a separate absolute version of that path be a solution ?

Yes, as it was before.

FtZPetruska: Alright, I've made the changes in [1a62565](https://github.com/kraflab/dsda-doom/pull/316/commits/1a62565ea90161c73af269b92dd6d9de73c132ba). Here's how it works now:
- `DSDAPWADDIR` can be set to a path relative to the prefix, or an absolute path.
- If `DSDAPWADDIR` is relative, get its absolute path.
- Get the relative path from where the executable is installed to where `dsda-doom.wad` is installed
- Save both the absolute and relative paths to config.h as, respectively, `DSDA_ABSOLUTE_PWAD_PATH` and `DSDA_BIN_TO_PWAD_PATH`.
- Search both paths in `I_FindFileInternal`.

prboom2/src/SDL/i_system.c
kraflab: We shouldn't have 2 functions with essentially the same name that do different things :^)
Let's call this `I_GetBasePath` to correspond to the SDL semantics. Probably `I_DoomExeDir` needs a better name but that's a separate topic.

kraflab:
I'm not sure I completely understand the problem statement. If you install something as opposed to just building it, don't you know where it is? Why would you be manually editing the installed files? Is the problem rather that on Linux it doesn't search the actual executable directory when looking for the pwad? If you want a portable installation, I would assume you would want to just have a directory with the executable and the wad file, like on windows.

FtZPetruska:
When following the GNU install directories, the layout isn't flat like on Windows, the convention is to install binaries in `bin`, and resources to `share`, so with the current defaults it gives something like:
```
/usr/local/ (prefix)
|- bin/
|  |- dsda-doom
|- share/
|  |- games/
|  |  |- doom/
|  |  |  |- dsda-doom.wad
```

For package managers such as Homebrew, following this directory structure is mandatory. Moreover, it is not acceptable to install `dsda-doom.wad` to the bin directory.

Searching only the absolute path written at build time works fine when the prefix is always the same (such as with package managers). But if you attempt to make a binary release and unzip it outside of where it was initially built to be installed to (say `/opt/local`, `/opt/dsda-doom`, or `~/.local`), then searching with an absolute path fails due to the prefix changing. Using a relative path from the binary to the dsda pwad is unaffected by the prefix change, because the structure under the prefix is preserved.

Hopefully this describes the issue better 😅

fabiangreffrath:
> But if you attempt to make a binary release

I still fail to see an actual use case for this outside the Windows thought model.

kraflab:
> Hopefully this describes the issue better 😅

Unfortunately I have the same question :^)

You are either installing it like a package, in which case why would you care where it is or how it's laid out, or you are not installing it like a package, in which case why do you need an arbitrary location for the wad file? Why not store the wad file with the executable if you want a portable directory that doesn't get installed?

Also, won't relative paths completely not work for the case suggested anyway? You can't just move something to any directory structure and have the relative path work.

kraflab:
To put this a different way: any portable zip of a program i've seen has all its dependencies within the folder, because that's the only thing that guarantees portability. So my question would be why we aren't just using the simplest solution for that case which would be the wad stored with the exe?

FtZPetruska:
> To put this a different way: any portable zip of a program i've seen has all its dependencies within the folder, because that's the only thing that guarantees portability. So my question would be why we aren't just using the simplest solution for that case which would be the wad stored with the exe?

That's fair. I don't think the PR should be closed though, as you asked earlier, no dsda-doom will not search for pwad in the executable's directory on Linux (e.g. running `./build/dsda-doom` is unable to find `dsda-doom.wad`) so it probably should be added.

kraflab:
What do you think about having a portable build flag that changes I_DoomExeDir to actually be the exe dir?

FtZPetruska:
> What do you think about having a portable build flag that changes I_DoomExeDir to actually be the exe dir?

I would prefer to leave the `I_DoomExeDir` as-is and additionally search SDL's base path instead of introducing a new build option to alter behaviour.

With the latest commits, the changes are as follow:
- Allow users to set `DSDAPWADDIR` to an absolute path in CMake.
- Search for WADs in `SDL_GetBasePath`.

kraflab:
Looks great, thanks 👍

PR #320, opened 2023-04-04
Update docs for LTO, and turn it off by default.

FtZPetruska:
- Turn LTO off by default.
- Update the documentation to include LTO:
  - Enable LTO for macOS and MSVC
  - Add a notice about the issues with LTO when using MinGW GCC.


kraflab:
Thanks 👍

PR #330, opened 2023-04-07
Fix skydome seam

### LINKED ISSUE #329: Thin seam on skydome
After replacing float with double here https://github.com/kraflab/dsda-doom/commit/e38c040e9480ffd36050da05f3a35e2d991cae2a

image

andrey-budko:
Closes #329

### DIFF
--- a/prboom2/src/gl_sky.c
+++ b/prboom2/src/gl_sky.c
@@ -44,6 +44,7 @@
 #include "gl_opengl.h"
 
 #include <SDL.h>
+#include <math.h>
 
 #include "doomstat.h"
 #include "v_video.h"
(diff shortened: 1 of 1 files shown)


kraflab:
Thanks!

andrey-budko:
Now I'm not sure about the speed of all these sin/cos.

Is this enough?
```diff
- angle_t topAngle = (angle_t)(c / (double)columns * ANGLE_MAX);
+ angle_t topAngle = (angle_t)((c == columns ? 0 : c) / (double)columns * ANGLE_MAX);
```

kraflab:
You mean to revert everything and just add that diff?

andrey-budko:
`finesine` for the first point (0 rad) and for the last (2pi rad) do not match with 'double' version - and probably that's the problem.

https://godbolt.org/z/qWbeYn6Ye

GZDoom uses precise sin/cos version though.
https://github.com/ZDoom/gzdoom/blob/master/src/common/rendering/hwrenderer/data/hw_skydome.cpp#L148

Do we care about speed in that place?

PR #331, opened 2023-04-07
Fix macOS CI.

### LINKED ISSUE #328: macOS CI build is broken
There is some issue with dependencies.

FtZPetruska:
Fixes #328 

---

Seems like running `brew upgrade` in CI does not work because of `go`.

Since `brew install` automatically upgrades the dependencies of the installed packages, we can get away with not running `brew upgrade`.


kraflab:
Thanks 👍

PR #332, opened 2023-04-07
Clean the temporary directory before unzipping.

FtZPetruska:
As mentioned in #133, the temporary directory should be cleaned to avoid loading unexpected files.

Since `M_remove` only works on empty directories, I added a new function to the `m_file` API ~~`M_RemoveDirRecursive`~~ `M_ClearDirectory`:
- ~~On Windows, it uses the `SHFileOperation` which requires a double null-terminated string.~~
- ~~On other platforms, use `nftw` with the `FTW_DEPTH` flag. The latter ensures `nftwfunc` is first called with the content of the (sub)directory before the the directory itself.~~
- Using `i_glob`, it removes every file in the temporary directory.

Other minor changes:
- Make `M_remove` return a non-zero value when `wpath` is NULL. The standard's `remove` and Win32's `_wremove` both return 0 on success.
- Instead of adding a trailing `/` in `D_AddZip`, I insert it when constructing the output filename. Functionally, it changes nothing, just makes logging look nicer:
  - Before:
  ```
  adding /tmp/d2isov2.zip//D2ISOv2.wad
  ```
  - After:
  ```
  adding /tmp/d2isov2.zip/D2ISOv2.wad
  ```

### DIFF
--- a/prboom2/src/m_file.c
+++ b/prboom2/src/m_file.c
@@ -36,7 +36,6 @@
 #include <windows.h>
 #include <io.h>
 #include <direct.h>
-#include <shellapi.h>
 #endif
 
 #ifdef HAVE_CONFIG_H
@@ -45,10 +44,6 @@
 
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
-#endif
-
-#ifdef HAVE_FTW_H
-#include <ftw.h>
 #endif
 
 #include <stdlib.h>
@@ -548,48 +543,5 @@ char *M_getenv(const char *name) {
   return env;
 #else
   return getenv(name);
-#endif
-}
-
-static int nftwfunc(const char *filename, const struct stat *stats, int flags, struct FTW *pfwt)
-{
-  (void)stats;
-  (void)flags;
-  (void)pfwt;
-  return M_remove(filename);
-}
-
-#define NFTW_LIMIT 64
-
-dboolean M_RemoveDirRecursive(const char *path)
-{
-#ifdef _WIN32
-  wchar_t *wpath;
-  int result;
-
-  SHFILEOPSTRUCTW file_op = {
-    NULL,
-    FO_DELETE,
-    NULL,
-    NULL,
-    FOF_NOCONFIRMATION | FOF_NOERRORUI | FOF_SILENT,
-    false,
-    NULL,
-    NULL};
-
-  wpath = ConvertUtf8ToWide(path);
-  if (!wpath)
-    return false;
-
-  file_op.pFrom = wpath;
-  result = SHFileOperationW(&file_op);
-
-  Z_Free(wpath);
-  return result == 0;
-#elif defined(HAVE_FTW_H)
-  return nftw(path, nftwfunc, NFTW_LIMIT, FTW_DEPTH | FTW_PHYS) == 0;
-#else /* !defined(_WIN32) && !defined(HAVE_FTW_H) */
-  lprintf(LO_WARN, "M_RemoveDirRecursive: Not supported on your platform.\n");
-  return false;
 #endif
 }
--- a/prboom2/src/m_file.c
+++ b/prboom2/src/m_file.c
@@ -283,7 +283,7 @@ int M_remove(const char *path)
   wpath = ConvertUtf8ToWide(path);
 
   if (!wpath)
-    return 0;
+    return -1;
 
   ret = _wremove(wpath);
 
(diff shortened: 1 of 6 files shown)


prboom2/src/m_file.c
```diff
@@ -545,3 +550,46 @@ char *M_getenv(const char *name) {
...
+{
+  (void)stats;
+  (void)flags;
+  (void)pfwt;
+  return M_remove(filename);
+}
+
+#define NFTW_LIMIT 64
+
+dboolean M_RemoveDirRecursive(const char *path)
```
bkoropoff: This is a pretty dangerous function, so it might be good to add a check that `path` begins with a directory that DSDA has business modifying, does not contain anything that might escape (`..` components) and `I_Error` out otherwise.

prboom2/src/m_file.c
```diff
@@ -545,3 +550,46 @@ char *M_getenv(const char *name) {
...
+#ifdef _WIN32
+  wchar_t *wpath;
+  int result;
+
+  SHFILEOPSTRUCTW file_op = {
+    NULL,
+    FO_DELETE,
+    NULL,
+    NULL,
+    FOF_NOCONFIRMATION | FOF_NOERRORUI | FOF_SILENT,
```
bkoropoff: You might want to add `FOF_NORECURSEREPARSE`.  It's deprecated on modern Windows (presumably it does not follow symlinks/junctions/shortcuts/etc. by default when deleting), but I'm not sure which version deprecated it.

rfomin:
I suggest using `i_glob.c` to find files recursively. We use it in Woof: https://github.com/fabiangreffrath/woof/blob/11c652b67e6144ac66f7b45e9942b4d963fa0542/src/d_quit.c#L52

FtZPetruska:
> I suggest using `i_glob.c` to find files recursively. We use it in Woof: https://github.com/fabiangreffrath/woof/blob/11c652b67e6144ac66f7b45e9942b4d963fa0542/src/d_quit.c#L52

Since the goal is to only cleanup files so they don't get wrongly loaded, I suppose `i_glob` is fine. It's unfortunate that it does not seem to return directories, but it's probably better than the current solution.

rfomin:
> It's unfortunate that it does not seem to return directories, but it's probably better than the current solution.

We can save directory names when unpacking and remove them with `rmdir()`. In Woof, we ignore the directory structure in zip archives and filter files when unpacking:  https://github.com/fabiangreffrath/woof/blob/c642d279937d27133661ac4f641aac5273460e46/src/d_main.c#L602

prboom2/src/m_file.c
```diff
@@ -545,3 +546,26 @@ char *M_getenv(const char *name) {
   return getenv(name);
 #endif
 }
+
+dboolean M_ClearDirectory(const char *path)
```
kraflab: This is non-recursive right? Maybe there is a more specific name. `M_RemoveFilesAtPath`?

kraflab:
Any other concerns here?

kraflab:
There are some zips that are packaged with a folder at the root. For instance crumpets.zip unpacks to crumpets.zip/crumpets/crumpets.wad and loading fails. I don't know if we should autoload recursively because sometimes there are optional files in subdirectories. But in a case where there is a folder of the same name as the zip, maybe it would make sense to move all the files inside that folder up. What do you think? This should probably be in a followup PR, but wanted to mention it here where there is already related discussion 😄 

Maybe that's too much effort though and we should change autoload to be recursive. There's no way to know if some files in a zip are optional or required, so such cases probably need to be managed by the user regardless.

FtZPetruska:
> There are some zips that are packaged with a folder at the root. For instance crumpets.zip unpacks to crumpets/crumpets.wad. I don't know if we should autoload recursively because sometimes there are optional files in subdirectories. But in a case where there is a folder of the same name as the zip, maybe it would make sense to move all the files inside that folder up. What do you think? This should probably be in a followup PR, but wanted to mention it here where there is already related discussion 😄

I've sort of addressed that in [58dea8e](https://github.com/kraflab/dsda-doom/pull/332/commits/58dea8ef6c06741bad6101a4c44ca60ea57c4d88). This change made so the folder structure inside the zip will be ignored and all the files are extracted in the temp directory. Similar to how Woof does it. This also helps eliminating uncertainty with the directory structure when cleaning it up since it guarantees no subdirectory.

With a build from this PR, I tried loading crumpets and the wad was extracted to `<tempdir>/crumpets.zip/crumpets.wad` and properly loaded.

If a WAD archive has this structure:
```
|- foo.wad
|- optional/
|  |- bar.wad
```
then yes, all the optional content would be loaded

kraflab:
Oh awesome 👍

PR #333, opened 2023-04-07
Check for C23 builtin bool/true/false.

FtZPetruska:
When compilers implements C23's [N2935](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2935.pdf), the `dboolean` enum in `doomtype.h` breaks compilation with the following error:
```
dsda-doom/prboom2/src/doomtype.h:48:15: error: expected identifier
typedef enum {false, true} dboolean;
              ^
dsda-doom/prboom2/src/doomtype.h:48:22: error: expected identifier
typedef enum {false, true} dboolean;
                     ^
```

This PR adds a simple check in CMake to see if the paper is implemented, and if it is, make `dboolean` an alias of `bool` similar to C++.


kraflab:
What's the size of the `bool` field in c23? The `dboolean`, despite its name, is 4 bytes.

FtZPetruska:
On Windows and macOS, using LLVM Clang 16, `sizeof(bool)` in C23 mode returns 1. Seems to match C++'s `bool` type where any non-zero value is narrowed to 1.

> The `dboolean`, despite its name, is 4 bytes.

Some platforms use 2 bytes by default for enums, e.g. SDL has this piece of code in `SDL_stdinc.h`:
```c
#if !defined(__ANDROID__) && !defined(__VITA__) && !defined(__3DS__)
   /* TODO: include/SDL_stdinc.h:174: error: size of array 'SDL_dummy_enum' is negative */
typedef enum
{
    DUMMY_ENUM_VALUE
} SDL_DUMMY_ENUM;

SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
#endif
```

bkoropoff:
Recent MSVC (like 2013 and on) has `stdbool.h`, so it might be best to just include that and make `dboolean` a type alias.

Does anything actually depend on `dboolean` being 4 bytes (e.g. serialization code)?  Otherwise it's better to use the standard type.

kraflab:
I'm pretty cautious about making a change to the size of something that's used in a million places, especially with the amount of hacks in the source that depend on sizes of things. Did you know hexen had a hack where the byte arg array for line specials was cast as a single int in a couple cases? That was a nice bug to track down when converting the array to ints for udmf 😄

Another fun one was where overflows were intentionally used to cause periodic effects.

bkoropoff:
Well, you could change `dboolean` to `int` and include `stdbool.h` to ensure you have `true`/`false`.  In C23 these should implicitly convert to int, so everything continues to work.

FtZPetruska:
Another option to keep an enum while not clashing with keywords is to use `#define` as such:

```c
typedef enum {dfalse, dtrue} dboolean;
#define false dfalse
#define true dtrue
```

With this, there should be no need to have a workaround for C++ either.

SDL had a similar issue where third party code was defining a similar bool enum (`typedef enum {false, true} bool;`) which broke with C23. In the end, they favoured using [`#define`s](https://github.com/libsdl-org/SDL/blob/05b701f12e55ae324425a5ed8e94d7e4184d7dcc/src/joystick/hidapi/SDL_hidapi_steam.c#L34).

What do you think ?

kraflab:
The define stuff seems like the least invasive change

bkoropoff:
The N2655 draft states "we think that it is important to impose diagnostics whenever user code tries to undefine or redefine these new keywords".  It's not a good idea to try to `#define` reserved language keywords.

kraflab:
But they wouldn't be defined in the context where they are reserved, right?

I mean we only need to define true / false in the case where they don't already exist. Otherwise it doesn't matter what names are in the enum, I assume.

kraflab:
Or is that then going to cause a problem for "is it in the enum" warnings? 😄

bkoropoff:
They are reserved in the C preprocessor as well, and the spec explicitly encourages compilers to complain when trying to `#define` them.  I think making `dboolean` an `int` and just including `stdbool.h` to get `true`/`false` pre-C23 is the easiest.

kraflab:
If we just change dboolean to int explicitly, will anything change? Is it true that `dboolean x = 3;` would have `x` as `3` and not `true` currently? I just want to make sure there isn't any magic happening somewhere.

bkoropoff:
Yeah, C will just implicitly convert integers to enums (even an invalid value not in the enum) without a peep unless you turn on some really high-octane warnings, so it doesn't change anything from the status quo.

kraflab:
Alright, then let's go for that.

FtZPetruska:
Made the change in [b2889eb](https://github.com/kraflab/dsda-doom/pull/333/commits/b2889ebca6d33575a4f77fd2dc6b865cc04dc433).

I also set the `C_STANDARD` property to `99` on the CMake target. According to [the docs](https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD.html), it is not a hard requirement unless `C_STANDARD_REQUIRED` is set to `ON`.

prboom2/src/doomtype.h
```diff
@@ -39,14 +39,14 @@
...
 #endif
 
-#ifndef __BYTEBOOL__
-#define __BYTEBOOL__
-/* Fixed to use builtin bool type with C++. */
-#ifdef __cplusplus
-typedef bool dboolean;
-#else
-typedef enum {false, true} dboolean;
+#ifndef __DBOOLEAN__
```
kraflab: Do we need these ifndef blocks? The whole file is wrapped in an ifndef.

FtZPetruska: Most likely not, they aren't checked anywhere else. Checking the blame, they were here from the initial revision. Would you like them removed?

kraflab: Yes please ✂️

kraflab:
Thanks

PR #335, opened 2023-04-08
build: remove -Wcast-align for now

bkoropoff:
There is too much code doing alignment-breaking casts at the moment.

Also reorder flags list so all -Wno- options occur at the end


kraflab:
By the way, I recently went through and resolved the remaining Linux gcc warnings since few were left.

PR #336, opened 2023-04-08
Move function declarations to headers

bkoropoff:
If a declaration is not in a header that is included by the file with the definition (so the compiler checks that they match), they can get out of sync during development and cause strange bugs.


kraflab:
Sorry but I don't agree with this change universally. I generally try to put commonly shared functions in headers, but there are cases where it just doesn't make sense to expose functions to many files by default and I think it can hurt the semantic value of header files. Basically I just see this as a failing of the C language and there's no perfect answer.

There may be some cases in this PR that make sense to move, but I would essentially have to do the work of the PR myself in order to confirm it makes sense, and there are other things I'm focused on.

kraflab:
To look at this from a different angle: let's say I run across a function that is referenced in this way. I would ask myself why it is like that and what is the best solution. It might be to do nothing but it might be an indicator that a separate file should be made and there should be some bigger refactoring. But mass-fixing every case may put things in a position where things are worse off conceptually.

bkoropoff:
Function arguments matching the prototype is a basic safety net (look at all the "fun" stuff heretic/hexen did with action function arguments and the resulting headaches).  If the prototype can get out of sync with the actual function then you don't know when you're breaking something until it crashes down the road.  I agree that header files are ugly in general, but if you need to access functions across compilation units then an ugly header is better than not using one.

kraflab:
I appreciate the thought, but so far I don't have the problem you're trying to solve, whereas I will immediately have the problem you will create :^)

PR #339, opened 2023-04-08
gl: improve comment explaining gld_RecalcVertexHeights

### LINKED ISSUE #334: [GL] `gld_RecalcVertexHeights` code shouldn't malign ATI
This function splits the left/right edges of wall polygons for each adjacent floor/ceiling so that there is an intermediate vertex for each height level, apparently because old ATI cards would have seams or other artifacts otherwise.  If this is not a concern on video cards dsda-doom needs to support, this code can be yanked out for extra performance and cleaner code.

bkoropoff:
Closes #334


kraflab:
👍

PR #341, opened 2023-04-09
More dead code/unused variables

bkoropoff:
Not sure about all of these, let me know which ones you want


prboom2/src/MUSIC/dbopl.c
```diff
@@ -1249,53 +1249,19 @@ void Chip__WriteReg(Chip *self, Bit32u reg, Bit8u val ) {
   }
 }
 
-Bit32u Chip__WriteAddr(Chip *self, Bit32u port, Bit8u val ) {
```
bkoropoff, line 1252: Might want to compare against upstream Dosbox OPL code and see if these are also unused there or there's missing emulation features that should be merged here.

prboom2/src/r_data.c
```diff
@@ -625,7 +625,7 @@ int R_SetSpriteByIndex(patchnum_t *patchnum, spritenum_t item)
 
 int R_NumPatchForSpriteIndex(spritenum_t item)
 {
-  if (item < 0 || item >= num_sprites)
```
bkoropoff, line 628: This is a "dead" condition because `spritenum_t` has no negative values, and I didn't see anything that cast negative values to it, but I'll double-check.

kraflab:
Hopefully I will have some time to go through this one soon.

prboom2/src/r_data.c
kraflab: I'd leave this as is just to be safe. Even if nothing is doing it now, it might at some point 😸

prboom2/src/MUSIC/midifile.c
```diff
@@ -752,6 +752,7 @@ void MIDI_RestartIterator(midi_track_iter_t *iter)
 
 
 
+#if 0
```
kraflab, line 755: I'd rather preserve this file as much as possible.

prboom2/src/MUSIC/dbopl.c
kraflab: I'd rather preserve this file as much as possible for now.

prboom2/src/r_drawflush.inl
```diff
@@ -45,14 +45,18 @@
 //
 static void R_FLUSHWHOLE_FUNCNAME(void)
 {
+#if !(R_DRAWCOLUMN_PIPELINE & RDC_FUZZ)
```
kraflab, line 48: Going to avoid making changes here for now because of the software fuzz PR, we can recheck this after that is merged.

kraflab:
I've merged most of the changes and left comments on the remaining ones, so I'll close this one.

PR #343, opened 2023-04-11
CMake: Track and install DLL dependencies outside vcpkg

FtZPetruska:
When using vpckg, dependent DLLs are copied to the build directory and installed alongside the executable (using `X_VCPKG_APPLOCAL_DEPS_INSTALL`). This PR aims to provide similar support for non-vcpkg users on Windows (MSYS2, manually-built dependencies).

To achieve this, every Find module had to be reworked to capture whether the library is `SHARED` or `STATIC`. For this to work:
- First, we search for a DLL using `find_file` with a list of the possible names it may have (it is not uncommon that vcpkg and MSYS2 will name them differently).
- We call `find_library` as usual.
- If a DLL is found or the library ends in `.dylib` or `.so`, it is safe to assume it is shared, otherwise assume static.
- When creating the target, if a DLL is found add it as the `IMPORTED_LOCATION` and the library as `IMPORTED_IMPLIB`.

For libraries such as Dumb or SDL2* that may have a debug variant also installed. Search for both a `_DLL_RELEASE` and `_DLL_DEBUG` variant and use a custom module similar to `SelectLibraryConfigurations` but for DLLs.

The SDL2 Find module has been updated to better match what upstream is doing, that is:
- When a static library is found, provide `SDL2::SDL2-static`.
- When a shared library is found, provide `SDL2::SDL2`.
- If only the static library is found, make `SDL2::SDL2` an alias of `SDL2::SDL2-static`.

During install, use the `RUNTIME_DEPENDENCIES` argument to capture all the dependencies and install them along the executable. This is specifically run:
- Only on `WIN32` platforms, while dependency tracking works for other platforms, it does not set RPATH, may just copy a bunch of shared libraries to `/usr/local/lib` if you're not careful. And it already works great *without*.
- When vcpkg is *not* provided. vcpkg provides its own mechanism to achieve that, and trying to use `RUNTIME_DEPENDENCIES` on top of it breaks.
- When CMake's version is =>3.21. This avoids bumping the minimum version all the way up to 3.21 when only Windows users benefit from it.

One drawback is that unlike vcpkg, DLLs are not copied to the build directory. This is okay for MSYS2 users as the required DLLs are in the PATH when ran from an MSYS Terminal. I tried using the `$<TARGET_RUNTIME_DLLS:tgt>` generator expression, but it does not list transitive dependencies unfortunately.

The build guide for Windows has also been updated accordingly.

---

I was able to test the changes on:
- Windows using vcpkg (static and shared)
- Windows using MSYS2 UCRT64
- macOS using vcpkg (static)
- macOS using Homebrew (shared)
- Ubuntu using vcpkg (static)
- Ubuntu using system libraries

### DIFF
--- a/prboom2/cmake/FindOgg.cmake
+++ b/prboom2/cmake/FindOgg.cmake
@@ -37,6 +37,8 @@ The following cache variables may also be set:
 
 ``Ogg_INCLUDE_DIR``
   The directory containing ``ogg/ogg.h``.
+``Ogg_DLL``
+  The path to the Ogg Windows runtime.
 ``Ogg_LIBRARY``
   The path to the Ogg library.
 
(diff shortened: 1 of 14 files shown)


rfomin:
I confirm it works in my MSYS2 installation, great work!

I have a question, let's say we want most of the libraries to be built statically and some of them dynamically, like SDL2. Will this system work? I tried it in Woof with overlay triplets, it works but `X_VCPKG_APPLOCAL_DEPS_INSTALL` fails.

FtZPetruska:
> I have a question, let's say we want most of the libraries to be built statically and some of them dynamically, like SDL2. Will this system work? I tried it in Woof with overlay triplets, it works but `X_VCPKG_APPLOCAL_DEPS_INSTALL` fails.

Hmm, I can't seem to be able to reproduce your issue. I tried building Woof with the following custom triplet and the install succeeded with `X_VCPKG_APPLOCAL_DEPS_INSTALL`:

```cmake
set(VCPKG_TARGET_ARCHITECTURE x64)
set(VCPKG_BUILD_TYPE release)
set(VCPKG_CRT_LINKAGE dynamic)
set(VCPKG_LIBRARY_LINKAGE static)
if(PORT MATCHES "sdl2")
  set(VCPKG_LIBRARY_LINKAGE dynamic)
endif()
```

rfomin:
> Hmm, I can't seem to be able to reproduce your issue...

I tried almost the same triplet, only with a static CRT. The installation works, but the dynamic DLLs are not copied. Anyway, thanks for the reply, I'll try again.

prboom2/cmake/FindOgg.cmake
```diff
@@ -48,32 +50,60 @@ pkg_check_modules(PC_OGG QUIET ogg)
 find_path(
   Ogg_INCLUDE_DIR
   NAMES ogg/ogg.h
-  HINTS ${PC_OGG_INCLUDEDIR})
+  HINTS "${PC_OGG_INCLUDEDIR}"
+)
+
+find_file(
+  Ogg_DLL
+  NAMES ogg.dll libogg.dll libogg-0.dll
```
kraflab, line 58: What are the consequences if there is a version with a different name for one of these find commands? Does any part of the build / install process fail?

FtZPetruska: I have tried purposefully making `find_file(Ogg_DLL)` fail and here's the behaviour I observed:
- Building still succeeds
- Whether I use vcpkg or not, the DLL is still copied
- The library is considered to be static, as such the static pkg-config flag would be used. This is mostly harmless, but may result in overlinking transitive dependencies.

kraflab:
👍

PR #345, opened 2023-04-12
Fix bug 344

bkoropoff:
The 2D frustum calculation for seg culling didn't match how `R_SetupPerspective` actually derived the projection matrix.  The fix derives the FOV directly from the projection matrix.

### DIFF
--- a/prboom2/src/gl_clipper.c
+++ b/prboom2/src/gl_clipper.c
@@ -335,6 +335,7 @@ static void gld_clipper_Clear(void)
 
 static angle_t gld_FrustumAngle(void)
 {
+  double fov;
   double floatangle;
   angle_t a1;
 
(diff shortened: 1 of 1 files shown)


prboom2/src/SDL/i_video.c
```diff
@@ -1278,7 +1278,8 @@ void I_UpdateVideoMode(void)
     sdl_renderer = SDL_CreateRenderer(sdl_window, -1, flags);
 
     // [FG] aspect ratio correction for the canonical video modes
-    if (SCREENHEIGHT == 200 || SCREENHEIGHT == 400)
+    if ((SCREENWIDTH == 320 && SCREENHEIGHT == 200) ||
```
kraflab: It's intended that this doesn't check width because it's supposed to support widescreen variants of the low res mode.

bkoropoff: A custom widescreen resolution will generally have square pixels, so this will just cause unexpected distortion.  I can drop it from the PR since it's unrelated.

prboom2/src/gl_clipper.c
```diff
@@ -350,10 +351,11 @@ static angle_t gld_FrustumAngle(void)
 
   // ok, this is a gross hack that barely works...
   // but at least it doesn't overestimate too much...
-  floatangle = 2.0f + (45.0f + (tilt / 1.9f)) * (float)gl_render_fov * ratio_scale / gl_render_multiplier / 90.0f;
-  a1 = (angle_t)xs_CRoundToInt(ANG1 * floatangle);
-  if (a1 >= ANG180)
+  fov = atan(1 / projMatrix[0]) * 360 / M_PI;
```
kraflab, line 354: There are a lot of places in the code base that use the `gl_render_fov` variable. Is this change going to cause any inconsistencies?

Also, is the comment here still the case? 😄

bkoropoff: This only affects seg culling, and using the GL projection matrix should prevent it from getting out of sync with actual rendering.

The actual horizontal render FOV is almost never `gl_render_fov`.  It's derived in a very roundabout manner by `M_ChangeFOV` and `R_SetupMatrix`.  This doesn't attempt to make the derivation any more sane.

The comment here is still correct, it's a very ad-hoc heuristic for the effective 2D view angle taking into account view pitch.

kraflab:
Thanks 👍 
Just out of curiosity, will the culling changes have any effect on performance one way or another?

Issue #310, opened 2023-04-02
Autoload is case-sensitive on Linux


bkoropoff:
If you load `SCYTHE.WAD`, autoloads in `scythe.wad/` won't be loaded, etc.  Probably need to do case folding somewhere.

orffen:
Also WAD loading is case-sensitive - see https://github.com/kraflab/dsda-doom/pull/346#issuecomment-1508373549

mooreye:
I think it should stay that way - UNIX filesystems are case-sensitive, meaning that DOOM.WAD and doom.wad can exist in the same dir as 2 completely different files with potentially different contents.

jjbarr:
I would be inclined to agree that the behavior should remain as-is. Case-sensitivity is a property of the filesystem, not a property of DSDA-Doom

kraflab:
I don't have any plan to change this behavior right now, so going to close this issue. May do something about it eventually, but I'm inclined to "trust the filesystem" when it comes to naming and what constitutes a difference.

PR #346, opened 2023-04-14
Convert autoload paths to lowercase

bkoropoff:
Fixes issue #310


kraflab:
Isn't this going to break the existing autoload for all linux users?

bkoropoff:
Hmm, yes, it's not backward compatible.  The fact that this wasn't done originally complicates things.

kraflab:
There is also the issue that using `-file scythe` doesn't work if the wad is `SCYTHE.WAD` and also that autodetecting and searching for required wads (e.g., via demo footers) fails for the same reason.

PR #347, opened 2023-04-14
Avoid work in signal handler for SIGINT/SIGTERM

### LINKED ISSUE #342: Hangs on interrupt/crash on Linux (and maybe macOS)
I noticed that dsda-doom would occasionally hang when hitting Ctrl-C to interrupt it.  It looks like this is because it registers a signal handler that calls [non-async-signal-safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html) functions, e.g.:

- `vsnprintf`
- `snprintf`
- `fprintf`
- `strsignal`
- Anything registered via `I_AtExit`, e.g.
   - `glDeleteTextures` (`gl_detail.c`'s exit handler)
   - `free` (via `Z_Free`)
   - ...

POSIX signals can interrupt the program in literally any thread at any point during execution, with arbitrary locks or resources held or internal structures in an inconsistent state.  Even interrupting `fprintf` and then calling it again can hang on the mutex protecting the stdio buffers.

For `SIGINT`, it's best to set a variable and have the main loop call `I_SafeExit` at the next opportunity.  For anything else, it's probably better to register nothing at all and just let the process die unceremoniously on `SIGSEGV`, etc.

bkoropoff:
Tell the main loop to exit instead.

Closes #342

### DIFF
--- a/prboom2/src/SDL/i_main.c
+++ b/prboom2/src/SDL/i_main.c
@@ -104,6 +104,13 @@ void I_Init2(void)
 
 int signal_context;
 
+static volatile sig_atomic_t interrupted = 0;
+
+dboolean I_Interrupted(void)
+{
+  return interrupted;
+}
+
 static void I_SignalHandler(int s)
 {
   char buf[2048];
@@ -119,6 +126,11 @@ static void I_SignalHandler(int s)
   I_Error("The game has crashed!\n"
           "Please report the following information: %s (0x%04x)",
           buf, signal_context);
+}
+
+static void I_IntHandler(int s)
+{
+  interrupted = 1;
 }
 
 static void PrintVer(void)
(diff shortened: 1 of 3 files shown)


prboom2/src/SDL/i_main.c
```diff
@@ -121,6 +128,11 @@ static void I_SignalHandler(int s)
           buf, signal_context);
 }
 
+static void I_IntHandler(int s)
```
kraflab, line 131: Could you tell me programmatically what happens when this function is called and when it returns? Does the program execution jump here and then back to the instruction it was on when it got the signal?

bkoropoff: Yes, essentially.  The OS arranges for the signal handler to run in a new frame on the call stack, after which execution resumes at precisely the point where the signal was delivered.

kraflab: Cool thanks

PR #349, opened 2023-04-15
PortMidi: Optimize reset messages

ceski-1:
This reduces the amount of messages sent during a reset, similar to Chocolate, Woof, etc. This also fixes a regression where wads with mid-level music changes were causing the game to pause briefly (https://github.com/coelckers/prboom-plus/files/7654830/MusChange.zip). Ideally, music should be handled on a separate thread, but this is a simple solution for now.


kraflab:
Thanks 👍

PR #350, opened 2023-04-16
wad_state: fix out-of-bound memory write

bkoropoff:


prboom2/src/dsda/wad_stats.c
```diff
@@ -44,7 +44,7 @@ static void dsda_EnsureMapCount(int count) {
     while (wad_stats.maps_size < wad_stats.map_count)
       wad_stats.maps_size = wad_stats.maps_size ? wad_stats.maps_size * 2 : 32;
     wad_stats.maps = Z_Realloc(wad_stats.maps, sizeof(*wad_stats.maps) * wad_stats.maps_size);
-    memset(wad_stats.maps + old_size * sizeof(*wad_stats.maps),
+    memset(wad_stats.maps + old_size,
```
kraflab, line 47: I don't see it, what was wrong?

bkoropoff: Pointer addition already implicitly multiplies by the pointee size.

kraflab: aha, true true, thanks

PR #352, opened 2023-04-19
Add simple wipe speed control

bkoropoff:


kraflab:
This isn't allowed to be used in the context people were requesting it. I also don't think it's worth having for either casual play or tasing, since it can be disabled if people don't want it and otherwise it isn't really an invasive effect.

PR #353, opened 2023-04-20
zip: Cleanup at exit and use unique names

FtZPetruska:
- [X] Ensure unique names
- [x] Clean up temporary directories when shutting down

---

This PR changes the extraction directory to be prefixed by a file count. This avoids an issue where passing two zips with the same filename would "ignore" the first one.

Some systems, most notably Windows, do not clean their temp directory automatically by default. How should we go about cleaning it up ?
- Make it a concern of the `zipfile` module. Every time a zip file is decompressed, its destination directory is added to a list, and a cleanup function is ran AtExit, so even if an error is encountered, directories will be cleaned up.
- Run cleanup in `I_Quit` like [Woof does](https://github.com/fabiangreffrath/woof/blob/11c652b67e6144ac66f7b45e9942b4d963fa0542/src/d_quit.c#L52). The idea would be to have `D_AddZip` to add the path to a list and clean it up when shutting down. The issue with that approach is that it is not run when an error is encountered.

What do you think @kraflab ?

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1525,24 +1525,16 @@ static void EvaluateDoomVerStr(void)
 
 static void D_AddZip(const char* zipped_file_name, wad_source_t source)
 {
-  dsda_string_t temporary_directory;
   char* full_zip_path;
-  static unsigned int file_counter = 0;
+  const char* temporary_directory;
 
   full_zip_path = I_RequireZip(zipped_file_name);
-  dsda_StringPrintF(&temporary_directory, "%s/%u-%s", I_GetTempDir(), file_counter, dsda_BaseName(zipped_file_name));
-  if (M_IsDir(temporary_directory.string))
-    if (!M_RemoveFilesAtPath(temporary_directory.string))
-      I_Error("D_AddZip: unable to clear tempdir %s\n", temporary_directory.string);
-  M_MakeDir(temporary_directory.string, true);
+  temporary_directory = dsda_UnzipFile(full_zip_path);
 
-  dsda_UnzipFile(full_zip_path, temporary_directory.string);
+  LoadWADsAtPath(temporary_directory, source);
+  LoadDehackedFilesAtPath(temporary_directory, true);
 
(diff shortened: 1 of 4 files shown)


prboom2/src/d_main.c
```diff
@@ -1527,9 +1527,10 @@ static void D_AddZip(const char* zipped_file_name, wad_source_t source)
 {
   dsda_string_t temporary_directory;
   char* full_zip_path;
+  static unsigned int file_counter = 0;
```
kraflab: Let's reorganize where the logic is and track things in the module. In order for that to make sense the module needs to do the decision making about where the zip file should be unzipped (i.e., the fact that it is a temporary directory). I would suggest the content of `D_AddZip` to be something like this:
```c
{
  char *temporary_directory;
  char *full_zip_path;

  full_zip_path = I_RequireZip(zipped_file_name);
  temporary_directory = dsda_UnzipFile(full_zip_path);

  if (!temporary_directory)
    I_Error("Unable to unzip %s", full_zip_path);

  LoadWADsAtPath(temporary_directory, source);
  LoadDehackedFilesAtPath(temporary_directory, true);

  Z_Free(temporary_directory);
}
```
The cleanup function can be called from the end of `I_EssentialQuit`

kraflab: Btw, just noticed this function is also missing `Z_Free(full_zip_path)` at the end.

FtZPetruska: Thanks for the feedback. I've made the changed in [ece9a63](https://github.com/kraflab/dsda-doom/pull/353/commits/ece9a637a5b9310437d84d1271f948fd17d1c4e9).

Only thing I did not add was this:
```c
if (!temporary_directory)
    I_Error("Unable to unzip %s", full_zip_path);
```
Since `I_Error` would have been called already from the error that would make `dsda_UnzipFile` return `NULL`.

> Btw, just noticed this function is also missing `Z_Free(full_zip_path)` at the end.

Hadn't noticed the `I_Require*` needed to be freed oops 😅

kraflab:
thanks 👍

PR #355, opened 2023-04-22
id_list: allow and handle negative ids

bkoropoff:
These actually show up in some maps.  Fixes linedef specials (and therefore demo desyncs) in telelab.


kraflab:
Thanks!

PR #357, opened 2023-04-24
GL: handle heretic torch flicker colormaps

bkoropoff:
Fixes crash when picking up torch.


kraflab:
Thanks 👍 

Is there any performance concern with this? I guess the size of glTexExID increases by a factor of 3.

PR #358, opened 2023-04-28
Remove vcpkg overlay ports.

FtZPetruska:
Using overlay ports is no longer necessary:
- FluidSynth has been updated upstream to 2.3.2, along with other fixes not included in the overlay (https://github.com/microsoft/vcpkg/pull/30625)
- the dumb port has been merged upstream (https://github.com/microsoft/vcpkg/pull/30740)

Moreover, by now, both changes have been rolled out on the GitHub Actions runner, so it should be safe to remove without breaking CI.


kraflab:
Thanks 👍

PR #359, opened 2023-04-30
Scale fuzz in the software renderer

Lovey847:
For a long time now, software renderer fuzz has been drawn using an algorithm that doesn't adapt to screen size, making fuzz harder to see at higher resolutions. This PR overwrites the old algorithm with a new one that scales the fuzz to match the window size.

From my tests at 1080p, it performs better than the original algorithm, but it could be optimized further.

Comparisons:
fuzzComparison1
fuzzComparison2
fuzzComparison3

### DIFF
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -44,6 +44,7 @@
 #include "g_game.h"
 #include "am_map.h"
 #include "lprintf.h"
+#include "i_system.h"
 
 #include "dsda/stretch.h"
 
@@ -112,12 +113,18 @@ static int fuzzpos = 0;
 // Fuzz cell size for scaled software fuzz
 static int fuzzcellsize;
 
-// Fuzz intensity table. Contains a pattern of fuzz intensities, where a
+// A pattern of fuzz intensities, where a
 // fuzz intensity is how much a fuzz cell darkens its contents
 static const byte fuzzintensity[FUZZTABLE] = {
   0, 1, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 2, 3, 4, 0, 1, 2,
   0, 0, 0, 0, 1, 0, 1, 0, 0, 1, 2, 0, 0, 1, 2, 3, 4, 0, 0, 0, 0, 1, 0, 0, 1, 0
 };
+
+// Fuzz intensity table for every colormap
+// A fuzz intensity table contains 5 tables of 256 colors, where each table
+// contains darkened colors corresponding to the intensity
+// (table 0 is intensity 0, table 1 is intensity 1, etc.)
+static byte *fuzzintensitytables;
 
 // render pipelines
 #define RDC_STANDARD      1
@@ -463,6 +470,38 @@ void R_InitBuffersRes(void)
   temp_x = 0;
 }
 
+static void R_InitFuzzIntensityTables(void)
+{
+  int i, j, k, l;
+  byte *table;
+
+  fuzzintensitytables = Z_Malloc(sizeof(*fuzzintensitytables) * 5 * 256 * numcolormaps);
+
+  for (i = 0; i < numcolormaps; i++)
+  {
+    table = fuzzintensitytables + 5 * 256 * i;
+
+    for (j = 0; j < 5; j++)
+    {
+      for (k = 0; k < 256; k++)
+      {
+        // Get accurate shade for fuzz intensity
+        table[j * 256 + k] = colormaps[i][6 * 256 + k];
+
+        for (l = 0; l < j; l++)
+        {
+          table[j * 256 + k] = colormaps[i][6 * 256 + table[j * 256 + k]];
+        }
+      }
+    }
+  }
+}
+
+static void R_FreeFuzzIntensityTables(void)
+{
+  Z_Free(fuzzintensitytables);
+}
+
 //
 // R_InitBuffer
 // Creats lookup tables that avoid
@@ -480,6 +519,12 @@ void R_InitBuffer(int width, int height)
     fuzzcellsize = (SCREENHEIGHT + 100) / 200;
   else
     fuzzcellsize = (SCREENWIDTH + 160) / 320;
+
+  if (!fuzzintensitytables)
+  {
+    R_InitFuzzIntensityTables();
+    I_AtExit(R_FreeFuzzIntensityTables, true, "R_FreeFuzzIntensityTables", exit_priority_normal);
+  }
 }
 
 //
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -107,18 +107,6 @@ static const byte   *tempfuzzmap;
 //#define FUZZOFF (SCREENWIDTH)
 #define FUZZOFF 1
 
-static const int fuzzoffset_org[FUZZTABLE] = {
-  FUZZOFF,-FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,
-  FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,
-  FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,
-  FUZZOFF,-FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,
-  FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,-FUZZOFF,FUZZOFF,
-  FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,
-  FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF
-};
-
-static int fuzzoffset[FUZZTABLE];
-
 static int fuzzpos = 0;
 
 // Lovey01 04/29/2023: Scaled software fuzz
@@ -491,18 +479,13 @@ void R_InitBuffersRes(void)
 
 void R_InitBuffer(int width, int height)
 {
-  int i;
-
   drawvars.topleft = screens[0].data;
   drawvars.pitch = screens[0].pitch;
 
-  for (i=0; i<FUZZTABLE; i++)
-    fuzzoffset[i] = fuzzoffset_org[i]*screens[0].pitch;
-
   if (!tallscreen)
-    fuzzcellsize = (WIDE_SCREENHEIGHT + 100) / 200;
+    fuzzcellsize = (SCREENHEIGHT + 100) / 200;
   else
-    fuzzcellsize = (WIDE_SCREENWIDTH + 160) / 320;
+    fuzzcellsize = (SCREENWIDTH + 160) / 320;
 }
 
 //
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -112,19 +112,11 @@ static int fuzzpos = 0;
 // Fuzz cell size for scaled software fuzz
 static int fuzzcellsize;
 
-// Fuzz colormap table. Contains colormaps for fuzz intensities, where a
-// fuzz intensity is how strong a fuzz cell is, or how much it darkens its
-// contents
-// NOTE: Formula to find a colormap for an intensity is
-// (6 + 21 - (9 - i) * (9 - i) * 5 / 19), where i is the fuzz intensity
-static const byte fuzzcmaps[FUZZTABLE] = {
-  6, 11, 6, 11, 6, 6, 11,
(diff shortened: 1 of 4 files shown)


prboom2/src/r_draw.c
```diff
@@ -107,20 +107,24 @@ static const byte   *tempfuzzmap;
...
-};
-
-static int fuzzoffset[FUZZTABLE];
-
 static int fuzzpos = 0;
 
+// Lovey01 04/29/2023: Scaled software fuzz
+static int fuzzcellsize;
+
+// NOTE: Formula is (6 + 21 - (9 - i) * (9 - i) * 5 / 19), where i is the fuzz
```
kraflab: Does this come from somewhere or did you just find something that worked?

Lovey847: If you're talking about the formula, I just worked out numbers in a calculator until it matched the sequence.

prboom2/src/r_draw.c
```diff
@@ -107,20 +107,24 @@ static const byte   *tempfuzzmap;
...
+// intensity
+// NOTE2: Not const because it's stored closer to fuzzpos this way
```
kraflab: Make it const. As far as I know, zero-initialized data is stored in a separate block from non-zero values, const or otherwise. But, even if that isn't the case, we aren't about to make decisions based on assumptions of where addresses are going to end up. That's a level of fragility that would be a new record even for the doom source.

Rather than violating the meaning of the data, you could instead package things together in a way that satisfies your concern while creating a clean interface that communicates purpose, something like this:
```c
typedef struct {
  int pos;
  int cell_size;
  const byte cmaps[FUZZTABLE];
} fuzz_t;

static fuzz_t fuzz = ...
```

prboom2/src/r_draw.c
```diff
@@ -107,20 +107,24 @@ static const byte   *tempfuzzmap;
...
-  FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,-FUZZOFF,FUZZOFF,
-  FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,
-  FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF
-};
-
-static int fuzzoffset[FUZZTABLE];
-
 static int fuzzpos = 0;
 
+// Lovey01 04/29/2023: Scaled software fuzz
```
kraflab: Please drop the annotation comments like this. I'm aware that some ports do this, but it's not generally done in dsda-doom.

prboom2/src/r_draw.c
Lovey847: Oh yeah you're right, forgot about bss.
Idk what your code preferences for new code is, especially when it concerns efficiency. I can make it const real quick

prboom2/src/r_draw.c
```diff
@@ -587,3 +591,14 @@ int R_GetFuzzPos()
 {
   return fuzzpos;
 }
+
+void R_ResetFuzzCol(int rows)
+{
+  R_ResetColumnBuffer(); // Flush current columns before changing fuzzpos
```
kraflab: Try to avoid redundant comments like this:
```c
A(); // call A before B
B();
```
The code is pretty clear here :^)

prboom2/src/r_draw.c
```diff
@@ -587,3 +591,14 @@ int R_GetFuzzPos()
...
+  fuzzpos = (fuzzpos + (rows / fuzzcellsize)) % FUZZTABLE;
+}
+
+void R_NewFuzzCol(int x, int rows)
+{
+  if (!(x % fuzzcellsize)) R_ResetFuzzCol(rows);
```
kraflab: It's a bit confusing in this code base because there are many styles, but please take at least 2 lines for an `if`.

prboom2/src/r_draw.c
kraflab: Basically what I'm asking is why and how you went about getting this formula and the values in the table. I can see that the formula and table are connected, but why this formula / why those values? Possibly it would help to spell out what `c` is in this table name or to use something else entirely. (I am also wondering if this variable name needs a comment "intensity" to explain what it is, why aren't we calling it "fuzz_intensity" and getting rid of the comment 😄)

prboom2/src/r_draw.c
Lovey847: Eh, it could look more like
`SeeminglyUnrelatedRoutine();`
`ChangingFuzzpos();`
to someone not as familiar with doom's software renderer. Maybe the comment could be better, such as "Make sure columns are drawn before changing fuzzpos"

prboom2/src/r_draw.c
```diff
@@ -475,13 +478,13 @@ void R_InitBuffersRes(void)
...
 {
-  int i;
-
   drawvars.topleft = screens[0].data;
   drawvars.pitch = screens[0].pitch;
 
-  for (i=0; i<FUZZTABLE; i++)
-    fuzzoffset[i] = fuzzoffset_org[i]*screens[0].pitch;
+  if (!tallscreen)
+    fuzzcellsize = (SCREENHEIGHT + 100) / 200;
```
kraflab, line 519: Is the `+ 100` and `+ 160` here basically just for rounding?

Lovey847: Yeah, rounding to nearest

prboom2/src/r_draw.c
kraflab: That comment wouldn't have any meaning either. The purpose of a comment would be to explain _why_ something is happening. It's very clear _what_ is happening. When the _what_ is unclear, it means there's a problem in the names of things or in how the code is organized generally (i.e., a comment is the wrong solution). I see a function purporting to reset the fuzz column, and it calls a function that resets the column buffer. That seems reasonable to me - of course that doesn't explain everything that happens, but the reader can choose to follow the function if they want more details.

prboom2/src/r_draw.c
kraflab: Humans are very bad at these kinds of low-level optimizations and are more likely to work against the compiler. The number one thing is for code to clearly communicate its purpose to the reader / compiler. Concerns to keep in mind are things like algorithmic complexity or high memory consumption. Otherwise something only needs to be optimized by a human if it is a demonstrable bottleneck.

prboom2/src/r_draw.c
Lovey847: I actually knew this was poorly documented, basically fuzz "intensity" was how strong a fuzz cell was, or how much it darkened its contents. With the old table of offsets, since I wasn't grabbing from the pixels above or below anymore, I had to keep track of the fuzz intensity by checking whether the fuzz offset was positive or negative. Then I made a table for the intensity values and got rid of the code that handled the intensity, so now it's unknown what the "intensity" is. cmap means colormap, though that might not be the right terminology since it's an offset into tempfuzzmap. If you're wondering why I left the note here in the first place, it's for anyone planning to optimize it, because the formula could possibly be faster than the lookup.

prboom2/src/r_draw.c
Lovey847: The proposed comment does explain the why. "Make sure columns are drawn before changing fuzzpos"

prboom2/src/r_draw.c
Lovey847: I usually try to optimize things too early, it's a bad habit I have. Thanks for the advice

prboom2/src/r_draw.c
kraflab: That is merely stating what the code does - as a reader I can usually assume that if A happens before B it is because A should happen before B. A comment comes in when it's unclear why that thing needs to happen at all or why it needs to happen at that moment, neither of which are covered by the comment here. Another case would be if there is something shockingly unexpected that you need to communicate to the reader to ensure they don't break it in the future (e.g., that there is a memory hack used on a struct and the position of some element must not change). Regardless of the semantics of "why" I don't think a comment is needed in this case.

Lovey847: It's stating it resets the column buffer to make sure the columns are drawn before changing fuzzpos, which explains why it's happening.
It can be unclear what "ResetColumnBuffer" means exactly, so that's why I added the comment.
It addresses something that could potentially be a mystery to the reader, and I don't think the reader should have to go sifting through the code to figure out why it's being called, especially when there was originally a comment that explained it.

prboom2/src/r_draw.c
Lovey847: This documentation is more descriptive now

prboom2/src/r_draw.c
```diff
@@ -587,3 +593,17 @@ int R_GetFuzzPos()
...
+
+void R_ResetFuzzCol(int rows)
+{
+  // Make sure previous columns are drawn before changing fuzzpos
+  R_ResetColumnBuffer();
+
+  fuzzpos = (fuzzpos + (rows / fuzzcellsize)) % FUZZTABLE;
+}
+
+void R_NewFuzzCol(int x, int rows)
```
kraflab: I don't feel that this function name is quite right. It's always tricky to name this kind of function though. I'd probably go with `R_CheckFuzzCol` for a vague connection to `R_CheckPlane`, which also implies that something may or may not happen. I'd also change `rows` to `height` in these functions.

Lovey847: Ah yeah I agree

prboom2/src/r_draw.c
```diff
@@ -107,20 +107,26 @@ static const byte   *tempfuzzmap;
...
+// Fuzz cell size for scaled software fuzz
+static int fuzzcellsize;
+
+// Fuzz colormap table. Contains colormaps for fuzz intensities, where a
+// fuzz intensity is how strong a fuzz cell is, or how much it darkens its
+// contents
+// NOTE: Formula to find a colormap for an intensity is
+// (6 + 21 - (9 - i) * (9 - i) * 5 / 19), where i is the fuzz intensity
+static const byte fuzzcmaps[FUZZTABLE] = {
+  6, 11, 6, 11, 6, 6, 11,
```
kraflab: So the `6` is coming from `tempfuzzmap[6*256+(col)]`, right? But where do the other numbers come from exactly?

Lovey847: Even if I did understand what the formula was doing at some point, it was written months ago and I don't really know now.

prboom2/src/r_draw.c
kraflab: This is what I meant above when I said if the core problem is actually an unclear name, the solution is to change the name or how the code is structured. In other words, to instead call a function like `R_DrawColumns`. But in this case I'm not suggesting to do that change, because we are in the context of the software renderer and should maintain the nomenclature, at least for this PR. Looking through that lens, the code makes sense as it is. Basically this isn't the place to solve the knowledge gap, because then we would need a thousand comments 😄

prboom2/src/r_draw.c
kraflab: It's important that we don't change the color of the fuzz. No pixel in the software fuzz should have a color that it couldn't have in vanilla. Basically I want to make sure solving the pattern doesn't introduce a separate problem :^)

The sample screenshots appear to have a bit darker shading, but I guess I will have to verify this.

prboom2/src/r_draw.c
Lovey847: Ah, you wanna remove it now so that in the future, when the function's called something more descriptive, the comment isn't mostly redundant. I can remove it in anticipation for that change.

prboom2/src/r_draw.c
Lovey847: Oh yeah the fuzz will be a slightly different shade than vanilla, but only because the colormaps don't offer that amount of granularity. The colormap values were chosen to be the closest fit. If this discrepancy is a problem, I'll see what I can do about getting more precise shades.

Lovey847: Now the fuzz's colors are accurate to vanilla

prboom2/src/r_drawflush.inl
```diff
@@ -161,12 +196,8 @@ static void R_FLUSHQUAD_FUNCNAME(void)
...
    int count;
 #if (R_DRAWCOLUMN_PIPELINE & RDC_FUZZ)
-   int fuzz1, fuzz2, fuzz3, fuzz4;
-
-   fuzz1 = fuzzpos;
-   fuzz2 = (fuzz1 + tempyl[1]) % FUZZTABLE;
-   fuzz3 = (fuzz2 + tempyl[2]) % FUZZTABLE;
-   fuzz4 = (fuzz3 + tempyl[3]) % FUZZTABLE;
+   // Only whole flushes are supported for fuzz
+   return;
```
kraflab, line 200: Why was code here before but now there's nothing? What changed?

Lovey847: I removed code that wasn't used by the new algorithm. There are 3 flush types: a whole flush which draws it column by column, a quad flush which draws 4 pixels in the same row at a time, and lastly a headtail flush which draws the pixels above and below the region a quad flush covers. You'll notice in the headtail flush that I put the same comment, only there it's a wrapper for the whole flush. Headtail flush and quad flush are called in sequence so they can't both be wrappers. As to why the new algorithm doesn't use a quad flush, I didn't implement it 'cause there's no need.

kraflab:
I'm most likely not going to merge this until after 0.26 fyi. So if it seems like I forgot about it, rest assured I will return to it :^)

PR #361, opened 2023-05-03
GL: fix mysterious single-frame missing walls

bkoropoff:
I was trying to track town a problem in my experimental branch where walls would occasionally not be rendered for a frame when rounding a corner (very noticeable with `flashing_hom`), and started to suspect that it might be an existing bug.  The GL renderer uses a `validcount`-like counter to make sure it renders a wall only once per frame, but it only uses a byte for the counter.  This wraps very rapidly even at modest framerates, so there's a roughly 1-in-256 chance that a wall that just became visible in a frame will not be rendered because its count happens to match the wrapped global count.

A related problem is probably responsible for the random rendering glitch during the wipe effect: the counter values for all walls are reset on map start, but not the global counter.  This means you have a 1-in-256 chance of *all* walls not being rendered during the wipe.

-----------------

- Reset rendermarker when resetting linerendered array
- Use int for rendermarker so it doesn't wrap so quickly

### DIFF
--- a/prboom2/src/gl_intern.h
+++ b/prboom2/src/gl_intern.h
@@ -535,7 +535,8 @@ typedef struct vbo_xy_uv_rgba_s
 
 // preprocessing
 extern byte *segrendered; // true if sector rendered (only here for malloc)
-extern byte *linerendered[2]; // true if linedef rendered (only here for malloc)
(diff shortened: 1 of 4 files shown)


kraflab:
Wow, this has been a mystery forever, awesome fix 😎

Issue #207, opened 2023-02-08
Custom colormap error in indexed light mode


kraflab:
https://www.doomworld.com/forum/topic/130661-hellcinerator-definitive-edition-feb-2023-large-mbf21-map/

![dsda-colourmap-example-2](https://user-images.githubusercontent.com/25529060/217593037-8302d5c4-d3ca-4325-bfc9-fb59542e4876.png)

bkoropoff:
This fixes itself if you switch to software mode and back to OpenGL, so I suspect a colormap texture is being cached and not cleared on a colormap change.

kraflab:
@bkoropoff have you looked into this more? I will start investigating after my current task is done, but I don't want to step on your toes.

bkoropoff:
I haven't looked at it beyond finding that something gets reset on re-init of GL.

kraflab:
- the colormap is wrong even if the player spawns in the alternate colormap area
- fixes on re-init, but only if you are in the alternate colormap area
- if you reload the map it keeps state (fixed or not fixed)
- if you switch maps and return, it is wrong again
- the default colormap area is always correct regardless of the conditions

Maybe the problem is in the level load precache stuff?

kraflab:
Fixed by https://github.com/kraflab/dsda-doom/pull/362

PR #362, opened 2023-05-04
GL: fix non-default indexed colormaps

bkoropoff:
Colormap lookups are done in the fragment shader, so don't do them when generating the texture.

Closes issue #207

### DIFF
--- a/prboom2/src/gl_texture.c
+++ b/prboom2/src/gl_texture.c
@@ -200,6 +200,8 @@ void* NewIntDynArray(int dimCount, int *dims)
   return buffer;
 }
 
+#define INVULN_PLAYER_CM 8
+
 // e6y
 // Get index of player->fixedcolormap for GLTexture().glTexExID array
 // There are 9 known values for player->fixedcolormap:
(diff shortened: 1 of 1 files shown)


kraflab:
This fixes 207 but it breaks the invuln colormap unfortunately

bkoropoff:
Not sure why invuln is handled differently, but I added a check for it.  Also expanded the fix to raws and patches.

prboom2/src/gl_texture.c
```diff
@@ -433,40 +436,26 @@ static void gld_AddPatchToTexture_UnTranslated(GLTexture *gltexture, unsigned ch
...
-            }
-            else
-            {
-              buffer[pos+0]=playpal[colormap[source[j]]*3+0];
-              buffer[pos+1]=playpal[colormap[source[j]]*3+1];
-              buffer[pos+2]=playpal[colormap[source[j]]*3+2];
-            }
+            // [XA] new indexed color mode: store the palette index in
+            // the R channel and get the colormap index in the shader.
+            buffer[pos + 0] = gltexture->player_cm == INVULN_PLAYER_CM ? colormap[source[j]] : source[j];
```
kraflab, line 443: I was wondering why we don't need a comp_sky check here, but there is a separate gld_AddIndexedSkyToTexture covering indexed mode. Just mentioning this for posterity.

kraflab:
Looks great 👍 
As far as I know, this was the last remaining error in indexed mode, it should now have parity 🥳

PR #363, opened 2023-05-04
Remove old shader mode

kraflab:
There were a lot of logical collapses in the gl code base when resolving everything to indexed light mode. Hi res and detail textures have disappeared, and various globals are no longer needed. There is no longer a "light mode" concept because there is only one opengl rendering style now.


andrey-budko:
dsda-doom.exe hg.wad
```
---------------------------
dsda-doom
---------------------------
createPatch: Unknown patch format STBAR.
---------------------------
ОК   
---------------------------
```

dsda-doom.exe vg.wad
```
---------------------------
dsda-doom
---------------------------
createPatch: Unknown patch format STKEYS0.
---------------------------
ОК   
---------------------------
```

kraflab:
I'm aware this costs support for a few wads, but I decided the benefits outweigh the costs. I'm interested in eventually coming up with a strategy to support these things in indexed mode somehow, but not yet.

kraflab:
Ah but it should detect the working patches. I guess we need to catch and disregard the hi res ones.

andrey-budko:
But it should not exit, right? There are "good" (no hires) STBAR and STKEYS0 lumps there.

kraflab:
Yes, thanks for the catch, this does need to be fixed. I had tested frozen time which loaded fine, but didn't check these.

andrey-budko:
Do we just need to restore this line?
`W_CoalesceMarkedResource("HI_START", "HI_END", ns_hires);`

kraflab:
Does that hide the lumps away when not using that namespace?

andrey-budko:
I think hires lumps are ns_global now, but they should not because of
```cpp

W_FindNumFromName(const char *name, int lump)
        { return W_FindNumFromName2(name, ns_global, lump); }

W_CheckNumForName(const char *name)
        { return W_CheckNumForName2(name, ns_global); }
```

kraflab:
Alright, should be a quick fix then, thanks for investigating 🤠

PR #364, opened 2023-05-05
Misc. texture cleanup

bkoropoff:
- Remove remainder of gl_detail.c
- Fail if multitexture is not available (required by indexed textures)
- Fail if non-power-of-2 textures are not available (indexed mode is completely broken without it, this seemed better than trying to fix it for extremely old GPUs)
- Remove various code paths that are effectively dead (e.g., mipmaps, texture filtering)
- Don't let user change texture format (it breaks indexed textures)
- Use GL_RG texture format for indexed textures to cut per-texture VRAM usage in half
- Deduplicate indexed textures that are identical to further save VRAM


kraflab:
That's some satisfying cleanup 😄 
Something like net -2500 lines including the last PR.
Going to give this a test run later and then make a build for the beta testers.

kraflab:
No issues on my side, thanks a bunch 👍

PR #365, opened 2023-05-05
heretic: show advisory patch on title page

fabiangreffrath:


prboom2/src/heretic/d_main.c
```diff
@@ -24,7 +24,12 @@
 
 static void Heretic_D_DrawTitle(const char *_x)
 {
-  D_SetPage("TITLE", 350, heretic_mus_titl);
+  D_SetPage("TITLE", 210, heretic_mus_titl);
```
kraflab, line 27: Why did this change?

fabiangreffrath: Because both states showing the title screen have to sum up to 350 tics, but the advisory patch is only displayed after 210 tics:

https://github.com/chocolate-doom/chocolate-doom/blob/0812f49b460239c873fbe46f16717d6aedf04070/src/heretic/d_main.c#L334-L344

kraflab: Ah ok

prboom2/src/d_main.c
```diff
@@ -612,7 +612,7 @@ static void D_PageDrawer(void)
   if (raven)
   {
     V_DrawRawScreen(pagename);
-    if (hexen && demosequence == 1)
+    if (raven && demosequence == 1)
```
kraflab: We're inside of a raven check, so you can just drop that part of the condition

kraflab:
Thanks

PR #368, opened 2023-05-07
Make GL fuzz effect emulate streakiness of software

bkoropoff:
The "streak" effect of software fuzz is due to runs of negative offsets in the fuzz offset table causing the same pixel to be copied and darkened several times in a row.  This shader emulates that using a lookup table derived from the original offset table.

Things to make this even more accurate:

- Make the start offset of each column be the sum of non-transparent pixels in all previous columns, rather than random as it is now.
- Actually copy offset background pixels and go through colormaps to perform darkening.  This would require a major change to the rendering pipeline to use intermediate textures so previous pixels can be read.

### DIFF
--- a/prboom2/data/lumps/glfp_fuzz.lmp
+++ b/prboom2/data/lumps/glfp_fuzz.lmp
@@ -2,12 +2,45 @@
 
 uniform sampler2D tex;
 uniform vec2 tex_d; // Texture dimensions
-uniform vec2 screen_res;
-uniform int time;
+uniform int count;
+
+const int darksize = 50;
+const float fdarksize = float(darksize);
+// Table of darkness steps for each pixel in a column, based on
+// offset table in software renderer
+const int darktable[darksize] = int[](
+  1, 2, 1, 2, 1, 1, 2, 1, 1, 2,
+  1, 1, 1, 2, 1, 1, 1, 2, 3, 4,
+  5, 1, 2, 3, 1, 1, 1, 1, 2, 1,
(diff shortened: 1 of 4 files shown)


bkoropoff:
doom01

Lovey847:
Here are more accurate multipliers for the fuzz, you'd replace "\<num\>" in the table with "dark\<num\>"
```
const float dark1 = 1.88235294117647058824;
const float dark2 = 3.41176470588235294118;
const float dark3 = 4.62745098039215686275;
const float dark4 = 5.64705882352941176471;
const float dark5 = 6.47058823529411764706;
```

How I calculated them: I used the algorithm `col = (col * (32 - 6) + 16) / 32` (algorithm for calculating the sixth lightmap in dcolors.c, which is the PLAYPAL and COLORMAP generator), and converted from a 0-255 range to what the shader wants.

Lovey847:
There's a slight issue with this algorithm at the moment, if you have 2 identical sprites drawn with fuzz, they'll have the same pattern because the only 2 factors for the RNG are the time passed and the texture X.
sameFuzz

bkoropoff:
Added the more accurate darkness steps and changed the uniform passed to the shader to a counter to avoid repeating patterns on the same frame.

[Screencast from 2023-05-09 17-32-16.webm](https://github.com/kraflab/dsda-doom/assets/2101303/25492ed9-4b7c-425d-b5e3-1ec694da9ae7)

kraflab:
Cool stuff 👍

PR #372, opened 2023-05-13
GL: tweak fuzz some more

bkoropoff:
- Update fuzz at game tick rate
- Use window space coordinates to compute fuzz for non-view weapon sprites, so scale is the same regardless of depth

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -750,13 +750,12 @@ void gld_DrawLine(int x0, int y0, int x1, int y1, int BaseColor)
 }
 
 
-void gld_StartFuzz(float width, float height)
+void gld_StartFuzz(int sprite, int width, int height, float ratio)
 {
   color_rgb_t color;
 
   // shader init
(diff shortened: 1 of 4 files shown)


prboom2/src/gl_main.c
```diff
@@ -2410,6 +2410,7 @@ void gld_ProjectSprite(mobj_t* thing, int lightlevel)
     scene_has_overlapped_sprites = true;
 
   sprite.index = gl_spriteindex++;
+  sprite.id = thing->index;
```
kraflab: The thing index is the incrementing value in the wad data, but it isn't a unique identifier. Every entity that spawns in the middle of a map has index `-1`. Maybe cast the thing address?

bkoropoff: This revealed some problems with handing large integers to the shader to convert to floats internally, so I had to rework things a bit, but I'm now using the address.

kraflab:
Thanks 👍 
I'm not sure how people will feel about this generally, but it's definitely the direction I want to go :^)

PR #375, opened 2023-05-16
detect patches in PNG format, substitute dummy patch

fabiangreffrath:
Fixes crashes with WADs that use a PNG patch as sky, e.g. egregor.wad

### DIFF
--- a/prboom2/src/r_patch.c
+++ b/prboom2/src/r_patch.c
@@ -324,11 +324,12 @@ static void FillEmptySpace(rpatch_t *patch)
 //
 //==========================================================================
 
-static dboolean CheckIfPatch(int lump)
+dboolean CheckIfPatch(int lump)
 {
   int size;
   int width, height;
   const patch_t * patch;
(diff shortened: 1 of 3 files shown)


prboom2/src/r_patch.c
```diff
@@ -339,6 +340,11 @@ static dboolean CheckIfPatch(int lump)
 
   patch = (const patch_t *)W_LumpByNum(lump);
 
+  // [FG] detect patches in PNG format early
+  magic = (const unsigned char *) patch;
+  if (magic[0] == 0x89 && magic[1] == 'P' && magic[2] == 'N' && magic[3] == 'G')
```
kraflab, line 345: This series of bytes is also a valid patch header, isn't it? :^)

kraflab:
I went with this instead: https://github.com/kraflab/dsda-doom/commit/6bf7a83828302ef2343bb589f4843e7f5ac548b9
Thanks for the heads up 👌

PR #376, opened 2023-05-17
Close wad file handles before cleaning temp files

bkoropoff:
You can't ordinarily delete a file on Windows when you still have it open.

------------

This prevents share access conflicts on Windows


kraflab:
👍

PR #379, opened 2023-05-18
Build fixes

jengelh:


kraflab:
I'm not going to update the .gitignore for everyone's local file location choices :^)
You can create your own local git exclusion if needed.
The extern issue: https://github.com/kraflab/dsda-doom/commit/65458c20db3f4a4c8af77521bb6c1ae1a40054f2

jengelh:
These are just the files that are generated under a normal Linux/cmake build. There is nothing overly specific to my machine in there.

Pedro-Beirao:
You should be using cmake from inside the build folder

fabiangreffrath:
> You should be using cmake from inside the build folder

You can even enforce that if you like:

https://github.com/fabiangreffrath/woof/blob/79bfb2b9321101e06916890d66803681a736c545/CMakeLists.txt#L32-L35

PR #380, opened 2023-05-20
GL: pass lookup table into fuzz shader as texture

bkoropoff:
This might address performance problems some people experience on older integrated GPUs.


bkoropoff:
@kraflab I got confirmation from someone with perf issues that this fixes it.

ghost:
> @kraflab I got confirmation from someone with perf issues that this fixes it.

👍👍

kraflab:
👍

PR #381, opened 2023-05-21
GL: don't stencil bleed from closed sectors

bkoropoff:
Fixes rendering glitch on cchest2 MAP05

------------

The flats of closed sectors aren't visible, so they shouldn't flood into adjacent missing wall textures.

## Before
doom00

## After
doom02


kraflab:
👍

PR #383, opened 2023-05-21
Fix zdoom polyobj rendering crash in software 

bkoropoff:


prboom2/src/hexen/po_man.c
```diff
@@ -66,7 +66,7 @@ static void ResetSegDrawingParameters(seg_t *seg)
   seg->v1->py = seg->v1->y;
   seg->v2->px = seg->v2->x;
   seg->v2->py = seg->v2->y;
-  seg->pangle = seg->angle;
+  seg->pangle = R_PointToAngle2(seg->v1->px, seg->v1->py, seg->v2->px, seg->v2->py);
```
kraflab, line 69: This definitely only affects rendering right?

bkoropoff: Yeah, it gets used in one place in r_segs.c for rendering

kraflab:
👍

kraflab:
Oops, this does break hexen 🙂

PR #386, opened 2023-05-26
GL: round weapon coordinates

bkoropoff:
Avoids a 1-pixel gap at the bottom of the weapon sprite in some resolutions.  A previous attempt to fix this did not work.

----

## Before
doom00

## After
doom01


kraflab:
👍

PR #387, opened 2023-05-26
GL: draw weapon sprite at full resolution

bkoropoff:
Avoids bad aliasing in some resolutions

-----

## Before

doom01

## After

doom00

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -1048,10 +1048,6 @@ void gld_EndDrawScene(void)
   glViewport(0, 0, SCREENWIDTH, SCREENHEIGHT);
   gld_Set2DMode();
 
-  glsl_PushMainShader();
-  R_DrawPlayerSprites();
-  glsl_PopMainShader();
-
   // e6y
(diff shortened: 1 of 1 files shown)


kraflab:
Why is it the weapon sprite in particular?

bkoropoff:
The weapon sprite is prominent, always drawn at the same scale, is fairly static, and was rendered with no scaling in the original resolution, so any aliasing particularly sticks out.

Lovey847:
From what I can see, this is drawing the weapon sprite on top of the render texture. What worries me about this approach is the weapon sprite crossing pixel boundaries, which you can see happening in the screenshot.

Top left half: Render texture
Bottom right half: Weapon sprite
Red lines: Where the render texture pixels don't line up with the weapon sprite
weaponTexturePixels

bkoropoff:
Yeah, the weapon sprite won't use the same pixel grid as the scene, just like the UI/HUD elements don't.  I think that's preferable to aliasing.

kraflab:
Thanks 👍

PR #388, opened 2023-05-26
GL: Add dithered fade mode

bkoropoff:
This can wait until 0.27 since 0.26 is in RC.

# Modes

- Smooth mode performs linear interpolation between colormap entries based on light level, removing light banding
- Dithered mode applies Bayer 8x8 dithering to selecting colormap entries, which gives a more retro look

## Normal

doom00

## Smooth

doom01

## Dithered

doom02

# Filtering

The second commit adds a "sharp bilinear" shader for render scaling.  This reduces aliasing from non-integer scales (particularly noticeable on dither patterns) while retaining sharp pixels.  It's equivalent to nearest-neighbor scaling by the greatest integer less than the scale factor, followed by bilinear interpolation for the remaining fractional scale.

## Nearest neighbor

This dither pattern is uneven due to nearest-neighbor aliasing.

doom02

## Sharp bilinear

This makes the pattern even while retaining as much sharpness as possible by only blending at the transition between source texels.

doom04

## Bilinear

For reference, this is what normal bilinear filtering does to the pattern.

foo

### DIFF
--- a/prboom2/src/gl_main.c
+++ b/prboom2/src/gl_main.c
@@ -1064,6 +1064,7 @@ void gld_EndDrawScene(void)
     GLEXT_glBindFramebufferEXT(GL_FRAMEBUFFER_EXT, 0);
 
     glBindTexture(GL_TEXTURE_2D, glSceneImageTextureFBOTexID);
+    glsl_SetTextureDims(0, SCREENWIDTH, SCREENHEIGHT);
 
     // Setup blender
     glColor3f(1.0f, 1.0f, 1.0f);
@@ -1073,6 +1074,7 @@ void gld_EndDrawScene(void)
     dsda_GLSetRenderViewport();
     // elim - Prevent undrawn parts of game scene texture being rendered into the viewport
     dsda_GLSetRenderSceneScissor();
+    glsl_PushSharpBilinearShader();
     glBegin(GL_TRIANGLE_STRIP);
     {
       glTexCoord2f(0.0f, 1.0f); glVertex2f(0.0f, 0.0f);
@@ -1084,6 +1086,8 @@ void gld_EndDrawScene(void)
 
     // elim - Set the scissor back to the full viewport so post-scene draws can happen (ie StatusBar)
     dsda_GLSetRenderViewportScissor();
+
+    glsl_PopSharpBilinearShader();
 
     gld_Set2DMode();
 
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -2958,7 +2958,7 @@ static const char* fake_contrast_list[] =
   NULL
 };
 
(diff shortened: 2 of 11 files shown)


kraflab:
I think these are fun but I just got through removing light modes, I'm not about to add more. But maybe I'll change my mind by the time 0.27 comes around.

bkoropoff:
> I think these are fun but I just got through removing light modes, I'm not about to add more. But maybe I'll change my mind by the time 0.27 comes around.

These are just variants of indexed (they're part of the same shader), so the code footprint is very small compared to the old lighting modes.

bkoropoff:
I refactored shader loading to support `#include` directives (there is a Google extension for this which shader validation tools understand) and passing `#define`s.  This makes shader code organization/reuse a lot easier.

prboom2/src/gl_main.c
```diff
@@ -1083,6 +1084,8 @@ void gld_EndDrawScene(void)
     // elim - Set the scissor back to the full viewport so post-scene draws can happen (ie StatusBar)
     dsda_GLSetRenderViewportScissor();
 
+    glsl_SetActiveShader(NULL);
```
kraflab: What do you think about migrating to a shader stack before doing this PR? There have been some bugs caused by nested logic and it might be good to clean that up before adding more.

kraflab:
Could you make a separate PR with just the refactoring but no feature changes? Then we can get that merged independently

bkoropoff:
This is rebased on #400, waiting for that first

prboom2/src/m_menu.c
```diff
@@ -2959,6 +2961,8 @@ setup_menu_t audiovideo_settings[] = {
   { "Vertical Sync", S_YESNO, m_conf, G_X, dsda_config_render_vsync },
   { "Uncapped Framerate", S_YESNO, m_conf, G_X, dsda_config_uncapped_framerate },
   { "FPS Limit", S_NUM, m_conf, G_X, dsda_config_fps_limit },
+  { "GL Lighting", S_CHOICE, m_conf, G_X, dsda_config_gl_light_mode, 0, gl_light_mode_list },
```
kraflab: Maybe "GL Light Banding" and `gl_light_banding_mode`

bkoropoff: Light fade mode?  Light drop-off mode?  Light quantization mode?

bkoropoff: "GL Distant Light Fade"?

kraflab: GL Light Fade sounds good to me

kraflab:
Can you split this PR into the fade one and the bilinear scaling one? I think we can merge the fade stuff.

kraflab:
Closing this one since I don't expect to merge it and it's out of date by this point. The smooth fade already merged is quite nice but I'm not convinced we need the dither - it's here if we want to reconsider it again later on.

PR #392, opened 2023-05-28
Add fake_contrast 2 setting

bkoropoff:
This forces smooth fake contrast except for walls that explicitly have it turned off.

----

Fake contrast can look a bit odd in some places because it only applies to exactly axis-aligned walls and not walls that are slightly off-axis.  This adds an extra configuration value to use the smooth fake contrast calculation even for older maps that don't explicitly enable it on some walls.

### DIFF
--- a/prboom2/src/dsda/configuration.c
+++ b/prboom2/src/dsda/configuration.c
@@ -25,6 +25,7 @@
 #include "gl_struct.h"
 #include "lprintf.h"
 #include "r_main.h"
+#include "r_segs.h"
 #include "s_sound.h"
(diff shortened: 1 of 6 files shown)


kraflab:
Could you rename this to `fake_contrast_mode` and add an enum so we can do `fake_contrast_mode != no_fake_contrast` and `fake_constrast_mode == smooth_fake_contrast` for the checks?

bkoropoff:
Done

bkoropoff:
I added a menu option so people can turn it on more easily in the alpha.  Let me know if you don't want it.

PR #393, opened 2023-06-02
GL: remove dead fog code

bkoropoff:
This is unused, and wouldn't work anyway because we use fragment shaders now


kraflab:
Thanks 👍

PR #394, opened 2023-06-03
build: accept mad.dll for libmad

bkoropoff:
Something changed in vcpkg and libmad is now `mad.dll` instead of `libmad.dll`

This won't fix github CI, only local builds.


bkoropoff:
Actually, I'm still not sure what's going on with my Windows build...

PR #401, opened 2023-06-11
Shaders: handle texture dimension uniforms generically

bkoropoff:
Since GLSL 1.10 doesn't let you query the size of a texture, we have to pass texture dimensions manually.  Make doing this generic and automatic, and use it for the fuzz shader.

### DIFF
--- a/prboom2/data/lumps/gls_fuzz.lmp
+++ b/prboom2/data/lumps/gls_fuzz.lmp
@@ -16,13 +16,13 @@ uniform sampler2D tex;
 uniform sampler2D fuzz;
 // Texture dimensions
 uniform vec2 tex_d;
+// Fuzz texture dimensions
+uniform vec2 fuzz_d;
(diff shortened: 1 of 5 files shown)


kraflab:
Is there a version that does let you query the size that would be easier to use?

bkoropoff:
GLSL 1.30 has `textureSize`, but it requires OpenGL 3.0

kraflab:
What kind of benefits would be get from requiring 3.0?

bkoropoff:
3.0 is generally more modern and complete, like support for integers throughout the entire stack (e.g. use non-normalized integers for indexed textures, use non-normalized integer coordinates to do colormap lookups), an overhaul of how shader inputs/outputs are specified to be more flexible and explicit, etc.  It has vertex array objects, which might speed up rendering.

kraflab:
If you think it's worth migrating at some point, feel free to mess around with it

PR #402, opened 2023-06-11
Allow tweaking fluidsynth chorus/reverb settings

bkoropoff:


kraflab:
@ceski-1 I think you mentioned you were interested in upgrading the fluidsynth code this year. Is that something you're going to look into? Just asking to possibly avoid duplicate labor on this topic 😄

ceski-1:
@kraflab Maybe some day, but these options don't hurt anything.

PR #404, opened 2023-06-12
Render at full window/screen resolution in HiDPI

bkoropoff:
If an application doesn't explicitly indicate HiDPI support, most OSes/desktops report a lower than actual window resolution and automatically scale the application's output up. dsda-doom handles its own scaling with GL, so use the full available resolution.


kraflab:
Are any changes needed on the software side?

Issue #283, opened 2023-03-25
Consider following XDG base directory specification on Linux


bkoropoff:
Putting everything in `~/.<appname>` has been considered bad practice on Linux for a while.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

- Configuration files usually go in ~/.config/<appname>
- Cache files (anything it's safe to delete) usually go in ~/.cache/<appname>
- Data files (e.g. saves) usually go in ~/.local/share/<appname>

This keeps the home directory clean, lets backup apps only back up important files, lets the cache directory be cleared to reclaim space, etc.

kraflab:
My main concern about changing anything like this is if it will cause problems for existing users

bkoropoff:
The code should detect an existing `~/.dsda-doom` directory and continue using it for config and save games if it's present.  That's how a lot of other Linux apps migrated to the XDG spec.

facespkz:
I remember submitting a patch for this to PrBoom+, only problem is the logic prioritized the path under home directory rather than the default.

ndfsa:
If I'm not mistaken, PrBoom+ has already implemented this in https://github.com/coelckers/prboom-plus/pull/470

kraflab:
https://github.com/kraflab/dsda-doom/commit/38fc904d32f434a8ead3d8c815753fa13e9cdb70

PR #412, opened 2023-07-20
[MacOS] Bug: Window unfocused using leftclick in certain conditions

sbritorodr:
## Before merging
Before merging, I do recommend testing this on Linux and Windows. I'm away of my main setup so I can't test this.

## Description of this bug:
In MacOS, if you pause the game and unpause it, there is a chance the game will unfocus and ungrab the mouse, loosing the player controls until you focus the window again. 

## Steps to reproduce
1. Start a new game in any IWAD, map and skill level
2. Pause the game
3. Move the mouse outside the window
4. Unpause the game
5. Left-click
6. Bug happens, and it's very annoying

## Hardware Tested
MacOS 13.4.1

## Fixes:
Now when mouse is grabbed, it will be moved inside the screen.


sbritorodr:
As you can see on my first commit, II want to move it to the center of the screen instead of top-left (just beacuse it's better to see the cursor), but I want to be as concise and fast as possible and not doing unnecesary divisions each time the function is called. So I want to know if there is a global screen size variables (who works both in software and gl mode).

I don't mind leaving this "as it is", but if you want to make it centered, please let me know those variables (or where is the center of the screen stored, because it should be stored elsewhere for weapon aiming)

sbritorodr:
Finally found both variables and used a bit-shift instead of a division.

kraflab:
> While the mouse is in relative mode, the cursor is hidden, the mouse position is constrained to the window, and SDL will report continuous relative mouse motion even if the mouse is at the edge of the window.

Any idea why `SDL_SetRelativeMouseMode(SDL_TRUE);` does not work? I can't reproduce this issue on windows.

sbritorodr:
No idea, but for sure this PR fixed the problem on my macos machine. I'm going to try something to see if this spits any SDL_Error or something. I'm working on fixing the xdg issue #283 [right now](https://github.com/sbritorodr/dsda-doom/tree/xdg) so I will work on why `SDL_SetRelativeMouseMode(SDL_TRUE);` doesnt work.

I'm not an expert with SDL, so I'm gonna investigate.

sbritorodr:
`SDL_SetRelativeMouseMode(SDL_TRUE);` does not return any error codes for me and according to [SDL docs](https://wiki.libsdl.org/SDL2/SDL_SetRelativeMouseMode) it's working as expected.

SDL_SetRelativeMouseMode(SDL_TRUE) will:
1. Hide the cursor
2. Move the game camera regardless if the mouse is inside the window or not.
3. Don't move the cursor (I think so?)

The problem is MacOS still detects the mouse is outside the window screen. So if you are in windowed mode and unpause the game while the cursor is not inside the game window it will keep the cursor outside the window and when you press left-click it will lost focus and controls.

Setting it to false the expected: Cursor not hidden, camera only moves if the cursor is inside the game screen.

Maybe it's a problem on how Unix-like or just MacOS handles focus. Have you tested this on linux as well?

If you need video proof just let me know. 
Cheers

kraflab:
`the mouse position is constrained to the window` is called into question here, so I'd like to know why. If the fix for a problem doesn't make any sense, it often points to a different problem and the fix is rather affecting a symptom and not solving the underlying issue. Or this line is misleading, in which case I'd like to see a source explaining why so we can learn from it.

kraflab:
@Pedro-Beirao does this issue also affect you?

Pedro-Beirao:
Yes it does. And I would also like to add that on fullscreen, the mouse is not shown when the game is paused

sbritorodr:
@Pedro-Beirao good point. However, I have no clue how to fix it. I do accept PR if you want to add a small change.

sbritorodr:
> Or this line is misleading, in which case I'd like to see a source explaining why so we can learn from it.

I think the line could be interpreted in various ways. Here is my interpretation:
I do use Blender sometimes and "constraining" usually means in 3D software that the movement and rotation of one object is directly copied to the constrained object. That mean exactly as expected, the mouse movement is copied to the camera movement and the mouse position (aka. cursor) is inside the boundaries of the window zone. This means that if I do a full 360º turn inside the game the mouse would not go away the window boundaries.

The problem is not related to this functionality, in fact, this works as intended in MacOS. Is when you disable the function (pause the game), move the cursor away the window boundaries and then activate the function (unpause) when the cursor is hidden but outside the window, so pressing leftclick unfcus the window.

It is important to notice that the mouse **position** is constrained to the window, not moving the cursor inside the window. Maybe `SDL_SetWindowGrab` gets a better job on doing this.

Sometimes when two completely different systems with completely different window system protocols have different behaviours. Or it's just a SDL macos speciffic bug and should be solved pening a ticket to SDL.

 I haven't found any other sources explaining this speciffic function. All data I found copies SDL documentation.

sbritorodr:
TL;DR: This SDL works as intended if the mouse is already inside the window in MacOS when is set to true. Somehow the function doesn't appear to work if it's set to true but the cursor is outside the window in the first place.

sbritorodr:
I was testing this by calling `SDL_GetError()` and it show me this error whenever I unpause the game:
`Unknown touch device id -1, cannot reset`
And at game startup:
`Invalid window`

Does it happen to you as well if you add `printf("%s\n", SDL_GetError());` below `SDL_GetRelativeMouseMode()`?

Captura de pantalla 2023-07-23 a las 22 14 55

sbritorodr:
Okay. I'm wrong. I've read the [SDL source code](https://github.com/libsdl-org/SDL/blob/8adab0b88ab1c0aa69242fffc48afdba085b1dc4/src/events/SDL_mouse.c#L1022) and `SDL_SetRelativeMouseMode()` does call the same function I've added, `SDL_WarpMouseInWindow()`. 

The problem is I don't know why the warp mouse function does not work when it's called with `SDL_SetRelativeMouseMode()` but it works perfectly when I added manually the function.

I'm afraid it's more f a SDL2 issue with MacOS more than dsda-doom fault, but when I called SDL_GetError() it dropped two errores: `Invalid Window` (start) and `Unknown touch device id -1, cannot reset` (pause-unpause).
I'm thinking about opening a ticket on SDL directly.

kraflab:
Your code passes `sdl_window` directly, which may be the distinction. Maybe the existing window is invalid for some reason (guessing based on the error) and sending this parameter explicitly fixes that. That might mean we are initializing something wrong elsewhere.

sbritorodr:
Does this happen in other platforms as well?

andrey-budko:
Same in quakespasm
https://github.com/sezero/quakespasm/pull/49

sbritorodr:
Made some changes just so it's exclusively to MacOS since `SDL_SetRelativeMouseMode` already does put the mouse in the center.

We could open a ticket to the SDL devs and link the proyect, maybe it's just a bug and need to be solved by the SDL side.

kraflab:
I haven't dug into this invalid window question, so I would hold off on opening a ticket elsewhere for now.

sbritorodr:
Hello, I've updated my local repository and then build again the program in release mde. This time I had a linker error at the end but it disappeared when I specified "Ninja" as a build system generator and then ran Ninja (github actions script is programmed this way for some reason).

The bug also disappeared, so I'm closing this pr as the bug is solved too. Thanks for the help.

PR #418, opened 2023-08-12
Add show WPN argument to weapon_text HUD component

unectomy:
This adds a new argument to the `weapon_text` HUD component. It lets the player toggle whether the WPN text of the component shows up.

image
image
image
image

### DIFF
--- a/prboom2/src/dsda/hud_components/weapon_text.c
+++ b/prboom2/src/dsda/hud_components/weapon_text.c
@@ -22,6 +22,7 @@
 typedef struct {
   dsda_text_t component;
   dboolean grid;
+  dboolean show_wpn;
 } local_component_t;
 
 static local_component_t* local;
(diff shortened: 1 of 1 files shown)


prboom2/src/dsda/hud_components/weapon_text.c
```diff
@@ -31,57 +32,106 @@ static void dsda_UpdateComponentText(char* str, size_t max_size) {
...
-      player->weaponowned[6] ? '7' : ' ',
-      player->weaponowned[7] ? '8' : ' ',
-      player->weaponowned[8] ? '9' : ' '
-    );
+  if (local->show_wpn) {
+    if (local->grid)
+      snprintf(
+        str,
+        max_size,
+        "%sW %s%c %s%c %c\n"
```
kraflab: There are some options here to make it cleaner.

We could have a `const char* format;` in the component struct, define 4 static format strings, and set the `format` when parsing the args in the init function. Then you won't need the `show_wpn` attribute and also don't need to duplicate the interpolation calls. The downside is the format string is no longer visible directly where we are interpolating.

Another option is to set `label_w`, `label_p`, `label_n`, and `label_wpn` when initializing (either being, e.g., `"W "` or `""`) and interpolating those labels into the existing format strings (see `composite_time.c`). That has the benefit of keeping the format strings visible in the print functions, but has the disadvantage of adding even more interpolations to them.

I would probably go for the second option.

unectomy:
I implemented the second option.

prboom2/src/dsda/hud_components/weapon_text.c
```diff
@@ -82,6 +87,23 @@ void dsda_InitWeaponTextHC(int x_offset, int y_offset, int vpt, int* args, int a
   local = *data;
 
   local->grid = args[0];
+  if (arg_count > 1)
+    local->show_wpn = args[1];
+  else
+    local->show_wpn = true;
+
+  if (local->show_wpn) {
+    snprintf(local->label_w, sizeof(local->label_w), "%sW ", dsda_TextColor(dsda_tc_exhud_weapon_label));
```
kraflab, line 97: I'm not sure if it's currently possible, but it might be possible at some point to change the colors without reinitializing the hud, so we can't bake them in at this point.

prboom2/src/dsda/hud_components/weapon_text.c
```diff
@@ -82,6 +87,23 @@ void dsda_InitWeaponTextHC(int x_offset, int y_offset, int vpt, int* args, int a
   local = *data;
 
   local->grid = args[0];
+  if (arg_count > 1)
+    local->show_wpn = args[1];
+  else
+    local->show_wpn = true;
```
kraflab: We don't need to store this flag in the component data.

unectomy:
Regarding your second review, I moved the flag to a local variable. For the first review, is there a change I can make to support its feature? I'm not sure how to do it.

kraflab:
I wrapped it up here: https://github.com/kraflab/dsda-doom/commit/3a6cba5d4b10df758ccb1403c3d384a7cf70f73e
Thanks!

PR #419, opened 2023-08-13
Add color coding for big_ammo HUD component

unectomy:
This adds color coding to the `big_ammo` HUD component, where the color of the component depends on how much ammo the player has, similar to health and armor HUD components.

image
image
image
image

### DIFF
--- a/prboom2/src/dsda/hud_components/big_ammo.c
+++ b/prboom2/src/dsda/hud_components/big_ammo.c
@@ -31,6 +31,9 @@ static void dsda_DrawComponent(void) {
   player_t* player;
   ammotype_t ammo_type;
   int ammo;
+  int cm;
+  int ammopct = 0;
+  int max_ammo;
 
   if (hexen)
     return;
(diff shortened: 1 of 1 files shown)


prboom2/src/dsda/hud_components/big_ammo.c
```diff
@@ -42,9 +45,24 @@ static void dsda_DrawComponent(void) {
     return;
 
   ammo = player->ammo[ammo_type];
+  max_ammo = player->maxammo[weaponinfo[player->readyweapon].ammo];
+
+  if (ammo == max_ammo)
+    cm = CR_LIGHTBLUE;
+  else {
+    if (max_ammo)
```
kraflab, line 53: If you check above, you'll see that we don't reach this point in the code if there is no max ammo.

prboom2/src/dsda/hud_components/big_ammo.c
```diff
@@ -42,9 +45,24 @@ static void dsda_DrawComponent(void) {
     return;
 
   ammo = player->ammo[ammo_type];
+  max_ammo = player->maxammo[weaponinfo[player->readyweapon].ammo];
```
kraflab, line 48: We can just get the ammo percent directly and use it below, `ammo_percent = ammo * 100 / player->maxammo[ammo_type]`

prboom2/src/dsda/hud_components/big_ammo.c
```diff
@@ -42,9 +45,24 @@ static void dsda_DrawComponent(void) {
...
+
+  if (ammo == max_ammo)
```
kraflab, line 50: Take a look at `big_health.c` to get an idea of how I would format this section.

kraflab (commented):
I left some comments but there's a bigger topic, which is that all of this needs to be extracted so it can also be used by the small text component, and we need to extract all the colors into the configurable text color system. This also needs to happen for the health component and probably other things. I'm going to take care of that some time soon, so I would let go of this PR for now.

PR #429, opened 2023-10-29
OPL updates

rfomin:
Unfinished work on OPL3 integration.

Is this PR the way I should work on it?


kraflab:
Yes this is the way.

rfomin:
All updates from Chocolate Doom are in, OPL3 mode is enabled by default.

Maybe we should rename the OPL2 player menu item to just OPL.

prboom2/src/dsda/configuration.c
```diff
@@ -756,6 +756,10 @@ dsda_config_t dsda_config[dsda_config_count] = {
     "mus_opl_gain", dsda_config_mus_opl_gain,
     dsda_config_int, 0, 1000, { 50 }
   },
+  [dsda_config_mus_opl_opl3mode] = {
+    "mus_opl_opl3mode", dsda_config_mus_opl_opl3mode,
```
kraflab, line 760: Why would someone want to disable this? What is the difference exactly?

kraflab:
> Maybe we should rename the OPL2 player menu item to just OPL.

Yes I think that makes sense.

prboom2/src/dsda/configuration.c
rfomin: OPL2 sounds worse and only mono. But most of the popular sound cards in the early 90s were OPL2.

kraflab: I see, so it's there for people who want the crustiest option :^)

kraflab:
Anything else needed or is it good to go?

rfomin:
> Anything else needed or is it good to go?

I think it's done.

kraflab:
Thanks!

PR #438, opened 2023-11-05
allows dehacked custom quit messages

T-117:
Allows setting custom quit messages though the normal dehacked processes.

### DIFF
--- a/prboom2/src/dstrings.c
+++ b/prboom2/src/dstrings.c
@@ -33,29 +33,30 @@
  */
 
 #include "dstrings.h"
+#include "d_deh.h"
 
 
 // killough 1/18/98: remove hardcoded limit, add const:
-const char *const endmsg[]=
+const char** endmsg[]=
 {
   // DOOM1
-  QUITMSG,
-  QUITMSG1,  // "please don't leave, there's more\ndemons to toast!",
-  QUITMSG2,  // "let's beat it -- this is turning\ninto a bloodbath!",
-  QUITMSG3,  // "i wouldn't leave if i were you.\ndos is much worse.",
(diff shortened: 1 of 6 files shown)


kraflab:
Do you have a link to a dehacked specification that includes this?

T-117:
I don't know about specifications other than maybe this https://zdoom.org/wiki/Strings
but I noticed playing in Woof that the quit message strings were supported but not so in dsda.  I didn't know if that was intentional but if not I figured I'd add the feature here.

kraflab:
Did you test it? It doesn't look like the dehacked strings are actually implemented.

prboom2/src/dstrings.c
```diff
@@ -40,22 +40,22 @@ const char *const endmsg[]=
...
   // DOOM1
   QUITMSG,
-  "please don't leave, there's more\ndemons to toast!",
-  "let's beat it -- this is turning\ninto a bloodbath!",
-  "i wouldn't leave if i were you.\ndos is much worse.",
-  "you're trying to say you like dos\nbetter than me, right?",
-  "don't leave yet -- there's a\ndemon around that corner!",
-  "ya know, next time you come in here\ni'm gonna toast ya.",
-  "go ahead and leave. see if i care.",  // 1/15/98 killough
+  QUITMSG1,  // "please don't leave, there's more\ndemons to toast!",
```
fabiangreffrath: Shouldn't you use the `s_*` substitutes here instead?

T-117: good call.  Fixed.

T-117:
@kraflab The most recent changes are tested and work as expected.

kraflab:
👍

PR #441, opened 2023-11-10
Update HUD doc with show_labels for the stat_totals component

unectomy:
This adds missing documentation to the HUD doc. Before, it was missing the `show_labels` argument for the `stat_totals` HUD component.


kraflab:
Oops, nice catch 👍

PR #442, opened 2023-11-12
Add an option to enable quit sounds

FozzeY:
If ENDOOM is enabled, play a random sound on confirming a quit message as per vanilla behaviour.


kraflab:
Why is this coupled to endoom being enabled? I realize it was in prboom+, but why?

FozzeY:
A way to preserve the original behaviour in case the lack of quit sounds in dsda-doom was intentional. Should this be its own config option or just enabled period?

kraflab:
Yes I mean that "quit sounds" should probably be an independent option if we want to bring it back.

kraflab:
Thanks 👍

Issue #425, opened 2023-09-25
Use ~/.config/dsda-doom instead of ~/.dsda-doom on linux


leia-uwu:
i want my home directory to be clean :p

Pedro-Beirao:
You can use ~/.dsda-doom, /usr/local/share/doom or any other folder with DOOMWADDIR

fabiangreffrath:
This is about the config file path, I guess.

leia-uwu:
yeah its about the config path

fabiangreffrath:
There was code in prb+ for that.

kraflab:
What about the people that currently use the port and have their config under ~/.dsda-doom?

fabiangreffrath:
We could attempt to read the config from the legacy directory first, and delete it afterwards, but always save it to the new location.

fabiangreffrath:
No wait. Try to read config from new location, if not found read from legacy location, always save to new location.

leia-uwu:
hmm, just move old config to new config location if the old config directory exists and the new config directory doesn't?
i tried doing that myself but i am bad with C

mooreye:
What about data files like autoload and IWADs? They would belong in `~/.local/share/dsda-doom` since they are user data, not configs?

wongandj:
Piping in to say I still think this would be a good idea. Most unix (i.e. linux and macos) have done this change by reading and updating to the old `$HOME` files if they are already there, else R/Wing to `$XDG_*` if they are defined, else R/Wing to their default values (`$HOME/.config/dsda-doom`, `$HOME/.local/share/dsda-doom`, etc.)

Configs should go to `$XDG_CONFIG_HOME/dsda-doom`, app data (saves, tranmaps) should go to `$XDG_DATA_HOME/dsda-doom`.

I think that the user data like iwad search dir, etc. should continue to search in just current dir, then exe dir, then `$DOOMWADDIR` if it is defined, etc., since that is the expected behavior for doom ports. It would be better to define autoloads in the config rather than in a set directory, like gzdoom does, though an enhancement would be to allow users to specify autoloads with shell expansion, e.g. `"~/doom/autoloads/all-all/*"` instead of having to list each file to load verbatim like in GZDoom; maybe a colon-separated list of paths?

It would also be good to allow users to specify where save, etc. files should go in the config, like they can currently do with screenshots, which would override the previously mentioned search behavior.

kraflab:
[...]

PR #443, opened 2023-11-12
Search for files in XDG_DATA_HOME and XDG_DATA_DIRS

DankRank:
In particular, WADs can now be placed in `~/.local/share/games/doom`.

---

Unrelated, but #425 will have to do something similar with XDG_CONFIG_HOME and XDG_CONFIG_DIRS for config files.

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -433,6 +433,31 @@ static const char *I_GetBasePath(void)
 
 // Reference for XDG directories:
 // <https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html>
+static const char *I_GetXDGDataHome(void)
+{
+  static char *datahome = 0;
+
+  if (!datahome)
+  {
+    const char *xdgdatahome = M_getenv("XDG_DATA_HOME");
+
+    if (!xdgdatahome || !*xdgdatahome)
+    {
+      const char *home = M_getenv("HOME");
+
+      if (!home)
+        home = "/";
+      datahome = Z_Malloc(strlen(home) + 1 + sizeof(".local/share"));
+      sprintf(datahome, "%s%s%s", home, !HasTrailingSlash(home) ? "/" : "", ".local/share");
+    }
+    else
+    {
+      datahome = Z_Strdup(xdgdatahome);
+    }
+  }
+  return datahome;
+}
+
 static const char *I_GetXDGDataDirs(void)
 {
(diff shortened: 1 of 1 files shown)


Pedro-Beirao:
This has been address by some other PR right? If theres any part of it that hasnt been addressed, we can re open and I'll take a look.

Its an old PR, so I wanted close it.

DankRank:
XDG_DATA_HOME was addressed in #491, so that commit is no longer necessary. The other two are still relevant. If you want, I can rebase it for your convenience.

DankRank:
Actually, nevermind, all three commits are still relevant. That PR was about configs (and wads) in `$XDG_DATA_HOME/dsda-doom`, while this one is about wads in `$XDG_DATA_HOME/games/doom`.

I also verified that this PR merges cleanly and correctly onto the current master branch, so it's ready for review.

prboom2/src/SDL/i_system.c
```diff
@@ -354,11 +354,47 @@ static const char *I_GetBasePath(void)
...
+#define PATH_SEPARATOR ":"
 #endif
 
+// Reference for XDG directories:
+// <https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html>
+static const char *I_GetXDGDataHome(void)
+{
+  static char *datahome = 0;
+
+  if (!datahome)
```
Pedro-Beirao: No need for this `if` right?

Pedro-Beirao:
I understand that it doesnt screw up anything, but Id prefer if al XDG related code was behind a `#ifdef __LINUX__`. Its easier to read.

prboom2/src/SDL/i_system.c
DankRank: It's necessary. `datahome` is a `static` variable, which means it's a global variable that's scoped to the function, and only initialized once at the start of the program. The effect of this is that the XDG_DATA_HOME path is only computed once. There's similar code in implementations of `I_ExeDir`.

DankRank:
> I understand that it doesnt screw up anything, but Id prefer if al XDG related code was behind a `#ifdef __LINUX__`. Its easier to read.

XDG is most certainly *not* linux-specific. It should be used on any operating system that has a /usr tree.

I'll add a `#ifndef _WIN32`, but note that `/usr/` search paths were not originally behind an ifdef.

DankRank:
rebased and added 3 more commits

Pedro-Beirao (approved):
Looks good. Thank you!

PR #444, opened 2023-11-14
Fix renderer capture

### LINKED ISSUE #437: -viddump outputs outdated frames on Linux
Video output from -viddump looks jittery when using OpenGL on X11 (both in OpenGL mode and in software mode as a SDL_Renderer backend). More specifically, a frame of video will occasionally be replaced with an older frame. The reason for this is that I_CaptureFrame is called after presenting the frame.

https://github.com/kraflab/dsda-doom/assets/4050457/a69efb9c-b61b-40e4-bf8d-8d686a96b3aa

After a call to SDL_GL_SwapWindow (or SDL_RenderPresent), the contents of the back buffer become [undefined](https://gamedev.stackexchange.com/questions/7316/what-is-the-contents-of-the-buffer-after-a-call-to-glswapbuffers). This means that to consistently capture the rendered pixels, glReadPixels (or SDL_RenderReadPixels) must be called before SDL_GL_SwapWindow. In practice, the back buffer either stays the same (i.e. it doesn't swap), or it receives the previous front buffer (i.e. it swaps, which is the usual behavior), but crucially this behavior can change from frame to frame. When it does, the dumped video either skips a frame, or displays the previously skipped frame. This also makes it go in an out of sync with the audio by one frame.

To fix this bug, calls to SDL_RenderPresent (in I_FinishUpdate) and SDL_GL_SwapWindow (in gld_Finish) must be pulled out and placed after any potential I_CaptureFrame calls. For correctness, M_ScreenShot should also only be called right before swapping.

<details>
<summary>Extra reproducibility info</summary>

It doesn't happen if I run it with `SDL_RENDER_DRIVER=software` in software rendering mode. Also I haven't been able to reproduce it on wayland. Also my GPU is AMD.

I've been trying to figure out what causes GL to choose to swap or not to swap, and it seems it's performance-dependent, but it can be tricky to reproduce. Sometimes the jitters are almost constant, sometimes they only happen once every few seconds. The first frame of the video is typically entirely black, but not always. It seems to be easier to get this bug in software mode than in OpenGL mode. 

Reproducibility doesn't really matter in this case though. Since back buffer is formally considered undefined after swapping, you can put SDL_RenderClear/glClear right after SDL_RenderPresent/SDL_GL_SwapWindow, and if something breaks of it (as -viddump and screenshots do), that's a bug in and of itself.
</details>

kraflab:
Fixes https://github.com/kraflab/dsda-doom/issues/437

### DIFF
--- a/prboom2/src/SDL/i_video.c
+++ b/prboom2/src/SDL/i_video.c
@@ -535,6 +535,34 @@ void I_ShutdownGraphics(void)
   DeactivateMouse();
 }
 
+static dboolean queue_frame_capture;
+static dboolean queue_screenshot;
+
+void I_QueueFrameCapture(void)
+{
+  queue_frame_capture = true;
+}
+
+void I_QueueScreenshot(void)
+{
+  queue_screenshot = true;
+}
+
+void I_HandleCapture(void)
+{
+  if (queue_frame_capture)
+  {
+    I_CaptureFrame();
+    queue_frame_capture = false;
+  }
+
+  if (queue_screenshot)
+  {
+    M_ScreenShot();
+    queue_screenshot = false;
+  }
+}
+
 //
 // I_FinishUpdate
 //
@@ -599,6 +627,8 @@ void I_FinishUpdate (void)
   SDL_RenderClear(sdl_renderer);
 
   SDL_RenderCopy(sdl_renderer, sdl_texture, &src_rect, NULL);
+
+  I_HandleCapture();
 
   // Draw!
   SDL_RenderPresent(sdl_renderer);
--- a/prboom2/src/d_event.h
+++ b/prboom2/src/d_event.h
@@ -82,7 +82,6 @@ typedef enum
   ga_completed,
   ga_victory,
   ga_worlddone,
-  ga_screenshot,
 
   // hexen
   ga_leavemap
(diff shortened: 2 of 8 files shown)


kraflab:
@dankrank could you try this out?

DankRank:
Yep, this fixes the jitter.

The first frame is still black though. I think what's happening is:
- D_DoomLoop calls I_QueueFrameCapture followed by D_Display
- D_Display realizes it needs to do a wipe and calls D_Wipe
- D_Wipe checks cap_wipescreen, sees that it zero, and *doesn't* call I_QueueFrameCapture
- But because frame capture was queued in D_DoomLoop, the first frame of wipe (completely black) is captured anyway

In other words, D_Wipe needs to unqueue the frame, when the capture check fails. I'm running this with `render_wipescreen 1` and `cap_wipescreen 0`, by the way (which is the default).

kraflab:
> The first frame is still black though.

I added an extra check for the wipe state, let me know if that fixes it. If you wouldn't mind checking if cap_wipescreen 1 also works I would appreciate it 🙂

DankRank:
I've tested with every combination of videomode and cap_wipescreen, now it works flawlessly. Thanks!

kraflab:
Thanks!

PR #446, opened 2023-11-18
Update .gitignore

jengelh:


kraflab:
The .gitignore already covers multiple file structure options. If you don't like what's there you can make a local .gitignore to cover the difference, but I don't think we need to keep adding more to this file.

jengelh:
>The .gitignore already covers multiple file structure options

Whatever it already has is not enough to cover cmake artifacts. And cmake is the build system you prescribe with dsda-doom… so please.

Pedro-Beirao:
cmake and make should be ran inside a build folder, not from /prboom2
That's why build/* is in the .gitignore

kraflab:
Closing in favor of the linked PR.

PR #447, opened 2023-11-18
build: enforce out-of-tree builds

jengelh:
https://github.com/kraflab/dsda-doom/pull/446#issuecomment-1817661300

>cmake and make should be ran inside a build folder, not from /prboom2

### DIFF
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -1,5 +1,18 @@
 cmake_minimum_required(VERSION 3.9)
 
+function(AssureOutOfSourceBuilds)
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
We have a much shorter variant of this in Woof:

https://github.com/fabiangreffrath/woof/blob/0f49f788bdad5a66a6290c14e6eb9c4f4c0956c0/CMakeLists.txt#L32-L35

kraflab:
Please adopt Woof's implementation if it also works.

kraflab:
Closing this as it's now nearly 4 months old, but feel free to revive it in the future.

PR #450, opened 2023-11-26
Fixed level name capitalization.

spiral6:
Doom 2 and Final Doom's level names are not capitalized which looks inconsistent with Doom 1's levels being capitalized. While in-game this doesn't affect too much as all strings are set to capital-only with the in-game font, this will help with a different feature I'm testing later on.


kraflab:
I'd rather not mess with these strings. Since capitalization is often not done (and this affects many many many strings even beyond these), any feature will need to be compatible with uncapitalized map titles (e.g., this problem will also exist for tons of pwads).

spiral6:
I will work around these then by adding a separate header file in the folder for the feature I'm working on that will duplicate these definitions.

spiral6:
As far as I can tell though, PWADs do not have this issue (from the ones I've tested, i.e. BTSX, Eviternity, SIGIL) and do not follow the same rules. This is strictly an oddity with Doom 2/Final Doom. Doom 1's level names are already capitalized.

PR #457, opened 2023-12-15
Load and mostly implement ACS on ZDoom format maps

spazzylemons:
This change loads BEHAVIOR lumps, if present, for ZDoom maps (Hexen and UDMF formats), allowing ACS to be used. It also mostly implements the ACS line specials for ZDoom.

None of the ZDoom ACS extensions have been implemented. This also does not fully implement all ACS specials. ACS_ExecuteWithResult requires a ZDoom ACS extension to be implemented for full functionality, and ACS_ExecuteAlways might require refactoring the ACS module to allow simultaneous execution of the same script.

This may be of little use now, but may come in handy if this port receives more extended ACS support, or if map makers consider targeting this configuration in the future. It's better than not having ACS at all, at least I think.


kraflab:
I don't have any interest in adding ACS to the port right now. Simply turning acs on is not an obstacle as you have found. Formally adding ACS to the port will require a new specification and heavy testing, and most likely a whole new implementation of the acs engine itself. Until I decide to go in that direction, it actually is worse to have something in the port than not, because support for it would be in a transient state, unhealthy for both mapping and demo recording. There are other priorities in the meantime.

Issue #460, opened 2023-12-16
Submit Flatpak to Flathub


spazzylemons:
I've created a mostly complete Flatpak repository for DSDA-Doom. Flatpak is a system for packaging sandboxed Linux applications for easy installation. Flatpak packages are primarily uploaded and downloaded from [Flathub](https://flathub.org/), which allows Linux users to install applications that they might not have access to via a package manager, without needing to compile anything. My motivation for making this was actually so that I could install DSDA-Doom on my Steam Deck, since the preferred method of installing non-Steam apps on the Steam Deck is through Flatpak due to system packages being reset after each update. I've outlined what I think would need to be done before submitting the package to Flathub in a README file in the repository: https://github.com/spazzylemons/DSDADoom-Flatpak. The main reason I ask here instead of submitting it myself is because it's preferred that the app authors maintain their own packages on Flathub. There's one small change to the code that allows the application to run under the Flatpak sandboxed filesystem that I made a pull request for: #461

kraflab:
Closing as the related change has been merged. If there are other things needed feel free to submit more requests.

mooreye:
@spazzylemons Is there everything ready now for submitting to Flathub? Have you by any chance already started getting dsda-doom onto Flathub?

mooreye:
@spazzylemons ping

spazzylemons:
I haven't uploaded it because it's preferred that the application maintainer do that, from what I understand. Last I worked on this, I was able to build it locally. I could see about uploading it myself if you'd like.

mooreye:
> I haven't uploaded it because it's preferred that the application maintainer do that, from what I understand. Last I worked on this, I was able to build it locally. I could see about uploading it myself if you'd like.

@spazzylemons The maintainer once expressed that he doesn't intend to do so: https://github.com/kraflab/dsda-doom/issues/251#issuecomment-1458562423

spazzylemons:
I'll look into getting the Flatpak package set up soon

mooreye:
> I'll look into getting the Flatpak package set up soon

Any news on this?

spazzylemons:
[...]

PR #461, opened 2023-12-16
Change data directory if running as Flatpak

spazzylemons:
This is part of an attempt I'm making to create a Flatpak port of DSDA-Doom to make it easier to install for more Linux users (see #460). This change has the application check if it is running under Flatpak by checking if the `/.flatpak-info` file exists, and if it does, it changes where the `.dsda-doom` folder is stored to be within the application's personal filesystem. Without this change, the application would need access to all files in the user's home folder.

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -295,8 +295,14 @@ const char *I_DoomExeDir(void)
   static char *base;
   if (!base)        // cache multiple requests
   {
-    char *home = M_getenv("HOME");
-    size_t len = strlen(home);
+    char *home;
+    size_t len;
+    // check if we're in Flatpak
+    if (access("/.flatpak-info", F_OK) == 0)
+      home = M_getenv("XDG_DATA_HOME");
+    else
+      home = M_getenv("HOME");
+    len = strlen(home);
 
     base = Z_Malloc(len + strlen(prboom_dir) + 1);
     strcpy(base, home);
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -286,31 +286,40 @@ const char* I_GetTempDir(void)
 
 #else
 
-const char *I_DoomExeDir(void)
+static char *I_ConcatDir(const char *var, const char *subdir)
 {
   static char buffer[PATH_MAX];
+  size_t len;
+
+  // Remove trailing slashes to be safe.
+  strncpy(buffer, var, sizeof(buffer));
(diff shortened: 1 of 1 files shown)


kraflab:
I think we probably want something more general here and should respect the xdg_data_home standard more broadly (there's another PR also moving in that direction). Maybe something like this:

1. Check if old $HOME/.dsda-doom exists -> if it does, then use it (i.e., be backwards compatible with existing installations)
2. Check if $XDG_DATA_HOME exists -> if it does, use $XDG_DATA_HOME/dsda-doom (I assume no preceding `.` is used in this case)
3. If it doesn't exist, use $HOME/.dsda-doom as before

What do you think?

spazzylemons:
Sounds good -- if $XDG_DATA_HOME isn't defined, should $HOME/.local/share be used as a fallback as per the standard, or should $HOME/.dsda-doom be created in that case?

kraflab:
Oh sure, you can go with the .local/share one for consistency

prboom2/src/SDL/i_system.c
```diff
@@ -285,24 +285,41 @@ const char* I_GetTempDir(void)
...
 
 #else
-// cph - V.Aguilar (5/30/99) suggested return ~/.lxdoom/, creating
-//  if non-existant
-// cph 2006/07/23 - give prboom+ its own dir
-static const char prboom_dir[] = {"/.dsda-doom"}; // Mead rem extra slash 8/21/03
+
+static char *I_ConcatDir(const char *var, const char *subdir)
+{
+  static char buffer[PATH_MAX];
```
kraflab: It's too dangerous to have this as a static, in case we reuse this function later on. Let's allocate the result using the string lengths of the args.

prboom2/src/SDL/i_system.c
```diff
@@ -285,24 +285,41 @@ const char* I_GetTempDir(void)
...
-
-    base = Z_Malloc(len + strlen(prboom_dir) + 1);
-    strcpy(base, home);
-    // I've had trouble with trailing slashes before...
-    if (base[len-1] == '/') base[len-1] = 0;
-    strcat(base, prboom_dir);
+    // First, try legacy directory.
+    base = I_ConcatDir(home, "/.dsda-doom");
+    if (access(base, F_OK) != 0)
+    {
```
kraflab, line 318: Once I_ConcatDir allocates its response, we'll need to Z_Free `base` here before getting the new value.

spazzylemons:
Making a new pull request for this mainly so that I can move these changes out of my main branch

PR #477, opened 2024-01-27
Color demo progress bar for Heretic and Hexen

kitchen-ace:
Palette colour 4 in those games is near-black, while 255 is white.

### DIFF
--- a/prboom2/src/e6y.c
+++ b/prboom2/src/e6y.c
@@ -756,7 +756,7 @@ int HU_DrawDemoProgress(int force)
 
   prev_len = len;
 
-  V_FillRect(0, 0, SCREENHEIGHT - 4, len - 0, 4, 4);
+  V_FillRect(0, 0, SCREENHEIGHT - 4, len - 0, 4, (dsda_Flag(dsda_arg_heretic) || dsda_Flag(dsda_arg_hexen)) ? 255 : 4);
   if (len > 4)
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Alternatively, you could determine the lightest and darkest color in the current palette during init and use these.

https://github.com/fabiangreffrath/woof/blob/8412588aff0b6247f7a601625480de614aa1064c/src/v_video.c#L259-L260

prboom2/src/e6y.c
```diff
@@ -756,7 +756,7 @@ int HU_DrawDemoProgress(int force)
 
   prev_len = len;
 
-  V_FillRect(0, 0, SCREENHEIGHT - 4, len - 0, 4, 4);
+  V_FillRect(0, 0, SCREENHEIGHT - 4, len - 0, 4, (dsda_Flag(dsda_arg_heretic) || dsda_Flag(dsda_arg_hexen)) ? 255 : 4);
```
kraflab, line 759: The shorthand for this is `raven ? 255 : 4`. But I would probably add a variable because I try to avoid having value branches like this all over the place.

Take a look at https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/dsda/global.c#L127 and add something like `g_demo_progress_color`.

kraflab:
https://github.com/kraflab/dsda-doom/commit/844d49ca19f655898faa735403d7705f8351407d

kitchen-ace:
Thanks, I got busy and then forgot all about this.

PR #485, opened 2024-03-20
New "Exit Normal" and "Exit Secret" cheat codes

XaserAcheron:
Adds a couple of new cheat codes (and keybinds):

`idexitn` -- exits the map normally, as if an exit linedef was used
`idexits` -- exits the map to the secret level, as if a secret exit linedef was used

These two are super-useful for testing out (U)MAPINFO work and verifying that intermissions work as intended, since IDCLEV and company skip the intermission entirely.


XaserAcheron:
Sorry to bother y'all with this, but before merging, would someone be able to double-check that the options appear in the menu correctly? I had this all tested and working, but needed to merge latest `master` to resolve a minor conflict and cmake no longer works for me, so I'm unable to test this out post-merge until all that gets resolved. 9_6

kraflab:
@XaserAcheron were you ever able to get cmake working?

XaserAcheron:
Sorry, I ended up getting mired in Real Life(tm) shenanigans -- it may be a while before I can get back to this one.

The fact there's bindable console commands makes this technically doable without this PR, though, so it's not really that essential. :P

kraflab:
We can follow up on discord or open a separate issue about that in the future in that case, I'll close this one.

PR #486, opened 2024-03-21
Fix player sounds being cut off by item pickups in vanilla complevels

jackrjli:
This is a different fix to the same problem that https://github.com/coelckers/prboom-plus/pull/330 was trying to solve.

When picking up an item, the vanilla code calls `S_StartSound` with a `NULL` origin, so the pickup sound does not cut off player sounds, which use `player->mo` as the origin. Boom changes the origin to the player and makes weapon pickups audible to other players, but also changes `s_sound.c` to distinguish between pickup sounds and non-pickup sounds so they don't cut each other off. `comp_sound` disables the pickup sound logic in `s_sound.c`, but did not change the sound origin back to `NULL` until this commit.


kraflab (approved):
Nice! Thanks for the explanation and the clean fix 👍

PR #490, opened 2024-05-01
Small OpenGL fixes

Pedro-Beirao:
- When a flat of height close to the map height limits has a Skytexture, it would result in HOM

- Remove SMALLDELTA
It caused walls to be rendered a fraction of a unit off. 
This had already been done in https://github.com/kraflab/dsda-doom/commit/948c28a354bf0f94fa95650bec46fbf1141a41f4,  but was reverted in https://github.com/kraflab/dsda-doom/commit/8ed83163272e9c8ac57585e2b4934e777cf598e2.  I tried all sorts of maps,  but couldnt replicate the mentioned "dots on seams between flats and walls",  so I assume that its a bug of the past.

### DIFF
--- a/prboom2/src/gl_intern.h
+++ b/prboom2/src/gl_intern.h
@@ -39,8 +39,6 @@
 
 #define MAXCOORD (32767.0f / MAP_COEFF)
 
-#define SMALLDELTA 0.001f
-
 typedef enum
 {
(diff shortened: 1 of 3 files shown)


kraflab:
Thanks a ton for investigating and fixing this stuff 🎉

PR #492, opened 2024-05-16
Fix z-fighting on bottom texture with skytexture

Pedro-Beirao:
If the sector height is 0 and the floor is a skytexture,  then in the bottom texture a sky hack would be drawn
I cant seem to figure out why this if statement exists in the first place

This change prevents the sky from being drawn if a bottom texture is already there
So this shouldnt break anything


kraflab:
@andrey-budko do you have any context on this code?

kraflab:
Going to merge this and hope if there are any issues they get discovered in the beta 🤠

PR #493, opened 2024-05-30
Allow macOS and Linux to find files in the executable directory

Pedro-Beirao:
Ive been meaning to make this PR for like a year :P

This is most useful for finding dsda-doom.wad. When installing/updating, you would need to copy dsda-doom.wad to some folder that dsda searched (~/.dsda-doom for example). This sucked, so adding the executable directory to the search path is pretty helpful.

Will create merge conflicts in PR https://github.com/kraflab/dsda-doom/pull/491

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -236,7 +236,7 @@ void I_SwitchToWindow(HWND hwnd)
   }
 }
 
-const char *I_DoomDir(void)
+const char *I_ConfigDir(void)
 {
   return I_ExeDir();
 }
@@ -289,7 +289,7 @@ const char* I_GetTempDir(void)
 
 #elif defined(AMIGA)
 
-const char *I_DoomDir(void)
+const char *I_ConfigDir(void)
 {
   return "PROGDIR:";
 }
@@ -310,7 +310,7 @@ const char* I_GetTempDir(void)
 // cph 2006/07/23 - give prboom+ its own dir
 static const char prboom_dir[] = {"/.dsda-doom"}; // Mead rem extra slash 8/21/03
 
-const char *I_DoomDir(void)
+const char *I_ConfigDir(void)
 {
   static char *base;
   if (!base)        // cache multiple requests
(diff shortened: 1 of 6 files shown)


prboom2/src/SDL/i_system.c
```diff
@@ -236,7 +236,12 @@ void I_SwitchToWindow(HWND hwnd)
   }
 }
 
-const char *I_DoomExeDir(void)
+const char *I_DoomDir(void)
```
kraflab: Shall we rather call this `I_ConfigDir`?

Pedro-Beirao: I was originally going to name it that, but opted with DoomDir since that folder hosts more than just the config file.
Made a new commit that changes the name

kraflab: Hm, maybe `I_DataDir`? It's maybe too generic, plus confusing since dsda-doom has a "data dir" as well that's one level below that. Any other ideas?

Pedro-Beirao: I like `I_ConfigDir`.
Its pretty clear what it represents, so no one will be confused.

prboom2/src/SDL/i_system.c
```diff
@@ -318,6 +328,32 @@ const char *I_DoomExeDir(void)
...
+      while (p > base && *p!='/' && *p!='\\')
+        *p--=0;
+      if (*p=='/' || *p=='\\')
+        *p--=0;
+      if (strlen(base) < 2 || !M_WriteAccess(base))
+      {
+        Z_Free(base);
+        base = (char*)Z_Malloc(1024);
+        if (!M_getcwd(base, 1024) || !M_WriteAccess(base))
+          strcpy(base, current_dir_dummy);
```
kraflab: Can we just do `strcpy(base, ".");`? Does this dummy variable need to persist for some reason?

Pedro-Beirao: No there is no reason it needs to persist. I simply copied over the code from the Windows function, didnt even try to understand it.
Made a new commit.

kraflab (approved):
Thanks 👍

PR #494, opened 2024-06-03
Dont limit background fps when doing timedemo or fastdemo

Pedro-Beirao:


prboom2/src/dsda/time.c
```diff
@@ -107,7 +108,7 @@ void dsda_LimitFPS(void) {
   int allow_limit;
   int fps_limit;
 
-  allow_limit = movement_smooth || !window_focused;
+  allow_limit = (movement_smooth || !window_focused) && !dsda_Arg(dsda_arg_timedemo)->found && !dsda_Arg(dsda_arg_fastdemo)->found;
```
kraflab: The shorthand for this is `dsda_Flag(<arg>)`

Pedro-Beirao: I looked at the code for recognising the command line arguments when making this change, so I didnt know dsda_Flag was a thing.
Made a new commit.

kraflab (approved):
Thanks 👍

PR #496, opened 2024-06-16
Lineskip practice trails

MarkJeronimus:
A toggle is added to the 'automap' section of the options menu. (No key binding is added.)

When the full automap cheat is used, this feature shows a trail of the last 350 unique coordinates of the player on the map (at least 10 seconds of movement). This can greatly aid speedrunners in finding successful lineskip set-ups.

The image below shows a successful lineskip in MAP01, southwards, if that line had a trigger: one tic your center is still above the line, next tic your entire hitbox is below the line.

image

PS. I hope the "local player" implementation is correct in case of network play (I haven't tested this)

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -447,9 +447,10 @@ typedef struct
  fixed_t x, y;
 } trailpoint_t;
 
-markpoint_t *trailpoints = NULL;
 #define TRAIL_LENGTH (350) // 10 seconds of trails
-int marktrailpos = 0; // pointer into the curcular buffer (sized TRAIL_LENGTH)
+trailpoint_t *trailpoints = NULL;
+int trailpos  = 0; // pointer into the curcular buffer (sized TRAIL_LENGTH)
+int trailsize = 0; // fill level of the circular buffer (for drawing)
 
 am_frame_t am_frame;
 
@@ -604,16 +605,17 @@ static void AM_updateTrail(void) {
 
     AM_GetMobjPosition(plr->mo, &pt, &angle);
 
-    last = (marktrailpos + (TRAIL_LENGTH - 1)) % TRAIL_LENGTH;
+    last = (trailpos + (TRAIL_LENGTH - 1)) % TRAIL_LENGTH;
     moved = trailpoints[last].x != pt.x ||
             trailpoints[last].y != pt.y;
 
     if (moved) {
-      // Override the previously set parameters in the same style as AM_setMarkParams()
-      trailpoints[marktrailpos].x = pt.x;
-      trailpoints[marktrailpos].y = pt.y;
-      trailpoints[marktrailpos].label[0] = '\0';
-      marktrailpos = (marktrailpos + 1) % TRAIL_LENGTH;
+      trailpoints[trailpos].x = pt.x;
+      trailpoints[trailpos].y = pt.y;
+      trailpos = (trailpos + 1) % TRAIL_LENGTH;
+
+      if (trailsize < trailpos)
+        trailsize = trailpos;
     }
   }
 }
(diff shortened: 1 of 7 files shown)


MarkJeronimus:
Thoughts:
- Works best with map textures turned off. Textures are on by default, and this may confuse users who didn't rtfm.
- The option only works with IDDT×2 (to ensure 'strict' no-cheat compatibility when recording actual speedruns), this may also confuse players that expect it to "just work" when the option is enabled.
- When the option is enabled, the trail is calculated even when not visible via IDDT. This is my suggestion, not a requirement.

prboom2/src/m_menu.c
```diff
@@ -3949,7 +3950,6 @@ setup_menu_t helpstrings[] =  // HELP screen strings
   {"STRAFE RIGHT",S_SKIP|S_INPUT,m_null,KT_X3,0,dsda_input_straferight},
   {"STRAFE"      ,S_SKIP|S_INPUT,m_null,KT_X3,0,dsda_input_strafe},
   {"AUTORUN"     ,S_SKIP|S_INPUT,m_null,KT_X3,0,dsda_input_autorun},
-  {"180 TURN"    ,S_SKIP|S_INPUT,m_null,KT_X3,0,dsda_input_reverse},
```
kraflab, line 3952: Why is this change here?

prboom2/src/g_game.c
```diff
@@ -3065,6 +3066,7 @@ void G_InitNew(int skill, int episode, int map, dboolean prepare)
 
   //jff 4/16/98 force marks on automap cleared every new level start
   AM_clearMarks();
+  AM_clearTrail();
```
kraflab, line 3069: Let's rather include the call to AM_clearTrail inside AM_clearMarks itself.

prboom2/src/m_menu.c
```diff
@@ -2701,6 +2701,7 @@ setup_menu_t auto_settings1[] =  // 1st AutoMap Settings screen
   { "Enable textured display", S_YESNO, m_conf, AU_X, dsda_config_map_textured },
   { "Things appearance", S_CHOICE, m_conf, AU_X, dsda_config_map_things_appearance, 0, map_things_appearance_list },
   { "Show Minimap", S_YESNO, m_conf, AU_X, dsda_config_show_minimap },
+  { "Lineskip practice trails", S_YESNO, m_conf, AU_X, dsda_config_map_trail },
```
kraflab, line 2704: Let's call this "Player position trail" as it could be relevant for more things than just line skips.

prboom2/src/am_map.c
```diff
@@ -1218,6 +1278,9 @@ void AM_Ticker (void)
 
   if (stop_zooming && leveltime - zoom_leveltime != 1)
     AM_StopZooming();
+
+  if (dsda_IntConfig(dsda_config_map_trail))
+    AM_updateTrail();
```
kraflab, line 1283: I think this isn't quite exactly where we want this to happen. The critical action is when the player's position is updated, especially if it means they can trigger lines at that moment. That happens at the tail end of P_TryMove. It's not quite the same as a once per tick position check because high speeds cause multiple P_TryMove calls within a tic (this is why wallrunning works, for example).

kraflab (commented):
Nice work! There are some small adjustments to be made but the feature is sound 👍

prboom2/src/m_menu.c
MarkJeronimus: Just something I stumbled upon while working with the sources. If there's a reason it should not be removed, then you can drop this commit (idk if that's how it works)

prboom2/src/g_game.c
MarkJeronimus: I consider trails separate from marks

prboom2/src/m_menu.c
MarkJeronimus: Sounds good to me.

prboom2/src/am_map.c
MarkJeronimus: Yup. I thought about split movements and moved it to the back of my mind, as something to tackle later, and then completely forgot about it.

kraflab:
I've adopted the foundations of this PR here: https://github.com/kraflab/dsda-doom/commit/ec542d1837c2eee9aa4d5d9257631c49be31622f (basically the core code with some changes but not hooked up yet). I'm testing different integrations and will probably extend this to a few different modes (tracking all position checks vs checking locations where lines are checked etc). I have a video on the discord if you're curious how that looks with fences for instance. I'll hook things up soon so no need to update this PR, but I'll leave it up until the feature is available in the dev branch.

MarkJeronimus:
Notes:
I made a typo (style error) in trailpoint_t. I indented one space instead of two. I like to keep in the style on whatever I'm working on and because this differs from my own style it wasn't obvious to me.

If you're drawing 'Cross marking center' and 'Bounding box' in the same color, it would make more sense to draw 'Bounding box' first instead of the other way around (which was only the case so the Bounding box was drawn on top)

I'll look for the discord. [edit] I can't find it.

kraflab:
Finished for now https://github.com/kraflab/dsda-doom/commit/60e9c15ed857a334fc6f779beb9628b8ace4ac25

@MarkJeronimus https://discord.gg/yxaFP8T59H

MarkJeronimus:
I found that invite before, but I was searching for DSDA, not Hell on Earth.

But the invite link says I'm banned there. Since I've never been there nor known anyone from there I can only assume it's whitelisted?

PR #497, opened 2024-06-23
Fix weapon sprite offset for irregular aspect ratios

Omnisthetics:
When using a custom resolution with an aspect ratio not found in the settings, the weapon sprite is rendered away from the bottom of the window. This PR keeps the sprite in line with the bottom at any resolution.

Before (720x720):
before

After (720x720):
after

### DIFF
--- a/prboom2/src/v_video.c
+++ b/prboom2/src/v_video.c
@@ -1288,9 +1288,12 @@ void SetRatio(int width, int height)
   tallscreen = (ratio_scale < ratio_multiplier);
   if (tallscreen)
   {
+    float ratio_quotient = (float)ratio_multiplier/ratio_scale;
+    float ratio_percentage = (ratio_quotient - 1) * 100.0;
+    psprite_offset = (int)(ratio_percentage*FRACUNIT);
+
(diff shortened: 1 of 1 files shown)


prboom2/src/v_video.c
```diff
@@ -1285,16 +1285,17 @@ void SetRatio(int width, int height)
   ratio_scale *= 3;
   ReduceFraction(&ratio_multiplier, &ratio_scale);
 
+  {
+    float ratio_quotient = (float)ratio_multiplier/ratio_scale;
+    float ratio_percentage = (ratio_quotient - 1) * 100.0;
+    psprite_offset = MAX(0, (int)(ratio_percentage*FRACUNIT));
```
kraflab: This math only needs to be done when `tallscreen` is true, right? Let's put it inside the tallscreen block and leave the original else block in place.

Omnisthetics: Correct. My initial commit had the math inside the block but I wanted to get a little fancy and use MAX to make the else branch unnecessary.

I'll revert my later commits.

kraflab:
Thanks 👍

PR #498, opened 2024-06-27
allow shift+<key> when naming savegame

Phytolizer:
useful for symbols such as ()

### DIFF
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -5000,7 +5000,10 @@ dboolean M_Responder (event_t* ev) {
     }
     else if (ch > 0)
     {
-      ch = toupper(ch);
+      if (shiftdown && ch >= 32 && ch <= 127)
+        ch = shiftxform[ch];
+      else
+        ch = toupper(ch);
       if (ch >= 32 && ch <= 127 &&
           saveCharIndex < SAVESTRINGSIZE-1 &&
           M_StringWidth(savegamestrings[saveSlot]) < (SAVESTRINGSIZE-2)*8)
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -5000,14 +5000,14 @@ dboolean M_Responder (event_t* ev) {
     }
     else if (ch > 0)
     {
-      if (shiftdown && ch >= 32 && ch <= 127)
-        ch = shiftxform[ch];
-      else
-        ch = toupper(ch);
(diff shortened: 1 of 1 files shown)


prboom2/src/m_menu.c
```diff
@@ -5000,7 +5000,10 @@ dboolean M_Responder (event_t* ev) {
     }
     else if (ch > 0)
     {
-      ch = toupper(ch);
+      if (shiftdown && ch >= 32 && ch <= 127)
```
JadingTsunami: I suppose you could move this inside the `if` below since the range check is done by that point.

Phytolizer:
Future work (not this PR): add full text entry akin to the console

This was mentioned on Discord, replicating it here for accessibility.

kraflab:
I'll probably take a look at full text entry prior to the next release (actually, I'm gonna be taking a look at the whole menu system and input is just a subset of that) and will hold onto this as a fallback if it ends up being too problematic.

kraflab:
Well here's one problem, the font drawing these strings is incomplete, so only some symbols work 😅

Phytolizer:
Hmm... is there a "default" glyph like the tofu in modern fonts? Maybe it could draw that as a failsafe

kraflab:
By default it draws an `!`. I tested out falling back on the hud font and it's pretty reasonable I think. It just affects a few symbols. There's a bigger problem though, which is that none of this works for heretic / hexen. There are 2 issues: their text drawing code works differently, assuming contiguous font data in the wad (so drawing some symbols causes an overrun into junk data), and since they never expected people to type symbols, the font has symbols in the wrong spots (I think `[` is `_` for instance).

kraflab:
Oh actually, if you type to enter shift + ] on master it actually crashes in heretic, so this is already an issue 😂

kraflab:
https://github.com/kraflab/dsda-doom/commit/c96b49addc8866305acefc71c42e3e3f3c283337 and https://github.com/kraflab/dsda-doom/commit/22e6580120602adbefab4e22fa6df5d0ad891348 fyi.

Going to need to set up guard clauses for raven or set it up to use the existing font arrays instead of diving into the lumps haphazardly.

kraflab:
PR committed here, excluding raven until a full solution is in place for that: https://github.com/kraflab/dsda-doom/commit/6334d11ff6dcb87b6fcfbf36dbea8ac98f7cd29f

Thanks!

PR #503, opened 2024-08-21
Fix potential hang after restoring keyframe due to invalid blocklist pointers.

PeterN:
Fix potential hang after restoring keyframe due to invalid blocklist pointers.

When restoring a keyframe, the blockmap is not cleared and blocklist pointers are modified during P_UnArchiveThinkers. The blockmap is then restored in UnArchiveBlockLinks, but this can leave blocklist pointers in an invalid state.

This can then lead to a hang later when the blocklist is iterated and may get stuck in an infinite loop.

The hang is fairly easy to reproduce using Legacy of Rust when Vassago are present after using keyframe restore.

I chose to simply wipe the blocklist pointers during restoring as this avoids adding extra functions or modifying much existing code. It's basically brute-forcing it.

### DIFF
--- a/prboom2/src/p_saveg.c
+++ b/prboom2/src/p_saveg.c
@@ -1610,6 +1610,10 @@ void P_UnArchiveThinkers(void) {
 
           mobj->state = states + (intptr_t) mobj->state;
 
+          // Wipe blocklist pointers. They will be restored correctly later in P_UnArchiveBlockLinks.
+          mobj->bprev = NULL;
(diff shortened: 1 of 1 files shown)


kraflab:
Thanks for the summary and solution! While it looks completely safe, I'll still need to run extensive regression testing as this code is extremely sensitive for TAS demo recording, so it might sit here for a bit until I have a chance to tackle that.

Pedro-Beirao:
Passed regression testing. Thank you!

PR #505, opened 2024-08-23
Fix music not automatically playing when using shareware iwad.

PeterN:
Fix music not automatically playing when using a shareware iwad.

musinfo is not tested or parsed for shareware iwads, but the musinfo list was not set appropriately.

This meant that when later trying to playing the music, an attempt is made to play an invalid music lump, which fails.

### DIFF
--- a/prboom2/src/p_setup.c
+++ b/prboom2/src/p_setup.c
@@ -3677,10 +3677,7 @@ void P_SetupLevel(int episode, int map, int playermask, int skill)
   lumpnum = W_GetNumForName(lumpname);
 
   // Must process musinfo to get default track before calling S_Start
-  if (gamemode != shareware)
-  {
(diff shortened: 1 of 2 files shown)


prboom2/src/p_setup.c
```diff
@@ -3681,6 +3681,10 @@ void P_SetupLevel(int episode, int map, int playermask, int skill)
   {
     S_ParseMusInfo(lumpname);
```
kraflab: Could you instead migrate the condition here into `S_ParseMusInfo`: `if (gamemode != shareware && W_LumpNameExists("MUSINFO"))`? That function already has the else condition so it should isolate the behavior better.

kraflab (approved):
Thanks 👍

PR #506, opened 2024-08-29
Add release automation and artifacts uploading

rfomin:
This will only upload the MSYS2 build. For Linux, we can add an AppImage integration.

For macOS we can try [CPack DMG generator](https://cmake.org/cmake/help/latest/cpack_gen/dmg.html#cpack_gen:CPack%20DragNDrop%20Generator), but I don't have access to Mac for testing. Maybe @Pedro-Beirao can help.

Push a new tag to the repo to create a new release:
```
git tag "v0.28.2"
git push origin tag "v0.28.2" 
```
`patch_notes/v0.28.2.md` will be used for the changelog. Example: https://github.com/rfomin/dsda-doom/releases/tag/v0.28.2


kraflab:
Does it also work with releases created through the github ui? I like to keep the hotfix patch notes in the main patch notes file for big releases (so e.g. v0.28.md has v0.28.2 in it as well), so it would be best if this didn't need a separate file to work.

rfomin:
> Does it also work with releases created through the github ui?

Yes. I pressed "Draft new release"->"Create new tag v0.28.4"->"Publish release", here the result: https://github.com/rfomin/dsda-doom/releases/tag/v0.28.4

> I like to keep the hotfix patch notes in the main patch notes file for big releases (so e.g. v0.28.md has v0.28.2 in it as well), so it would be best if this didn't need a separate file to work.

I think we can search and copy lines from these files in a script, but it might be easier if we add RELEASE_NOTES.md to the root. Or you just edit the release in the github ui.

Are you interested in builds for Linux and macOS? Adding a Linux AppImage is easy, while macOS can take some time.

kraflab:
I'm fine with just adding the notes myself in the ui for now. We can always improve the process later on if needed. I think it would make sense to include the other platforms as well. We could start with just msys2 and linux for now and Pedro could take care of the macos one later on if it's a bit more involved.

prboom2/ICONS/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+if(UNIX AND NOT APPLE)
+    install(FILES dsda-doom.desktop DESTINATION share/applications)
+    install(FILES dsda-doom.svg DESTINATION share/icons/hicolor/scalable/apps)
```
rfomin, line 3: @fabiangreffrath Is an SVG icon enough? Or do we need to add a 128x128 PNG file?

fabiangreffrath: Svg is fine.

kraflab:
Hm, any idea why the build is twice the size of mine? 🤔 
Also, could we add the `COPYING` file that's included with current builds as well?

rfomin:
> Hm, any idea why the build is twice the size of mine? 🤔

It seems that the autobuild has more SDL_image dependencies ([libjxl](https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-libjxl), [libavif](https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-libavif), etc). These are optional, like "plugins".

To get full control over the libraries, we can switch to vcpkg.

kraflab:
> > Hm, any idea why the build is twice the size of mine? 🤔
> 
> It seems that the autobuild has more SDL_image dependencies ([libjxl](https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-libjxl), [libavif](https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-libavif), etc). These are optional, like "plugins".
> 
> To get full control over the libraries, we can switch to vcpkg.

I guess it's fine as it is, the size is not too big in the end.

rfomin:
> Also, could we add the `COPYING` file that's included with current builds as well?

Done in https://github.com/kraflab/dsda-doom/pull/506/commits/b25741d7e693b11ca7a391896435110e38dca19b

I think it's ready at this point. I need to learn how things are done on macOS to make a good package. I have plans to make a macOS version of the Woof distribution.

kraflab:
Thanks 👍

PR #512, opened 2024-09-29
Random sound pitch for arbitrary sample rate

FozzeY:
This fixes an oversight where the sound pitch to stepping lookup table did not account for different sample rates of the sounds, which resulted in the effect not working properly at higher sample rates.

### DIFF
--- a/prboom2/src/SDL/i_sound.c
+++ b/prboom2/src/SDL/i_sound.c
@@ -111,7 +111,7 @@ typedef struct
 
 channel_info_t channelinfo[MAX_CHANNELS];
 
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Fixed differently in PrBoom+:

https://github.com/coelckers/prboom-plus/commit/02eeb49dbcad7042d6ac572f2d04f1b1b090acee

FozzeY:
Oh wow, I didn't think to check PrBoom+ for that. It does look a lot better.

kraflab:
Thanks 👍

PR #514, opened 2024-10-03
Add a new thinker thread for teleports

MarkJeronimus:
This fixes the extreme lag in Oversaturation when thousands of zombies try to teleport in, and the th_misc class thinker list is around 60000 elements long.

Bofu correctrly voices his concern about desyncs. I think, if the teleport destinations in the list can be guaranteed to be in the same order, there should be no desyncs.

### DIFF
--- a/prboom2/src/p_telept.c
+++ b/prboom2/src/p_telept.c
@@ -145,7 +145,7 @@ static mobj_t* P_TeleportDestination(short thing_id, int tag)
   FIND_SECTORS(id_p, tag)
   {
     register thinker_t* th = NULL;
-    while ((th = P_NextThinker(th,th_misc)) != NULL)
+    while ((th = P_NextThinker(th,th_teleport)) != NULL)
       if (th->function == P_MobjThinker) {
         register mobj_t* m = (mobj_t*)th;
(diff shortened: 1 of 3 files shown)


kraflab:
I don't think this is a safe change - not because of the order of the teleporter list, but because they aren't in the old list anymore. Couldn't you potentially give the teleport destination monster properties with dehacked? You can also give it more physical properties in udmf that might matter. Rather than changing the core thinker list, I would suggest maintaining a separate teleporter list (probably in p_telept, exposed by stuff like `P_RegisterTeleportDestination`).

MarkJeronimus:
It doesn't desync on this monstrous demo, so I think we're safe.
https://www.doomworld.com/forum/post/2459281

MarkJeronimus:
Lol I didn't even see your reply even when I replied. Yeah, that could be a problem in maps that use more exotic stuff if there's a need for teleport destinations in the main list. You're free to rewrite this as you see fit. For Oversaturation, Vytaan is happy using it as we speak, even if it's a one-off build.

kraflab:
Closing this PR but anyone who wants to take the alternate approach can create a new one in the future.

PR #515, opened 2024-10-04
Update CI workflow

rfomin:
* Switch from nuget to regular actions cache for vcpkg build.

* Don't use bash on Windows.


kraflab:
Thanks 👍

PR #516, opened 2024-10-05
Update INSTALL file with distribution-specific instructions

otrw:
Updated INSTALL instructions with distribution-specific package information (Fedora 40 & Ubuntu 24.04) and CMake build instructions as well as minor spelling corrections. These instructions have been tested on both virtual machines and bare metal installations to ensure accuracy.

The existing ./configure information has been preserved, though I'm unsure if this method is still supported. Happy to amend if it is no longer the case.

Testing environments:
- Fedora 40: Tested on VM and bare metal
- Ubuntu 24.04: Tested on VM and bare metal
All tests resulted in successful builds and installations of DSDA-Doom.

### DIFF
--- a/prboom2/INSTALL
+++ b/prboom2/INSTALL
@@ -1,239 +1,220 @@
 Basic Installation (ZIP, tarball, Git)
 ==================
+Prerequisites
 
-  1. To compile DSDA-Doom, you need at least SDL2. To enable various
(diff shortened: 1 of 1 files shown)


kraflab:
Thanks 👍 
I don't know about ./configure, but it can be removed separately if it's not relevant anymore.

PR #517, opened 2024-10-06
Config variable for updating unexplored parts in automap mode

FozzeY:
A setting for vanilla behaviour I'd like to see return.

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -152,6 +152,8 @@ FILE    *debugfile;
 
 dboolean advancedemo;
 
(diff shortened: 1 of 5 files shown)


kraflab:
What's the use case of this option?

FozzeY:
For completionists it helps to make sure that all areas of the level are checked in first person, no accidental reveals when backtracking in automap mode, for example. I haven't researched whether it was a design decision or a limitation in vanilla, but commercial ports preserve this behaviour.

kraflab:
I don't feel like this is worth adding another option to be honest.

PR #527, opened 2024-10-30
Midi volume fixes

ceski-1:
1. Some midis neglect to set an initial volume, causing issues when the song loops (e.g. 1994TU.wad map17, drums fade out forever). Vanilla Doom and the OPL emulator reset midi volume at the end of each song loop, so this change does the same for Fluidsynth and Portmidi.

2. Users continue to report that the volume curve is too loud, so I'm reverting an old change I made. I already did this for Woof and will probably do the same for other ports.


kraflab:
Thanks 👍

Issue #530, opened 2024-11-11
Support UMAPINFO in heretic


kraflab:
We have a request for at least a partial implementation of UMAPINFO in heretic for the QTR2 project (happening some time in 2025).

These features specifically have been requested:
- map name
- map sky texture
- midi track
- next map, regular
- next map, secret
- par time
- author
- level number
- Intertext
- enterpic / exitpic

Most of these probably just involve turning it on, since it's currently disabled for heretic. Full UMAPINFO support isn't really possible since the spec has stuff like doom boss actions in it and no defined heretic actor list etc.

thecnoNSMB:
The Heretic patch in [this Doomworld thread](https://www.doomworld.com/forum/topic/136079-quality-of-life-patches-for-various-wads/) also uses UMAPINFO to define an episode for the extra deathmatch map(s).

PR #532, opened 2024-11-16
Heretic umapinfo support

Pedro-Beirao:
Issue https://github.com/kraflab/dsda-doom/issues/530

This is an attempt to implement umapinfo support in the current intermission/finale codebase. I will want to rewrite/refactor some of these files to better organize the specific intermission/finale code for doom, heretic and hexen, and the implementation of umapinfo. Currently, heretic and hexen are just "patched in" with the old doom code, making it hard to work with

Since there is no spec for umapinfo in heretic, I went with a mix of: do like vanilla heretic and do like doom's umapinfo
```
levelname, label, author, next, nextsecret, skytexture, music, endgame, endpic, nointermission, episode - simple implementation

partime - heretic does not have partimes. now it displays it like doom

exitpic, enterpic - in the first 3 eps, heretic has 3 intermission screens: 'status', 'finished' and 'entering'. in the other eps, it only has 1, the 'status' screen. here, i followed this behaiviour. but should it instead be like doom and only have a 'status' and a 'entering' screens on all eps?

enddemon - introduced, similar to endbunny

endcast - not implemented yet

intertext, intertextsecret, interbackdrop, intermusic - implemented

bossaction - implemented
```

~~Old test wad: [heretic-umapinfo.wad.zip](https://github.com/user-attachments/files/17787623/heretic-umapinfo.wad.zip)~~
Test wad: [heretic-umapinfo.wad.zip](https://github.com/user-attachments/files/21702806/heretic-umapinfo.wad.zip)


Pedro-Beirao:
Heretic vanilla behaviour:
- In the first 3 episodes, first the **Stats** screen is shown, then the **Finished**, then the **Entering**
Pressing Use in **Stats** will send you straight to **Entering**.
- In the other episodes there is only the **Stats** screen, which contains a "Finished" text at the bottom. Pressing Use will send you straight to the next level.

So to streamline this in UMAPINFO, if a level defines `enterpic`, it will use the behaviour of the first 3 episodes.

Pedro-Beirao:
The actor names are still point of discussion, but the rest should be all ok.
The heretic intermission and finale code is quite different from doom's, hopefully everything is implemented correctly, let me know if you find any problem

prboom2/src/dsda/palette.c
```diff
@@ -73,14 +73,16 @@ void dsda_SetPlayPal(int index) {
   playpal_index = index;
 }
 
-void dsda_FreePlayPal(void) {
-  int playpal_i;
+void dsda_FreePlayPal(int playpal_i) {
+  if (playpal_data[playpal_i].lump) {
+    Z_Free(playpal_data[playpal_i].lump);
+    playpal_data[playpal_i].lump = NULL;
```
andrikpowell, line 79: we prob want to do a full clear, just in case:

```
void dsda_FreePlayPal(int playpal_i) {
  if (playpal_data[playpal_i].lump) {
    Z_Free(playpal_data[playpal_i].lump);
    playpal_data[playpal_i].lump = NULL;
  }
  if (playpal_data[playpal_i].colours) {
    Z_Free(playpal_data[playpal_i].colours);
    playpal_data[playpal_i].colours= NULL;
  }
  playpal_data[playpal_i].length = 0;
}
```

`I_UploadNewPalette()` can leak old SDL colors

it has this line:
```
((playpal_data->colours == NULL) ||
 (cachedgamma != usegamma) ||
 force)
```

SDL can receive the old palette instead of the new palette when there are 2 custom palettes.

length isn't really required, idk if it'd affect anything. but prob doesn't hurt.

andrikpowell: we should also probably clear `transparent` and `duplicate`, because they are being reused from old custom palette (see `dsda_InitPlayPal()`).

In contrast `darkest` and `lightest` do get updated inside dsda_InitPlayPal() unconditionally, so I guess they technically aren't required here.

```
  playpal_data[playpal_i].transparent = 0;
  playpal_data[playpal_i].duplicate = 0;
  playpal_data[playpal_i].darkest = 0;
  playpal_data[playpal_i].lightest = 0;
```

prboom2/src/dsda/mapinfo/u.c
```diff
@@ -321,7 +334,12 @@ void dsda_UFDrawer(void) {
       }
       break;
     case FINALE_STAGE_ART:
-      if (gamemapinfo->flags & MapInfo_EndGameBunny)
+      if (gamemapinfo->endpalette[0])
+      {
+        V_SetPlayPal(playpal_custom);
```
andrikpowell, line 339: We are calling this on every rendered frame.
We should probably have some kind of caching system, so that we mark it, and don't constantly set it here.

This happens here, and in heretic's `f_finale.c`.

prboom2/src/umapinfo.cpp
andrikpowell: ok this is causing an issue with the kex heretic iwad, because it includes a umapinfo.

basically once you get to the underwater screen, you are stuck and can't do anything. I'm thinking that the reason is that it's using VICTORY2

This is kind of an overall problem here in that we are parsing doom logic here and not heretic logic.

prboom2/src/heretic/f_finale.c
```diff
@@ -85,17 +94,29 @@ void Heretic_F_StartFinale(void)
...
   finalecount = 0;
-  FontABaseLump = W_GetNumForName("FONTA_S") + 1;
-  S_ChangeMusic(heretic_mus_cptd, true);
 }
 
 static dboolean F_BlockingInput(void)   // Avoid bringing up menu when loading Heretic's custom E2 palette
 {
-  return finalestage == 1 && gameepisode == 2;
+  return (finalestage == 1) &&
+          ((endgameflags & (MapInfo_EndGameClear|MapInfo_EndGameAny)) ? endpalette != NULL : gameepisode == 2);
```
andrikpowell: `endpalette != NULL` should be `endpalette && endpalette[0]`

prboom2/src/heretic/f_finale.c
```diff
@@ -276,32 +298,64 @@ void F_DrawUnderwater(void)
...
-      case 5:
-        V_DrawRawScreen("CREDIT");
-        break;
-    }
+        return;
+      }
+
+      if (W_LumpNameExists(endpic))
+      {
+        V_DrawNamePatch(0, 0, 0, endpic, CR_DEFAULT, VPT_STRETCH);
```
andrikpowell: This should be `V_DrawRawScreen(endpic);`

prboom2/src/d_main.c
```diff
@@ -249,9 +249,9 @@ void D_PostEvent(event_t *ev)
     }
   }
 
-  if (heretic && gamestate == GS_FINALE && F_Responder(ev))
-    dsda_InputFlushTick(); // When under Heretic's Underwater Palette, make menu event invisible
-  else
+  if (gamestate == GS_FINALE)
```
andrikpowell: ok let me explain exactly why this is wrong...

what we want to happen is have the finale responder listen for a key press, and then consume it (dsda_InputFlushTick()).

Then once that happens, it will move up to the title screen.

The important thing here is that we are consuming a key press... So the way this is currently set up is that if i hit ESC, then it is not consumed, immediately moves to the title and brings up the menu.

That's incorrect here. Heretic before consumed the key press, so if i pressed ESC, it moved to the title, but if i wanted to actually open the menu, then i had to press it again once on the title...

So this matches what it was before:
```
if (gamestate == GS_FINALE && F_Responder(ev))
  dsda_InputFlushTick();
else if (M_Responder(ev))
    dsda_InputFlushTick(); // If the menu used the event, make it invisible
else
  G_Responder(ev);
```

I do take back what i said about removing F_Responder in G_Responder(), I think it's better placed here.

prboom2/src/dsda/mapinfo/u.c
Pedro-Beirao: So there is `dsda_UStartFinale()` which inits the palette only once. This is the most intensive part.
Then here the palette is simply set, doesnt cost much to do it every frame. I dont think its worth it adding even more logic to this.

prboom2/src/dsda/palette.c
Pedro-Beirao: Right, I was just going off the original code, but we should do that yes.

prboom2/src/d_main.c
Pedro-Beirao: But what about handling the cast? This will make it impossible to open the menu in the cast screen.

prboom2/src/umapinfo.cpp
Pedro-Beirao: I had this whole if tree guarded for doom only, but seems that condition got lost when I rebased on elf's UMAPINFO flags PR.

prboom2/src/d_main.c
Pedro-Beirao: Tried your code, thats exactly what happens. Cant open the menu on cast screen.

prboom2/src/info.h
```diff
@@ -5928,7 +5928,7 @@ typedef enum {
   DOOM_NUMMOBJTYPES,
 
   // heretic
-  HERETIC_MT_ZERO,
+  HERETIC_MT_ZERO = 0,
```
andrikpowell: Revert this, this is very very bad and causes desyncs for hexen. Also remove the = 0 for hexen. we want to keep all the thing numbers, since everything in dsda works on that assumption.

Take a look at the other comment to see how to address this for actor types.

prboom2/src/dsda/name.c
andrikpowell: I'd suggest not using raven here, but I guess it doesn't really matter at this point.

Replace
```
      return dsda_GetDehMobjIndex(i);
```
with:
```
      return heretic ? dsda_GetDehMobjIndex(HERETIC_MT_ZERO + i) :
                       dsda_GetDehMobjIndex(i);
```

This goes with removing `= 0` on mobj types which this pr does. This causes desyncs in hexen, as dsda is designed around mobj lists including all mobjs.

prboom2/src/dsda/mapinfo/u.c
andrikpowell: So in software this isn't a big deal...

Problem is OpenGL does this on every frame:
```
if (V_IsOpenGLMode())
  gld_FlushTextures();
```

I think we can just have a slight guard here and in heretic's `V_SetPlayPal(playpal_custom)`:
```
if (gamemapinfo->endpalette[0] && playpal_index != playpal_custom)
{
  V_SetPlayPal(playpal_custom);
}
```

andrikpowell:
This PR should be good to go now. It has passed demo testing.

PR #533, opened 2024-11-16
Add -track_reality and reset tracking in casual play

Pedro-Beirao:
Suggested in the discord. Tracks whether reality and/or almost reality have been broken.

Also noticed that -track_* parameters were not persisting beetween resets/new games without recording demos, so they get reset properly now.


fabiangreffrath (approved):
LGTM

PR #536, opened 2024-11-22
Implement blocky fuzz

Pedro-Beirao:
Based on woof and bits from https://github.com/kraflab/dsda-doom/pull/359

That PR didnt "melt" the pixels downwards like it should, so i adapted woof's code

Test wad by ceski [fuzz3.wad.zip](https://github.com/user-attachments/files/17875510/fuzz3.wad.zip)

### DIFF
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -107,7 +107,7 @@ static const byte   *tempfuzzmap;
 //#define FUZZOFF (SCREENWIDTH)
 #define FUZZOFF 1
 
-static const int fuzzoffset[FUZZTABLE] = {
+static const int fuzzoffset_org[FUZZTABLE] = {
   FUZZOFF,-FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,
   FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,
   FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,
@@ -116,6 +116,8 @@ static const int fuzzoffset[FUZZTABLE] = {
   FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,
   FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF,FUZZOFF,-FUZZOFF,FUZZOFF
 };
+
+static int fuzzoffset[FUZZTABLE];
 
 static int fuzzpos = 0;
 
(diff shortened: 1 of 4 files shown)


prboom2/src/r_drawflush.inl
```diff
@@ -45,6 +43,86 @@
...
+
+        ++fuzzpos;
+
+        // Clamp table lookup index.
+        fuzzpos &= (fuzzpos - FUZZTABLE) >> (8 * sizeof(fuzzpos) - 1); // killough 1/99
+
+        lines = fuzzcellsize;
+    } while (count);
+
+    if (cutoff)
```
rfomin: We probably don't need the `cutoff` code. It is left over from the old Woof implementation.

Pedro-Beirao: Right, I assumed I was missing something about it. removed

prboom2/src/r_draw.c
```diff
@@ -479,9 +480,11 @@ void R_InitBuffer(int width, int height)
 
   drawvars.topleft = screens[0].data;
   drawvars.pitch = screens[0].pitch;
-
-  for (i=0; i<FUZZTABLE; i++)
```
rfomin, line 483: We don't need declaration of `i` variable if you remove this cycle.

Cycle is not a bad idea, maybe we should implement it in Woof as well.

prboom2/src/r_draw.c
```diff
@@ -587,3 +590,16 @@ int R_GetFuzzPos()
...
+  R_ResetColumnBuffer();
+
+  fuzzpos = (fuzzpos + (height / fuzzcellsize)) % FUZZTABLE;
+}
+
+void R_CheckFuzzCol(int x, int height)
+{
+  if (!(x % fuzzcellsize))
+    R_ResetFuzzCol(height);
+}
```
rfomin: No "newline" at the end of the file. A small thing, but it can be annoying https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

prboom2/src/r_drawflush.inl
```diff
@@ -45,6 +43,76 @@
...
+
+    int yl = tempyl[temp_x - 1];
+    int yh = tempyh[temp_x - 1];
+
+    if (!yl)
+    {
+        yl = 1;
+    }
+
+    if (yh == viewheight - 1)
```
rfomin: To be honest, I'm not sure how "compositional" DrawColumn functions work in PrBoom+, but maybe we don't need the `if (yh == viewheight - 1)` check either.

prboom2/src/r_draw.c
Pedro-Beirao: Reimplemented cycle, even though it just saves 1 mult operation

prboom2/src/r_drawflush.inl
Pedro-Beirao: We dont need either of those checks. The values in temp_yl and temp_yh seems to always be valid.
Removed both conditions. I shouldve checked before

rfomin (approved):
LGTM

Pedro-Beirao:
Thanks!

PR #539, opened 2024-11-23
Don't cutoff secret revealed sound

Pedro-Beirao:
This has been asked a few times ever since https://github.com/kraflab/dsda-doom/pull/486

### DIFF
--- a/prboom2/src/s_sound.c
+++ b/prboom2/src/s_sound.c
@@ -328,7 +328,9 @@ void S_StartSoundAtVolume(void *origin_p, int sfx_id, int volume, int loop_timeo
     return;
 
   // killough 4/25/98
-  if (sfx_id & PICKUP_SOUND ||
+  if (sfx_id == g_sfx_secret)
+    params.sfx_class = sfx_class_secret;
+  else if (sfx_id & PICKUP_SOUND ||
       sfx_id == sfx_oof ||
       (compatibility_level >= prboom_2_compatibility && sfx_id == sfx_noway))
     params.sfx_class = sfx_class_important;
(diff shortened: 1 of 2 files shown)


prboom2/src/s_sound.c
```diff
@@ -930,9 +932,11 @@ static int S_getChannel(void *origin, sfxinfo_t *sfxinfo, sfx_params_t *params)
     return channel_not_found;
 
   // Only allow one sound per origin
+  // Preserve the secret revealed sound, unless a new one is called
   for (cnum = 0; cnum < numChannels; cnum++)
     if (channels[cnum].active && channels[cnum].origin == origin &&
-        (comp[comp_sound] || channels[cnum].sfx_class == params->sfx_class))
+        (comp[comp_sound] || channels[cnum].sfx_class == params->sfx_class) &&
```
fabiangreffrath, line 938: Couldn't you just turn this into a `>=` comparison, so that sounds with a class "higher" in the enum are never interrupted by sounds "lower" in the hirarchy?

Pedro-Beirao: Thats not what the previous code did.

Vanilla: Only 1 class allowed for the origin
Boom+: It allows both a sfx_class_none and a sfx_class_important for the same origin.

This fix adds a new sfx_class_secret allowed for that origin, and is not affected by the vanilla limit of 1.

fabiangreffrath:
@rfomin We could achieve the same in Woof by introducing a `sg_secret` sound group only for `ds_secret`. What do you think? Currently, the `ds_secret` sound is in the `sg_none` sound group and could thus get interrupted by about any other sound.

rfomin:
> @rfomin We could achieve the same in Woof by introducing a `sg_secret` sound group only for `ds_secret`.

Sounds good.

Pedro-Beirao:
So, good to merge?

fabiangreffrath (approved):
Fine with me, thank you!

fabiangreffrath:
> Sounds good.

Actually not necessary, as it turns out. The secret sound emits from `NULL` origin and most other sound that emit from that origin (i.e. oof and pickup sounds) are already in their own sound group. Weapon sounds emit from the player's coordinates and are thus not from the same origin.

PR #540, opened 2024-11-25
Automap Pan / Zoom Enhancements

andrikpowell:
This is a very simple PR that bounds the automap pan / zoom speed to the "always run" key (or the hold "run" key) to alternate the speed.

Old code wasn't based on the "always run" key, and only sped up when holding shift. Now it works like the "always run" key works in game. AKA if you press the key while set to "always run", the speed with slow down. This should make it easier to use the Automap Tag Finder.

I also split the zoom and pan speeds for further control (very useful for the Tag Finder).


Pedro-Beirao (approved):

Pedro-Beirao:
Lets merge. I prefer to get it approved by others as well, but this is simple enough

Issue #511, opened 2024-09-28
turn off auto switch weapon when pickup


Sol1vaN:
hi there,
there's a way to disable this feature?
i don't like it and i want to disable it.
on "quality of life" menu there is an option called something like:
"boom weapon auto swithc"
but this do nothing, still the weapons change when i pickup one.

dsda have an way to change correctly this?
or via console command, editing a cfg file at least?

kraflab:
Boom weapon autoswitch occurs when running out of ammo, so that's a separate thing. I think it's worth adding this feature to the casual play settings, although it won't be available while demo recording.

Pedro-Beirao:
Done in https://github.com/kraflab/dsda-doom/pull/541

PR #541, opened 2024-11-27
Add "Auto Switch Weapon on Pickup" option

andrikpowell:
Took a stab at implementing an Auto Switch option asked for in (#511).

It's a rather simple casual only setting that's disabled in strict mode, demo recording/playing, netgame, and deathmatch.

Let me know if there's better precedent when it comes to netgame / deathmatch compat for options.

### DIFF
--- a/prboom2/src/p_inter.c
+++ b/prboom2/src/p_inter.c
@@ -108,7 +108,21 @@ static weapontype_t GetAmmoChange[] = {
     wp_mace
 };
 
-#define autoswitch (allow_incompatibility && !deathmatch && !netgame ? dsda_IntConfig(dsda_config_switch_weapon_on_pickup) : true)
+//
+// P_AutoSwitchWeapon
+// Autoswitches player to a weapon,
+// Based on config and other conditions.
+//
+
+static void P_AutoSwitchWeapon(player_t *player, weapontype_t weapon)
+{
+  int autoswitch_config = dsda_IntConfig(dsda_config_switch_weapon_on_pickup);
+  int autoswitch = ((allow_incompatibility && !deathmatch && !netgame) ? autoswitch_config : true);
+
+  if (!autoswitch) return;
+
+  player->pendingweapon = weapon;
+}
 
 //
 // P_GiveAmmo
@@ -136,8 +150,7 @@ static dboolean P_GiveAmmoAutoSwitch(player_t *player, ammotype_t ammo, int olda
         weaponinfo[i].ammopershot <= player->ammo[ammo]
       )
       {
-        if (autoswitch)
-          player->pendingweapon = i;
+        P_AutoSwitchWeapon(player, i);
         break;
       }
     }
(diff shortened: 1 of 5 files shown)


prboom2/src/p_inter.c
```diff
@@ -187,43 +190,47 @@ static dboolean P_GiveAmmo(player_t *player, ammotype_t ammo, int num)
     {
```
Pedro-Beirao, line 203: You can put `if (!autoswitch) return true;` before `if (heretic)` and remove these 2 conditions

prboom2/src/p_inter.c
```diff
@@ -2298,7 +2307,8 @@ dboolean Heretic_P_GiveWeapon(player_t * player, weapontype_t weapon)
         player->bonuscount += BONUSADD;
         player->weaponowned[weapon] = true;
         P_GiveAmmo(player, wpnlev1info[weapon].ammo, GetWeaponAmmo[weapon]);
-        player->pendingweapon = weapon;
+        if (autoswitch)
+          player->pendingweapon = weapon;
```
Pedro-Beirao: The condition below is using brackets, so lets follow the style. Ik its not consistent at all, but its better if new code tries to match whats already there.

And maybe you should just extract this `if` into a new `P_AutoSwitchWeapon`

prboom2/src/p_inter.c
```diff
@@ -108,6 +108,8 @@ static weapontype_t GetAmmoChange[] = {
     wp_mace
 };
 
+#define autoswitch (allow_incompatibility && !deathmatch && !netgame ? dsda_IntConfig(dsda_config_switch_weapon_on_pickup) : true)
+
```
Pedro-Beirao, line 126: Put () around `allow_incompatibility && !deathmatch && !netgame`. I wouldnt think the ternary captures all those conditions

prboom2/src/p_inter.c
andrikpowell: Ah great catch! Also simplifies this section much more nicely :)

prboom2/src/p_inter.c
andrikpowell: ~~Hmmm we could add a `P_AutoSwitchWeapon`, although not all of these cases are `player->pendingweapon = weapon`...~~

~~The `player->pendingweapon` is the same, but the result is variable. So doing a `P_AutoSwitchWeapon()` would require running a variable through, which at that point idk if it's worth the hassle.~~

~~That is unless you were thinking of replacing all autoswitch cases with a function... although the conditions around each autoswitch seem quite variable, so it doesn't seem like a major boon.~~

Actually regarding the brackets style, I'm fine with adding brackets. Although I usually tend to avoid using brackets for 1 line conditions, as I usually find it looks cleaner.

prboom2/src/p_inter.c
andrikpowell: So in my test, this does actually work correctly. (i.e. I tested putting `allow_incompatiblity` at the end) and it still worked as intended.

Lol I will admit it is probably better practice to put `()` around the conditions though, so I'll do it anyway. :P

prboom2/src/p_inter.c
andrikpowell: Ignore my thoughts on `P_AutoSwitchWeapon`... I changed my mind

andrikpowell:
Implementing `P_AutoSwitchWeapon` has simplified the code significantly. Ready for another review :)

Pedro-Beirao (approved):

Pedro-Beirao:
Merging. Thanks!

Issue #531, opened 2024-11-15
Windowed mode does not scale properly in software mode


ghost:
Windowed mode does not scale properly in software mode

Pedro-Beirao:
It scales properly for me in both software and opengl. Tested in Windows and MacOS.
What OS are you using? And please attach a video of this issue

Cooler-1:
Windows 10


https://github.com/user-attachments/assets/76283109-cf81-4367-b03a-2c00235a35d7

KernelLeak:
Would I be correct in saying that it's not so much the image not scaling properly, but the window changing it's size along with the resolution in OpenGL mode, but staying pretty much the same in software mode?

Pedro-Beirao:
Ok I was wrong, I can reproduce. Thanks for the report.
The window size changes, but it doenst change to the correct selected size.

I had tested it with big changes (like 1080p -> 200p), so it seemed fine at first glance

For example
200p -> 400p makes the windows slightly bigger, but not double the size

Pedro-Beirao:
Looking into the code, software is "correct". In order to not make a tiny (almost unusable) 200p window, it scales it to atleast 480p.
https://github.com/coelckers/prboom-plus/commit/3a578e6ec59c0e371a5a46cc179677341a7cd837

Note that you can still resize the window with your mouse to make it smaller.

I'll change OpenGL to do the same. Thanks for the report!

Cooler-1:
I use Magpie to scale for CRT emulation and the window not being pixel accurate breaks it.

Pedro-Beirao:
> I use Magpie to scale for CRT emulation and the window not being pixel accurate breaks it.

Thats an interesting use case. @rfomin, seems like a good reason to allow initial window size of 200p. Shall I change this?

rfomin:
> > I use Magpie to scale for CRT emulation and the window not being pixel accurate breaks it.
> 
> Thats an interesting use case. @rfomin, seems like a good reason to allow initial window size of 200p. Shall I change this?

Sure, I'm not against it.

Magpie seems good. We should probably add CRT emulation shaders at some point.

Pedro-Beirao:
Done in https://github.com/kraflab/dsda-doom/pull/563

PR #543, opened 2024-11-30
Window improvements

Pedro-Beirao:
- Window can be resized on mac now.
That comment is no longer accurate on new mac versions.

- Make sure initial scale is always >= 640x480 on OpenGL
This was done for software in https://github.com/coelckers/prboom-plus/commit/3a578e6ec59c0e371a5a46cc179677341a7cd837
Woof has a min size of 480p, in our case, we allow the user to manually reduce the size from 480p to the render size using the mouse
Addresses https://github.com/kraflab/dsda-doom/issues/531

About the "actualheight" code. Do we really want to preserve the aspect ratio problem of old VGA screens? It doesnt even work on opengl.
If someone wants it, then they should play around with the "aspect ratio" option.

### DIFF
--- a/prboom2/src/SDL/i_video.c
+++ b/prboom2/src/SDL/i_video.c
@@ -1275,13 +1275,7 @@ void I_UpdateVideoMode(void)
       init_flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
   }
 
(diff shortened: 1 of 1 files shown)


Pedro-Beirao:
I'll bring up that last question some other time.

PR #544, opened 2024-11-30
Add horizontal mouse wheel support

andrikpowell:
This was slightly more difficult than it should've been. Code is adapted from Woof, and so it works the same way.

Required a delay event only for mouse scroll actions, requiring a new type of input. Vertical and Horizontal scroll now use the same new input. (This is all done in the same way as Woof)

### DIFF
--- a/prboom2/src/SDL/i_video.c
+++ b/prboom2/src/SDL/i_video.c
@@ -361,33 +361,29 @@ static void I_GetEvent(void)
       case SDL_MOUSEWHEEL:
         if (mouse_enabled && window_focused)
         {
-          int button = -1;
+          int mouseb;
 
           if (Event->wheel.y > 0)
-          {
-            button = KEYD_MWHEELUP;
-          }
+            mouseb = KEYD_MWHEELUP;
           else if (Event->wheel.y < 0)
-          {
-            button = KEYD_MWHEELDOWN;
-          }
+            mouseb = KEYD_MWHEELDOWN;
           else if (Event->wheel.x < 0)
-          {
-            button = KEYD_MWHEELLEFT;
-          }
+            mouseb = KEYD_MWHEELLEFT;
           else if (Event->wheel.x > 0)
+            mouseb = KEYD_MWHEELRIGHT;
+          else
+            mouseb = 0;
+
+          if(mouseb)
           {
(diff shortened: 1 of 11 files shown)


rfomin:
I'm pretty sure we shouldn't allow using the mouse wheel for movement, it's considered cheating. Woof used to have a check for this: fabiangreffrath/woof#703 It was removed during the refactoring, and I didn't restore it, but Woof is not as strict as DSDA-Doom. We should at least ask on the speedrunning discord.

Also: https://github.com/kraflab/dsda-doom/blob/d9650296a46178015fc5f61022b1cfe40b7e8ab6/patch_notes/v0.17.md?plain=1#L12

andrikpowell:
> I'm pretty sure we shouldn't allow using the mouse wheel for movement, it's considered cheating.

I agree. Checking with the previous versions, the mousewheel up/down is disabled when it comes to movement or turning, and adding the new mouseb input type unintentionally removed that limitation.

I've since reverted the new mouse button type to use the traditional way input was dealt with. I can't seem to find the exact code that blocks the movement/turning specifically for the mousewheel, so this is my temp fix. On top of that DSDA has it's own input system that's quite a bit different from Woof, so going the Woof route may not be the best approach.

One thing we will be missing out on though is the mouse buttons having a delay tied to them. The most easiest way to see this was in movement/turning (even though we don't want this, it's a good test to see if the delay input is working as it should).

The best example of this is binding the automap zoom in/out to the horizontal scroll, as without the delay it is choppy instead of a short smooth gradual step. Although then again, I don't think there's many key actions where you would notice a difference between a gradual or instant input, besides movement/turning.

Pedro-Beirao:
This is okay to merge right? It simply expands the current mouse wheel implementation

rfomin:
> This is okay to merge right? It simply expands the current mouse wheel implementation

I guess it's okay. I don't know DSDA rules about mouse wheel, what actions are allowed? I tried to bind the mouse wheel to different actions in "Key Bindings" and most of them don't work, looks confusing.

Pedro-Beirao:
There was a big disussion about this ban in the discord in 2023, but I cant seem to find out exactly what actions are allowed.
Anyway, lets merge, and Thanks!

PR #545, opened 2024-11-30
Limit keys that can call the menu from the titlepic

Pedro-Beirao:
Having any key bring up the menu can cause problem if say, that key is used for speedup.
Instead of leaving it up to the user to not acidentaly press a sensitive key when trying to open the menu, lets just limit it to the keys that make the most sense.

This also reorganizes the order of some conditions to allow input to reach certain actions


rfomin (approved):
LGTM.

Issue #467, opened 2023-12-27
Heretic's and Hexen's titlepic doesn't stretch


mooreye:
Steps to reproduce:
- have the setting `Status bar and menu appearance: fit to width`
- run Heretic or Hexen
- start a new game
- press the end game key to return to title menu
- see that the titlepic doesn't stretch to fill horizontal space, instead part of the screen from the ended game remains and fills the sides, see pic:

![2023-12-27T23:49:41,713497501+01:00](https://github.com/kraflab/dsda-doom/assets/93423596/234cf45d-4a58-42fe-b257-e62dcde01f45)

- Doom games don't have this issue, their titlepic stretches to the sides.

PR #546, opened 2024-12-01
Raven RawScreen Aspect Fix

andrikpowell:
So a thing that's bothered me for quite a while is that the "RawScreen"s that Raven uses have only been rendered using the "Not Adjusted" stretch setting... regardless of the stretch settings.

This PR fixes that and the issue mentioned here (#467). They are both related to the same issue in different ways. The problem with the specific issue, was that the "RawScreen" under "Fit to Width" was supposed to fit to width (imagine that), but code was set to prevent that.

Now the "RawScreen"s will look as they should for all three settings: "Not Adjusted", "Doom Format", and "Fit to Width". I will mention that the "RawScreen" code seems a bit WIP, so eventually there may be a better way to approach this further down the line.

One of the reasons the format was previously forced was probably due to the Raven Statusbar not being rendered properly under "Not Adjusted"... But the "RawScreen"s don't have that issue.

I haven't been able to fix the Raven Statusbar for "Not Adjusted", so I have it using the "Doom Format" stretch settings (just as DSDA Doom currently does).

### DIFF
--- a/prboom2/src/dsda/stretch.c
+++ b/prboom2/src/dsda/stretch.c
@@ -158,7 +158,7 @@ void dsda_ResetExTextOffsets(void) {
 static void InitStretchParam(stretch_param_t* offsets, int stretch, enum patch_translation_e flags) {
   memset(offsets, 0, sizeof(*offsets));
 
-  switch (stretch) {
+  switch (stretch_hud(stretch)) {
     case patch_stretch_not_adjusted:
       if (flags == VPT_ALIGN_WIDE) {
         offsets->video = &video_stretch;
@@ -271,10 +271,7 @@ void dsda_EvaluatePatchScale(void) {
       SCREENHEIGHT < 200 || WIDE_SCREENHEIGHT < 200)
     render_stretch_hud = patch_stretch_fit_to_width;
 
-  if (raven && render_stretch_hud == 0)
-    render_stretch_hud++;
-
-  switch (render_stretch_hud) {
+  switch (stretch_hud(render_stretch_hud)) {
     case patch_stretch_not_adjusted:
       wide_offset2x = SCREENWIDTH - patches_scalex * 320;
       wide_offset2y = SCREENHEIGHT - patches_scaley * 200;
--- a/prboom2/src/st_stuff.c
+++ b/prboom2/src/st_stuff.c
@@ -403,7 +403,7 @@ void ST_SetScaledWidth(void)
   if (width == 0)
       width = ST_WIDTH;
 
-  switch (render_stretch_hud)
+  switch (stretch_hud(render_stretch_hud))
   {
     case patch_stretch_not_adjusted:
       ST_SCALED_WIDTH  = width * patches_scalex;
--- a/prboom2/src/v_video.c
+++ b/prboom2/src/v_video.c
@@ -1417,7 +1417,7 @@ void V_DrawRawScreen(const char *lump_name)
 void V_DrawRawScreenSection(const char *lump_name, int source_offset, int dest_y_offset, int dest_y_limit)
 {
   int i, j;
-  float x_factor, y_factor;
+  float x_factor = 0, y_factor = 0;
   int x_offset, y_offset;
   const byte* raw;
 
@@ -1440,11 +1440,24 @@ void V_DrawRawScreenSection(const char *lump_name, int source_offset, int dest_y
     }
   }
 
-  x_factor = (float)SCREENWIDTH / 320;
-  y_factor = (float)SCREENHEIGHT / 200;
-
-  if (y_factor < x_factor)
-    x_factor = y_factor;
+  switch (render_stretch_hud) {
+    case patch_stretch_not_adjusted:
+      x_factor = (float)SCREENWIDTH / 320;
+      y_factor = (float)SCREENHEIGHT / 200;
+      if (y_factor < x_factor)
+        x_factor = y_factor;
+      break;
+    case patch_stretch_doom_format:
+      x_factor = (float)WIDE_SCREENWIDTH / 320;
+      y_factor = (float)WIDE_SCREENHEIGHT / 200;
+      if (y_factor < x_factor)
+        x_factor = y_factor;
+      break;
+    case patch_stretch_fit_to_width:
+      x_factor = (float)SCREENWIDTH / 320;
+      y_factor = (float)SCREENHEIGHT / 200;
+      break;
+  }
 
   x_offset = (int)((SCREENWIDTH - (x_factor * 320)) / 2);
   y_offset = (int)((dest_y_offset * y_factor) - (source_offset * y_factor / 320));
--- a/prboom2/src/dsda/stretch.h
+++ b/prboom2/src/dsda/stretch.h
@@ -51,6 +51,10 @@ typedef enum
   patch_stretch_max
 } patch_stretch_t;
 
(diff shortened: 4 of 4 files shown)


prboom2/src/dsda/stretch.c
```diff
@@ -293,6 +299,14 @@ void dsda_EvaluatePatchScale(void) {
...
   }
 
+  // Raven HUD breaks in "patch_stretch_not_adjusted",
+  // so let's use "patch_stretch_doom_format" settings
+  if (raven && render_stretch_hud == 0) {
+      ST_SCALED_HEIGHT = g_st_height * WIDE_SCREENHEIGHT / 200;
+      wide_offset2x = SCREENWIDTH - WIDE_SCREENWIDTH;
+      wide_offset2y = SCREENHEIGHT - WIDE_SCREENHEIGHT;
+  }
+
```
Pedro-Beirao: Why did you make this change? It does the exact same thing
Althought instead of `render_stretch_hud++`, you could do `render_stretch_hud = patch_stretch_doom_format`

prboom2/src/dsda/stretch.c
```diff
@@ -183,6 +183,15 @@ static void InitStretchParam(stretch_param_t* offsets, int stretch, enum patch_t
...
 
+  // Raven HUD breaks in "patch_stretch_not_adjusted",
+  // so let's use "patch_stretch_doom_format" settings
+  if (raven && stretch == 0)
+  {
+    offsets->video = &video_stretch;
+    offsets->deltax1 = (SCREENWIDTH - WIDE_SCREENWIDTH) / 2;
+    offsets->deltax2 = (SCREENWIDTH - WIDE_SCREENWIDTH) / 2;
+  }
+
```
Pedro-Beirao: Do like the code below. Put an `if` before the `switch` statement that sets `stretch`to `patch_stretch_doom_format` if necessary

prboom2/src/st_stuff.c
```diff
@@ -416,6 +416,11 @@ void ST_SetScaledWidth(void)
       break;
   }
 
+  // Raven HUD breaks in "patch_stretch_not_adjusted",
+  // so let's use "patch_stretch_doom_format" settings
+  if (raven && render_stretch_hud == 0)
+      ST_SCALED_WIDTH  = width * WIDE_SCREENWIDTH / 320;
+
```
Pedro-Beirao: Same here

prboom2/src/v_video.c
```diff
@@ -1440,11 +1440,29 @@ void V_DrawRawScreenSection(const char *lump_name, int source_offset, int dest_y
...
+    case patch_stretch_fit_to_width:
+      x_factor = (float)SCREENWIDTH / 320;
+      y_factor = (float)SCREENHEIGHT / 200;
+      break;
+    default:
+      // This should never happen
+      x_factor = 0;
+      y_factor = 0;
+      break;
+  }
```
Pedro-Beirao, line 1460: No need for the `default` path

prboom2/src/dsda/stretch.c
andrikpowell: Original code changes every game asset that's "not adjusted" to become "doom format".

My code only does that to the statusbar as the statusbar doesn't display properly in "not adjusted".

But my code allows all other assets to be able to render as "not adjusted", which previously was not the case.

prboom2/src/dsda/stretch.c
andrikpowell: My thoughts process here is that this is a temporary fix, and by keeping this code isolated, it will be easier to cull once we figure out how to render the statusbar correctly in "not adjusted".

By putting an `if` in the `switch`, we are solidifying temp code that shouldn't be final.

prboom2/src/v_video.c
andrikpowell: Only reason I added a default path was to fix GCC warnings.

Since I've started working on DSDA Doom code, it's been special in that GCC had no warnings in the continuous integration, and so whenever I work with the code, I try and keep it that way.

prboom2/src/dsda/stretch.c
Pedro-Beirao: Oh I see. The old code permanently changes the setting.
Missed that 😅

Pedro-Beirao: But instead of doing `== 0`, do `== patch_stretch_not_adjusted`for readability

prboom2/src/v_video.c
Pedro-Beirao: All other instances of this switch statement are not using `default`. Thats why the comment.
And idk why the CI would not give you any warning, there are tons of warnings on my machine

andrikpowell: What I'm particularly referencing is the GCC Linux and MSYS2 build log from the GitHub continuous integration.

When I started working on DSDA code these showed absolutely no warnings during the build process and so I tend to try and keep it that way.

I agree I wasn't crazy about using `default` there, I was just finding a way to squash a possible undefined warning relating to `x_factor` and `y_factor`.

andrikpowell:
Is this better?

Pedro-Beirao (approved):

fabiangreffrath:
Most of these changes appear redundant with code that is alrady there. It would be easier to implement a function (or macro) that returns `(raven && stretch == patch_stretch_not_adjusted) ? patch_stretch_doom_format : stretch` and use this in the `switch` statements instead of redefining in a later paragraph what was already checked before. This would also hold up the "temporary" aspect of these changes, as that function could later be changed to just return `stretch` (or get removed entirely).

andrikpowell:
> Most of these changes appear redundant with code that is alrady there. It would be easier to implement a function (or macro) that returns `(raven && stretch == patch_stretch_not_adjusted) ? patch_stretch_doom_format : stretch` and use this in the `switch` statements instead of redefining in a later paragraph what was already checked before. This would also hold up the "temporary" aspect of these changes, as that function could later be changed to just return `stretch` (or get removed entirely).

I think this is close to what you were suggesting? (check force push)

fabiangreffrath:
> I think this is close to what you were suggesting? (check force push)

Yes, except that it now uses the global `render_stretch_hud` in `InitStretchParam()` instead of the passed `stetch` parameter.

andrikpowell:
> Yes, except that it now uses the global `render_stretch_hud` in `InitStretchParam()` instead of the passed `stetch` parameter.

How do you wanna get around that?

Because the only way to force that `switch` statement to use `patch_stretch_doom_format` is to not use the passed `stretch` parameter.

I guess we could have a `if (stretch_hud != stretch)` then `stretch == stretch_hud` above?

fabiangreffrath:
> How do you wanna get around that?

```c
#define stretch_hud(a) ((raven && (a) == patch_stretch_not_adjusted) ? patch_stretch_doom_format : (a))
```

andrikpowell:
> > How do you wanna get around that?
> 
> ```c
> #define stretch_hud(a) ((raven && (a) == patch_stretch_not_adjusted) ? patch_stretch_doom_format : (a))
> ```

oh fascinating! This is actually the first time I've seen a `#define` statement with arguments! That's actually really cool.

Force pushed new version with that included.

fabiangreffrath (approved):
Looks good now, thank you!

Pedro-Beirao:
Thank you both!

PR #547, opened 2024-12-02
Implement dark overlay for menus and automap

andrikpowell:
This PR adds the dark overlay that Woof has.

I've implemented it via `V_DrawShaded` which gets directed to two different functions depending on OpenGL or Software (this is how most render functions currently work in DSDA Doom)

The `V_DrawShaded` function includes a gradual fade option (`dboolean animate`) that's used in Nyan Doom, although I've set it to false in both instances that it's used here to match Woof's implementation.

The automap now has 3 overlay settings in the same way Woof does: "No Overlay", "Overlay 1" (what DSDA already had), and "Overlay dark" which adds a translucent background. The dark overlay is also compatible with the DSDA Doom Minimap.

~The final commit here is probably the most controversial, but I wanted to bring it up here. It's because of this commit that implementing the dark overlay was actually harder to get working for DSDA Doom compared to Nyan Doom. In order for Nyan Doom to support overlapping HUD icons and statusbar animations, I have the statusbar always refresh in Software just as it does in OpenGL. In my testing I haven't run into issues or slowdown by doing this, and it fixes many statusbar bleed issues with PWADs such as "Pirate Doom II" and "200 Line Christmas".~

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -1005,6 +1005,18 @@ dboolean AM_Responder
 {
   static int bigstate=0;
 
+  if (dsda_InputActivated(dsda_input_map_overlay) && (automap_input || dsda_ShowMinimap()))
+  {
+    dsda_CycleConfig(dsda_config_automap_overlay, true);
+    dsda_AddMessage(automap_overlay == 0 ? s_AMSTR_OVERLAYOFF :
+                    automap_overlay == 1 ? s_AMSTR_OVERLAYON :
+                    "Overlay Mode Dark");
+    AM_SetPosition();
+    AM_activateNewScale();
+
+    return true;
+  }
+
   if (!automap_input)
   {
     if (dsda_InputActivated(dsda_input_map))
@@ -1143,17 +1155,6 @@ dboolean AM_Responder
   {
     dsda_ToggleConfig(dsda_config_automap_rotate, true);
     dsda_AddMessage(automap_rotate ? s_AMSTR_ROTATEON : s_AMSTR_ROTATEOFF);
-
-    return true;
-  }
-  else if (dsda_InputActivated(dsda_input_map_overlay))
-  {
-    dsda_CycleConfig(dsda_config_automap_overlay, true);
-    dsda_AddMessage(automap_overlay == 0 ? s_AMSTR_OVERLAYOFF :
-                    automap_overlay == 1 ? s_AMSTR_OVERLAYON :
-                    "Overlay Mode Dark");
-    AM_SetPosition();
-    AM_activateNewScale();
 
     return true;
   }
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -57,6 +57,7 @@
 #include "m_misc.h"
 #include "m_bbox.h"
 #include "d_main.h"
+#include "m_menu.h"
 
 #include "dsda/id_list.h"
 #include "dsda/input.h"
@@ -616,7 +617,7 @@ void AM_SetPosition(void)
     f_x = f_y = 0;
     f_w = SCREENWIDTH;
 
-    if (automap_overlay)
+    if (automap_overlay > 0)
     {
       f_h = viewheight;
     }
@@ -1147,7 +1148,10 @@ dboolean AM_Responder
   }
   else if (dsda_InputActivated(dsda_input_map_overlay))
   {
-    dsda_ToggleConfig(dsda_config_automap_overlay, true);
+    dsda_CycleConfig(dsda_config_automap_overlay, true);
+    dsda_AddMessage(automap_overlay == 0 ? s_AMSTR_OVERLAYOFF :
+                    automap_overlay == 1 ? s_AMSTR_OVERLAYON :
+                    "Overlay Mode Dark");
     AM_SetPosition();
     AM_activateNewScale();
 
@@ -2769,6 +2773,9 @@ void AM_Drawer (dboolean minimap)
   if (!automap_active && !minimap)
     return;
 
+  if (automap_active && automap_overlay == 2 && minimap)
+    return;
+
   V_BeginAutomapDraw();
 
   if (automap_follow)
@@ -2792,6 +2799,8 @@ void AM_Drawer (dboolean minimap)
 
   if (!automap_overlay) // cph - If not overlay mode, clear background for the automap
     V_FillRect(FB, f_x, f_y, f_w, f_h, (byte)mapcolor_p->back); //jff 1/5/98 background default color
+  if (automap_overlay == 2 && !M_MenuIsShaded())
+    V_DrawShaded(FB, f_x, f_y, f_w, f_h, FULLSHADE);
 
   if (map_textured)
   {
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -160,6 +160,25 @@
 
 extern dboolean  message_dontfuckwithme;
 
+/////////////////////////////
+//
+// booleans for setup screens
+// these tell you what state the setup screens are in, and whether any of
+// the overlay screens (automap colors, reset button message) should be
+// displayed
+
+dboolean setup_active      = false; // in one of the setup screens
+dboolean set_keybnd_active = false; // in key binding setup screens
+dboolean set_weapon_active = false; // in weapons setup screen
+dboolean set_status_active = false; // in status bar/hud setup screen
+dboolean set_auto_active   = false; // in automap setup screen
+dboolean setup_select      = false; // changing an item
+dboolean setup_gather      = false; // gathering keys for value
+dboolean colorbox_active   = false; // color palette being shown
+dboolean set_general_active = false;
+dboolean level_table_active = false;
+
+
 extern const char* g_menu_flat;
 extern int g_menu_save_page_size;
 extern int g_menu_font_spacing;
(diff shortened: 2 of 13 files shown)


andrikpowell:
I've done a force-push as I found an issue regarding the overlay menu and STBAR... However this force push fixes all the STBAR issues for Doom and Raven.

andrikpowell:
> The final commit here is probably the most controversial, but I wanted to bring it up here. It's because of this commit that implementing the dark overlay was actually harder to get working for DSDA Doom compared to Nyan Doom. In order for Nyan Doom to support overlapping HUD icons and statusbar animations, I have the statusbar always refresh in Software just as it does in OpenGL. In my testing I haven't run into issues or slowdown by doing this, and it fixes many statusbar bleed issues with PWADs such as "Pirate Doom II" and "200 Line Christmas".

I've decided to extract this and place it into another pull request... It's not that related to this PR, and is honestly a separate issue altogether.

rfomin:
Set to OpenGL mode, go to the general menu and wait a few minutes - the background will start to flicker. On my system it seems to happen more often with VSync turned off.

Pedro-Beirao:
So the map overlay keybind cycles thru these 3 options. I dont like that since I use overlay when playing, but when I want to check the map name, coords, etc, I switch to no overlay.

This will make me have to cycle thru 1 more option. Its one of those small changes that affect the players in annoying ways. I think it would be best to have an option in the automap menu to decide if the overlay is faded or not.

Pedro-Beirao:
[Bug] Being in the faded menu, but then pressing CMD+Q (Mac keybind to bring up the "Are u sure u want to quit" prompt)and then cancelling: keeps the titlepic faded, and is only fixed after pressing Esc to bring up the menu again

This doesnt happen with the Texture option

Pedro-Beirao:
Can you remove the animate logic altogether? Since its always false + it looks better with no fade

andrikpowell:
I've removed *(or extracted for Nyan Doom)* the overlay fade logic. Point is it's now gone for DSDA Doom.

In addition, I've fixed the Mac "CMD + Q" overlay issue.

I spent a full day trying another approach for the overlay logic, but it wasn't very compatible with disabling itself when the automap overlay is present, so I think the current way is better... I really don't wanna talk about how long I spent on this...

I plan on moving onto the revising the automap logic next :)

andrikpowell:
Ready for automap dark overlay review

prboom2/src/am_map.c
```diff
@@ -616,7 +617,7 @@ void AM_SetPosition(void)
     f_x = f_y = 0;
     f_w = SCREENWIDTH;
 
-    if (automap_overlay)
+    if (automap_overlay>0)
```
rfomin: Sorry to be picky, but I think we should try to use a uniform style for the new code: `if (automap_overlay > 0)` with spaces.

The original PrBoom+ was badly formatted, but all the new code from Kraflab has the same style (files in the dsda folder).

prboom2/src/m_menu.c
```diff
@@ -5823,6 +5828,31 @@ void M_StartControlPanel (void)
...
 }
 
+
+/////////////////////////////////////////////////////////////////////////////
+//
+// Menu Shaded Overlay Stuff
+//
+// This displays a dark overlay under certain screens of the menus
+
+int screenshade = 20;
```
rfomin: We should use `static const` or macro if it is a constant. Do we need this variable?

prboom2/src/m_menu.c
```diff
@@ -1189,81 +1189,6 @@ void M_Options(int choice)
   M_SetupNextMenu(&OptionsDef);
 }
 
-/////////////////////////////
-//
-// M_QuitDOOM
-//
```
rfomin: Why did you move this code?

rfomin (commented):
It works well from my testing, thank you!

prboom2/src/am_map.c
andrikpowell: Easy enough to fix 👍

prboom2/src/m_menu.c
andrikpowell: So in Nyan Doom, I have extra code where this variable is not constant... And so I was trying to keep the code close to the same between the two ports, while deleting extra animation fade code for DSDA Doom.

prboom2/src/m_menu.c
andrikpowell: Beside moving this code, I've added the two lines:
```
setup_active = false;
currentMenu = NULL;
```
This is to avoid the overlay getting applied to the quit message when sending the SDL_Quit response while the overlay is open. The `M_QuitDOOM` code needed to be moved down below the `setup_active` declaration to change the variable status.

rfomin: It's easier to move the `setup_active` declaration to the beginning of the file. If you move a big chunk of code, we can't see in `diff` what was changed in it.

prboom2/src/m_menu.c
rfomin: Are there plans to merge code from Nyan into DSDA-Doom or vice versa?

andrikpowell: No, but isn't it better to keep the implementation of the same feature in both source ports as close as possible? As in if I add something to Nyan that DSDA wants in the future or vice-versa, it will be easier if DSDA is already using a very similar structure.

I mention this because there are certain things I can "optimise / simplify", but I haven't because I wanted to keep the structure relatively close to Nyan for consistency sake.

prboom2/src/m_menu.c
andrikpowell: Decided to move back `M_QuitDOOM` and move the menu active declarations to the top of the file. I'd rather not separate them, as the comments already have them grouped together.

prboom2/src/am_map.c
andrikpowell: Should be resolved in latest commit

prboom2/src/m_menu.c
```diff
@@ -5823,6 +5828,31 @@ void M_StartControlPanel (void)
...
+
+dboolean M_MenuIsShaded(void)
+{
+  int Options = (setup_active || currentMenu == &OptionsDef || currentMenu == &SoundDef);
+  return fadeBG() && Options;
+}
+
+static void M_ShadedScreen(int scrn)
+{
+  V_DrawShaded(scrn, 0, 0, SCREENWIDTH, SCREENHEIGHT, screenshade);
```
rfomin: In general, it is good practice to reduce the number of global variables as much as possible. Doom is not good at this, but let's not make it worse. Let's remove the global variable `screenshade`, but keep the function parameter `screenshade`, then the difference with Nyan Doom would be minimal:
```diff
diff --git a/prboom2/src/am_map.c b/prboom2/src/am_map.c
index 3650a72c9..eeda64153 100644
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -2801,7 +2801,7 @@ void AM_Drawer (dboolean minimap)
   if (!automap_overlay) // cph - If not overlay mode, clear background for the automap
     V_FillRect(FB, f_x, f_y, f_w, f_h, (byte)mapcolor_p->back); //jff 1/5/98 background default color
   if (automap_overlay == 2 && !M_MenuIsShaded())
-    V_DrawShaded(FB, f_x, f_y, f_w, f_h, screenshade);
+    V_DrawShaded(FB, f_x, f_y, f_w, f_h, SCREENSHADE);
 
   if (map_textured)
   {
diff --git a/prboom2/src/m_menu.c b/prboom2/src/m_menu.c
index ce58856f3..0cbccddf4 100644
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -5835,8 +5835,6 @@ void M_StartControlPanel (void)
 //
 // This displays a dark overlay under certain screens of the menus
 
-int screenshade = 20;
-
 dboolean fadeBG(void)
 {
   return dsda_IntConfig(dsda_config_menu_background) == 1;
@@ -5850,7 +5848,7 @@ dboolean M_MenuIsShaded(void)
 
 static void M_ShadedScreen(int scrn)
 {
-  V_DrawShaded(scrn, 0, 0, SCREENWIDTH, SCREENHEIGHT, screenshade);
+  V_DrawShaded(scrn, 0, 0, SCREENWIDTH, SCREENHEIGHT, SCREENSHADE);
 }
 
 //
diff --git a/prboom2/src/m_menu.h b/prboom2/src/m_menu.h
index 2048b4000..a69bc267c 100644
--- a/prboom2/src/m_menu.h
+++ b/prboom2/src/m_menu.h
@@ -51,7 +51,7 @@ dboolean M_Responder (event_t *ev);
 
 dboolean fadeBG(void);
 dboolean M_MenuIsShaded(void);
-extern int screenshade;
+#define SCREENSHADE 20
 
 // Called by main loop,
 // only used for menu (skull cursor) animation.
```

andrikpowell: So in Nyan Doom, `screenshade` is a variable that changes due to some gradual fade logic. To keep this consistent between both DSDA and Nyan, I think I'll replace `int screenshade` here with a `#define FULLSHADE 20` which Nyan uses to compare `screenshade` to.

rfomin (approved):
LGTM, thank you.

andrikpowell:
Ok, fixed up PR and simplified commits 👍

Pedro-Beirao:
~~Please make Textured be the default mode for Menu Background. Like it was before~~

And I see that Textured only "kicks in" when going into sub-menus of Options, while Dark is straight when going into Options. Its inconsistent, but I kind of like it like this (and woof does the same). So no need to change this, just wanted to note it

Pedro-Beirao (approved):

Pedro-Beirao:
Thanks!

PR #549, opened 2024-12-07
IDMUS cheat support UMAPINFO

Pedro-Beirao:
Reported on discord. Fixes IDMUS only playing IWAD lumps

The dsda_*Music() functions were hardcoded to only use the current map, so I adapted dsda_MapMusic() to take any map.
Maybe we should do the same for the others, but I can't think of a reason we would need to call them with a map param.

### DIFF
--- a/prboom2/src/dsda/mapinfo/u.c
+++ b/prboom2/src/dsda/mapinfo/u.c
@@ -170,16 +170,15 @@ int dsda_UMusicIndexToLumpNum(int* lump, int music_index) {
   return false;
 }
 
-int dsda_UMapMusic(int* music_index, int* music_lump) {
+int dsda_UMapMusic(int* music_index, int* music_lump, int episode, int map) {
   int lump;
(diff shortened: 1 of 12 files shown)


prboom2/src/dsda/mapinfo/u.c
```diff
@@ -170,16 +170,15 @@ int dsda_UMusicIndexToLumpNum(int* lump, int music_index) {
   return false;
 }
 
-int dsda_UMapMusic(int* music_index, int* music_lump) {
+int dsda_UMapMusic(int* music_index, int* music_lump, int episode, int map) {
   int lump;
+  struct MapEntry* entry = dsda_UMapEntry(episode, map);
+  int level_num;
```
fabiangreffrath: Do we need `int level_num` here?

Pedro-Beirao: We don't. I missed that.

fabiangreffrath (approved):
LGTM, thanks!

PR #554, opened 2024-12-11
Update svg icon to best practice standards

fpiesche:
The SVG icon in its current form does not actually follow [best practice recommendations](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines/#app-icon) for icon design, and so using it in e.g. a Flatpak build (which performs some linting on the metadata such as icons, `.desktop` files etc) will cause the build to fail:

- the canvas isn't square
- the icon fills too much of the canvas

Additionally, the svg file is unoptimized and contains a fair amount of unnecessary data. Simply loading it in Inkscape and re-saving as an Optimized SVG reduces the file size by around 30kB.

I've fixed all of these issues - the svg is now placed on a 1024x1024 canvas, centered and slightly shrunken to leave some border around it, and saved as an Optimized SVG (which applies [Scour](https://github.com/scour-project/scour) to the output file to scrub unnecessary data). No actual changes were made to the artwork.


fabiangreffrath (approved):
LGTM, thank you!

Pedro-Beirao:
Thanks!

PR #556, opened 2024-12-12
consider the (rare) case that $HOME is unset

fabiangreffrath:


Pedro-Beirao (approved):
Didnt know it was possible for it to not be set

fabiangreffrath:
I'd call it a broken configuration, but dsda-doom shouldn't crash in that case.

PR #558, opened 2024-12-14
Allow to run in a read-only environment

fabiangreffrath:


Pedro-Beirao (approved):
Interesting use case. Thanks!

PR #561, opened 2024-12-21
Fix undefined behaviour

suve:
The MAPINFO parser has a couple of places where data is put into an `std::vector`, and then a pointer to the first element of the vector is taken through the `&vector[0]` pattern. Unfortunately, using `operator[]` to access a non-existent element of `std::vector` is undefined behaviour[^0]; which means that, should the vector be empty, then nasal demons galore.

This commit adds "is the vector empty?" checks in relevant places.

[^0]: https://en.cppreference.com/w/cpp/container/vector/operator_at


fabiangreffrath (approved):
Looks reasonable.

suve:
An alternative solution could be to use the [.data()](https://en.cppreference.com/w/cpp/container/vector/data) method, which returns a pointer to the underlying array. Note that the returned pointer is **not** guaranteed to be NULL if the vector is empty.

Pedro-Beirao:
Thanks!

PR #562, opened 2024-12-22
Fallback to getpwuid() if $HOME is unset

suve:
Recently, commit 7621899c4025b4b4af811ed81ce0a08eca2b6a6e added some fallback code for the rare case when `$HOME` is unset. However, that code simply falls to using the root directory as the home. This PR makes the game use the `getpwuid()`[^0] function instead, which retrieves information about the user profile - including the home directory - from the OS.

One minor issue with this PR is that (at least on Fedora Linux) it causes some "duplicate definition" warnings:
```
prboom2/BUILD/config.h:18:9: warning: "HAVE_SYS_TYPES_H" redefined
   18 | #define HAVE_SYS_TYPES_H
      |         ^~~~~~~~~~~~~~~~
In file included from /usr/include/SDL2/SDL_config.h:58,
                 from /usr/include/SDL2/SDL_stdinc.h:31,
                 from /usr/include/SDL2/SDL_main.h:25,
                 from /usr/include/SDL2/SDL.h:32,
                 from prboom2/src/SDL/i_sound.c:43:
/usr/include/SDL2/SDL_config-x86_64.h:73:9: note: this is the location of the previous definition
```
There are a couple of different ways this could be solved:

1. Hide the `#cmakedefine` behind an `#ifdef`, so it only defines the symbol if not already present.

2. Move the "are all three headers present?" check to CMakeLists and then have only `#cmakedefine GETPWUID_FALLBACK` in the config file.

3. Probably something else I haven't considered.

Each has its pros and cons, so I thought it's best to discuss first.

[^0]: https://linux.die.net/man/3/getpwuid

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -56,6 +56,10 @@
 
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
+#ifdef HAVE_GETPWUID
+#include <sys/types.h>
+#include <pwd.h>
+#endif
 #endif
 
 #ifdef _MSC_VER
@@ -315,7 +319,13 @@ const char *I_ConfigDir(void)
     const char *home = M_getenv("HOME");
     if (!home)
     {
-      home = "/";
+#ifdef HAVE_GETPWUID
+      struct passwd *user_info = getpwuid(getuid());
+      if (user_info != NULL)
+        home = user_info->pw_dir;
+      else
+#endif
+        home = "/";
     }
 
     // First, try legacy directory.
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -91,6 +91,7 @@ check_symbol_exists(mmap "sys/mman.h" HAVE_MMAP)
 check_symbol_exists(CreateFileMapping "windows.h" HAVE_CREATE_FILE_MAPPING)
 check_symbol_exists(strsignal "string.h" HAVE_STRSIGNAL)
 check_symbol_exists(mkstemp "stdlib.h" HAVE_MKSTEMP)
+check_symbol_exists(getpwuid "unistd.h;sys/types.h;pwd.h" HAVE_GETPWUID)
 
 include(CheckIncludeFile)
 
--- a/prboom2/cmake/config.h.cin
+++ b/prboom2/cmake/config.h.cin
@@ -14,6 +14,7 @@
 #cmakedefine HAVE_CREATE_FILE_MAPPING
 #cmakedefine HAVE_STRSIGNAL
 #cmakedefine HAVE_MKSTEMP
+#cmakedefine HAVE_GETPWUID
 
 #cmakedefine HAVE_SYS_WAIT_H
 #cmakedefine HAVE_UNISTD_H


fabiangreffrath:
Check for the symbol being available instead of the headers:

```patch
diff --git a/prboom2/CMakeLists.txt b/prboom2/CMakeLists.txt
index e6ce9a85e..0d60ad7c4 100644
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -91,13 +91,12 @@ check_symbol_exists(mmap "sys/mman.h" HAVE_MMAP)
 check_symbol_exists(CreateFileMapping "windows.h" HAVE_CREATE_FILE_MAPPING)
 check_symbol_exists(strsignal "string.h" HAVE_STRSIGNAL)
 check_symbol_exists(mkstemp "stdlib.h" HAVE_MKSTEMP)
+check_symbol_exists(getpwuid "unistd.h;sys/types.h;pwd.h" HAVE_GETPWUID)
 
 include(CheckIncludeFile)
 
-check_include_file("sys/types.h" HAVE_SYS_TYPES_H)
 check_include_file("sys/wait.h" HAVE_SYS_WAIT_H)
 check_include_file("unistd.h" HAVE_UNISTD_H)
-check_include_file("pwd.h" HAVE_PWD_H)
 check_include_file("asm/byteorder.h" HAVE_ASM_BYTEORDER_H)
 check_include_file("dirent.h" HAVE_DIRENT_H)
 
diff --git a/prboom2/cmake/config.h.cin b/prboom2/cmake/config.h.cin
index b25f09bb2..f24fb1e52 100644
--- a/prboom2/cmake/config.h.cin
+++ b/prboom2/cmake/config.h.cin
@@ -14,11 +14,10 @@
 #cmakedefine HAVE_CREATE_FILE_MAPPING
 #cmakedefine HAVE_STRSIGNAL
 #cmakedefine HAVE_MKSTEMP
+#cmakedefine HAVE_GETPWUID
 
-#cmakedefine HAVE_SYS_TYPES_H
 #cmakedefine HAVE_SYS_WAIT_H
 #cmakedefine HAVE_UNISTD_H
-#cmakedefine HAVE_PWD_H
 #cmakedefine HAVE_ASM_BYTEORDER_H
 #cmakedefine HAVE_DIRENT_H
 
diff --git a/prboom2/src/SDL/i_system.c b/prboom2/src/SDL/i_system.c
index be6e4f1de..323aa82b4 100644
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -56,12 +56,9 @@
 
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
-#ifdef HAVE_PWD_H
-#ifdef HAVE_SYS_TYPES_H
-#include <pwd.h>
+#ifdef HAVE_GETPWUID
 #include <sys/types.h>
-#define GETPWUID_FALLBACK
-#endif
+#include <pwd.h>
 #endif
 #endif
 
@@ -324,7 +321,7 @@ const char *I_ConfigDir(void)
     {
       home = "/";
 
-#ifdef GETPWUID_FALLBACK
+#ifdef HAVE_GETPWUID
       struct passwd *user_info = getpwuid(getuid());
       if (user_info != NULL) {
         home = user_info->pw_dir;
```

fabiangreffrath:
Also:

```patch
diff --git a/prboom2/src/SDL/i_system.c b/prboom2/src/SDL/i_system.c
index 323aa82b4..9d70c8c94 100644
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -319,14 +319,13 @@ const char *I_ConfigDir(void)
     const char *home = M_getenv("HOME");
     if (!home)
     {
-      home = "/";
-
 #ifdef HAVE_GETPWUID
       struct passwd *user_info = getpwuid(getuid());
-      if (user_info != NULL) {
+      if (user_info != NULL)
         home = user_info->pw_dir;
-      }
+      else
 #endif
+        home = "/";
     }
 
     // First, try legacy directory.
```

suve:
> Check for the symbol being available instead of the headers

I somehow completely missed the `check_symbol_exists()` call above. Think that's called tunnel vision...

> Also:

Heh, that was how I wrote it originally, but I know a lot of people are very much against splitting if-else blocks across conditional compilation boundaries, so I changed it.

Thanks for the suggestions; I've amended the commit to include them.

prboom2/src/SDL/i_system.c
```diff
@@ -315,7 +319,13 @@ const char *I_ConfigDir(void)
     const char *home = M_getenv("HOME");
     if (!home)
     {
-      home = "/";
+#ifdef HAVE_GETPWUID
+      struct passwd *user_info = getpwuid(getuid());
+      if (user_info != NULL)
+        home = user_info->pw_dir;
```
fabiangreffrath, line 325: Not sure if the struct pointer that `getpwuid()` returns remains static in memory. Any other process calling `getpwuid()` might cause its memory address to change, so it should be safer to call `strdup()` on the `pw_dir` string.

suve: According to the man page:
```
The return value may point to a static area, and may be overwritten
by subsequent calls to getpwent(3), getpwnam(), or getpwuid().
(Do not pass the returned pointer to free(3).)
```
Unless there are multiple threads running which may call this function, this should be safe. Once the `base` static variable is initialized, this won't get called again.

fabiangreffrath (approved):
LGTM, thank you!

Pedro-Beirao:
Thank you both!

PR #563, opened 2024-12-23
More window improvements

Pedro-Beirao:
- Allow < 640x480 windows https://github.com/kraflab/dsda-doom/issues/531
- Add low res aspect ratio correction on OpenGL
   On software 200p and 400p, the image is stretched to match 240p and 480p. This makes OpenGL follow it
   I meantioned before that I thought this behaviour should be removed instead, but Ive changed my opinions as I explored the issue more
- Allow setting desired_fullscreen in the menu, even if -window or -fullscreen
   If you had one of these params on, the desired_fullscreen menu option would not work
   Unlike these other params, -geom already allowed the menu option to work
   This is useful for the launcher, as it uses these params on every launch

### DIFF
--- a/prboom2/src/SDL/i_video.c
+++ b/prboom2/src/SDL/i_video.c
@@ -1223,7 +1223,6 @@ void I_UpdateVideoMode(void)
 {
   int init_flags = SDL_WINDOW_ALLOW_HIGHDPI;
   int screen_multiply;
-  int actualheight;
   int render_vsync;
   int integer_scaling;
   const char *sdl_video_window_pos;
@@ -1273,11 +1272,11 @@ void I_UpdateVideoMode(void)
   // [FG] aspect ratio correction for the canonical video modes
   if (SCREENHEIGHT == 200 || SCREENHEIGHT == 400)
   {
-    actualheight = 6*SCREENHEIGHT/5;
+    ACTUALHEIGHT = 6*SCREENHEIGHT/5;
   }
   else
   {
-    actualheight = SCREENHEIGHT;
+    ACTUALHEIGHT = SCREENHEIGHT;
   }
 
   if (desired_fullscreen)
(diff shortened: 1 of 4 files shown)


rfomin:
What is the difference between 400p and 480p modes? 480p is stretched in the engine, but 400p is stretched by SDL?

fabiangreffrath:
In my understanding, in 480p the vertical field of view is rendered to 480 pixels which are then transferred 1:1 to screen, whereas in 400p the vertical field of view is rendered to 400 pixels which are then stretched by factor 1.2:1 on their way to screen.

rfomin:
Why do we need both 480p and 400p modes? They look very similar. I think we should pick the ones that look better and cut the rest.

Pedro-Beirao:
This clearly shows that 200p is the correct resolution, but it should be rendered in a 240p window, due to https://doomwiki.org/wiki/Aspect_ratio

On 240p, some UI elements get screwed like this (only on software)
Screenshot 2024-12-23 at 18 50 44

On stretched 200p they look fine
Screenshot 2024-12-23 at 18 55 04

fabiangreffrath (approved):
LGTM, thanks!

TheGershon:
Unless I'm missing something obvious, this seems to have broken 200px (or any multiple of it) height, forcing 240px instead.
image
As you can see, this is clearly 240px tall despite being set to 426x200. The same happens at 640x400, forcing it to be 480 instead. It's now impossible to get perfect square pixels which can then be stretched externally. A quick rollback to 0.28.2 works as expected:
image

Pedro-Beirao:
@TheGershon This is the intended behaviour. Perfect square pixels were never a thing in doom, but when opengl was implemented, the aspect ratio correction was not applied to 200p and 400p. Note that on software, it does apply the correction.

I also don't think any other source port supports square pixels.

Btw, it doesnt force 240p, it scales the 200p pixels to a 240p frame. 240p is a wrong rendering size, as can be seen when it warps the text in weird ways. Stretched 200p is correct

Is there a reason you'd like this to be possible?

TheGershon:
> Perfect square pixels were never a thing in doom

I'm aware of this. [Like the Issue Author you quoted](https://github.com/kraflab/dsda-doom/issues/531#issuecomment-2521144235), I use an external application to stretch 200 -> 240 along with a CRT shader. Not having square pixels makes it impossible to not look ugly.

Pedro-Beirao:
Hmm I see, it is a very specific setup, and we were (afaik again) the only ones (accidentally) supporting it.
I think its worth it having specific support for square pixels, but keep it hidden in the config file.

Thanks for the suggestion.

rfomin:
> I also don't think any other source port supports square pixels.

It is possible to turn off aspect ratio correction in Chocolate Doom/Crispy Doom (in the setup program) and in Woof (config only). Maybe we should add such option to DSDA-Doom.

Pedro-Beirao:
I had searched in choco's setup and didnt find that option. But good to know

rfomin:
> I had searched in choco's setup and didnt find that option.

It's "Configure Display"->"A" (Advanced)->"Force correct aspect ratio".

image

PR #573, opened 2025-01-15
Fix "Show Alive Monsters" Bug (When switching from OpenGL to Software)

andrikpowell:
- Fully disable option in Software (since it's OpenGL only)
- Fixed a bug when switching from OpenGL to Software with "Show Alive Monsters" active (it no longer makes monsters / items invisible in Software)
- Added an "Action Only Supported in OpenGL" message when attempting to use feature in Software (This also resets the option to off)


fabiangreffrath (approved):
Looks correct.

Pedro-Beirao (approved):

PR #582, opened 2025-01-23
Reintroduce complevel check for Ultimate Doom Episode 4

Spectere:
It looks like kraflab may have accidentally introduced a regression when they refactored the episode handler in 874ff9525a775ca5bdfda526396fd4bb18b636fc. Traditionally, Thy Flesh Consumed could only be selected in complevel 3 and when an Ultimate Doom IWAD is being used, but from that point forward the complevel check was dropped and only a `gamemode == retail` check remained. This patch simply reintroduces the complevel check.

It's safe to say that this isn't going to impact too many people, but it technically restores the correct, pre-0.27.0 behavior. :)


Pedro-Beirao (approved):

rfomin (approved):

Pedro-Beirao:
Thank you!

PR #584, opened 2025-01-24
Dark Overlay Fixes

andrikpowell:
Sorry Re-did this PR with a different branch name.

I'm kinda treating this PR as my Dark Overlay omnibus PR.

### DIFF
--- a/prboom2/src/r_data.c
+++ b/prboom2/src/r_data.c
@@ -379,6 +379,7 @@ static void R_InitSpriteLumps(void)
 static void R_InitColormaps(void)
 {
   int i;
+  extern const byte* colormap_lump;
   // MAP_FORMAT_TODO: not sure about this
   if (hexen)
   {
@@ -393,7 +394,8 @@ static void R_InitColormaps(void)
     numcolormaps = lastcolormaplump - firstcolormaplump;
   }
   colormaps = Z_Malloc(sizeof(*colormaps) * numcolormaps);
-  colormaps[0] = (const lighttable_t *)W_LumpByName("COLORMAP");
+  colormap_lump = W_LumpByName("COLORMAP");
+  colormaps[0] = (const lighttable_t *)colormap_lump;
   for (i=1; i<numcolormaps; i++)
     colormaps[i] = (const lighttable_t *)W_LumpByNum(i+firstcolormaplump);
   // cph - always lock
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -499,6 +499,11 @@ void D_Display (fixed_t frac)
     D_DrawPause();
   }
 
+  V_BeginUIDraw();
+  if (M_MenuIsShaded())
+    M_ShadedScreen(0);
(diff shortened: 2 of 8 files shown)


andrikpowell:
Let me know if you have a better idea for the Hexen fix (Software is what has the issue).
Problem is that the black colour is taken from `colormaps[0]` and that gets replaced when it swaps the colormap with the "FOGMAP" during the maps with fog.

There are two ways to approach this:
1. Current way is setup to check for "FOGMAP" specifically, and use the main colormap if true
2. Instead of checking for "FOGMAP", we can just check if the fadetable is set to anything other than the main "COLORMAP"

Edit: I actually think the second option is better, since it's same way that `dsda_HexenApplyFadeTable()` is done

Pedro-Beirao (commented):
I dont know how common it is in the code, but I dont like seeing W_GetNumForName being called every frame.

Could you use the variable `LevelUseFullBright`? It is set to false if FOGMAP is being used. No need for that new function

andrikpowell:
> I dont know how common it is in the code, but I dont like seeing W_GetNumForName being called every frame.
> 
> Could you use the variable `LevelUseFullBright`? It is set to false if FOGMAP is being used. No need for that new function

Ok, here's a completely different approach. I went with your `LevelUseFullBright` idea, but I also addressed running the "COLORMAP" lookup on every frame as well.

Curious what you think :)

prboom2/src/r_data.c
```diff
@@ -393,7 +394,8 @@ static void R_InitColormaps(void)
     numcolormaps = lastcolormaplump - firstcolormaplump;
   }
   colormaps = Z_Malloc(sizeof(*colormaps) * numcolormaps);
-  colormaps[0] = (const lighttable_t *)W_LumpByName("COLORMAP");
+  colormap_lump = W_LumpByName("COLORMAP");
+  colormaps[0] = (const lighttable_t *)colormap_lump;
```
Pedro-Beirao, line 398: Isnt colormap_lump always the same as colormaps[0]?
I may be overlooking something but it seems like you could simply call colormaps[0] in V_DrawShaded

andrikpowell: No, that's the problem. colormaps[0] ends up getting replaced with the FOGMAP.

That's why `darkcolormap` becomes `colormaps[0]` (scrn is usually 0) if the FOGMAP is found:
```darkcolormap = colormaps[scrn]```

This is me compensating for that fact, by putting the original "COLORMAP" into memory, so that even if `colormaps[0]` is replaced, we can still access the original non-fog colormap.

andrikpowell: This is from the `dsda_HexenApplyFadeTable()` function that happens after `colormaps[0]` is defined:

```colormaps[0] = (const lighttable_t *) W_LumpByNum(fade_lump);```

Pedro-Beirao: Thanks for the explanation

Pedro-Beirao (approved):

PR #586, opened 2025-01-25
"Boom Translucent Sprites" Tweaks / Fixes

andrikpowell:
- Added the option to the menu _(you can choose where to put it)_
- Allowed option to affect sprites in-game
- Added a config only option for making ghosts in Vanilla translucent


Pedro-Beirao:
Just to repeat what we talked in discord.
- Translucent ghosts shouldnt be limited to vanilla since they are possible in boom
- Allow having translucency in vanilla, but make the default be only boom and upwards

fabiangreffrath:
> * Translucent ghosts shouldnt be limited to vanilla since they are possible in boom

"impossible", right?

Pedro-Beirao:
> "impossible", right?

Possible 😄 
https://www.doomworld.com/forum/topic/54008-weird-demos/?page=8&tab=comments#comment-2766234

It was discovered during a speedruning competition last year

Pedro-Beirao:
Also this https://doomwiki.org/wiki/Ghost_monster

prboom2/src/dsda/configuration.c
```diff
@@ -1211,6 +1211,10 @@ dsda_config_t dsda_config[dsda_config_count] = {
     "boom_translucent_sprites", dsda_config_boom_translucent_sprites,
     CONF_BOOL(1), NULL, NOT_STRICT, deh_changeCompTranslucency
   },
+  [dsda_config_vanilla_translucent_ghosts] = {
+    "vanilla_translucent_ghosts", dsda_config_vanilla_translucent_ghosts,
+    CONF_BOOL(0), NULL, NOT_STRICT
```
jackrjli: IMO this shouldn't be allowed in strict mode.

jackrjli:
> > "impossible", right?
> 
> Possible 😄 https://www.doomworld.com/forum/topic/54008-weird-demos/?page=8&tab=comments#comment-2766234
> 
> It was discovered during a speedruning competition last year

Would this setting detect Boom ghosts? I haven't tested, but Boom ghosts are a little different from vanilla ghosts.

prboom2/src/dsda/configuration.c
andrikpowell: Correct, in Strict mode this is should be disabled.

Pedro-Beirao:
> Would this setting detect Boom ghosts? I haven't tested, but Boom ghosts are a little different from vanilla ghosts.

Havent looked at the code for that, nor how boom ghosts work, but probably not.
Still I think we should support both, just for the sake of it. Shouldnt be hard.

andrikpowell:
Since it's most definitely worth disabling translucent ghost monsters in strict, I think it's also worth disabling vanilla translucency in strict as well.

The best example for my reasoning is this: "fighting an archviles with translucent flame sprites in vanilla could be considered cheating, as there'd be more visibility."

Because of this I do actually think having a separate "translucency in vanilla" option is better than having it merged into the main "translucency" setting, as it's simpler to check a bool value to track strict features than a non-bool value.

I still think this arrangement is simpler and easier to understand:
- Translucent Sprites - yes/no
- Translucency in Vanilla - yes/no
- Ghost Monsters are translucent - yes/no

As opposed to this:
- Translucent Sprites - off/on/on + vanilla
- Ghost Monsters are translucent - yes/no

It's a little tricker to name the "on" option in the merged example with multiple values, as I don't want to use "boom and up" or "boom+". The reason I don't like that is because it becomes much more confusing when the third option goes down in complevel.

I think that if we want to go with a multiple values option, the settings above would be slightly easier to understand... Altho I still think that trying to force options together makes things more confusing.

I'm also unsure if we'd like the "ghost monster" option in the menu or not, as this PR currently does not. Woof does have this in the menu.

jackrjli:
I'm in favor of disabling vanilla translucency in strict mode. In casual play I actually disable Boom translucency (for aesthetics) but do so using an autoloaded DeHackEd patch rather than through the config option specifically to keep the translucency on the arch-vile flames (and the BFG explosion, useful when firing point-blank shots). So I think it's significant enough that it ought to be restricted.

FWIW Woof takes the same approach, disabling sprite translucency for vanilla complevels when using its strict mode.

andrikpowell:
Ok, new update. It was kinda annoying to allow 2 values on strict but change it when it was set to the third (I don't think DSDA has ever done that). But that's how the strict works for the third vanilla translucency option.

Also the "branch conflict" is just because I renamed "dsda_config_boom_translucent_sprites" to "dsda_config_translucent_sprites" in the backend, and I guess it didn't like that (because of this branch being out-of-date).

Might be worth combining both the strict tracking for "vanilla translucency" and "ghost translucency" into just one tracking value.

Pedro-Beirao:
So why is translucency being tracked at all? They shouldnt be customizable in strict mode.

andrikpowell:
> So why is translucency being tracked at all? They shouldnt be customizable in strict mode.

Because previously (or currently) `boom_translucent_sprites` is not a strict setting. So for boom, disabling it doesn't not affect strict. I guess the thought process is that you don't get a boon from sprites being opaque?

We can force it to 1 for strict if want... It'd make setting strict for the setting much easier.

I think you misunderstand what strict mode is. Setting something to strict mode is different from completely blocking it from demos. It's that it's allowed for casual demos... Thus for a strict setting we need to track everything that's allowed for casual demos to see if a feature is being used in a demo.

andrikpowell:
@Pedro-Beirao Updated strict mode tracking features to use correct numbers, as well as extending the demo footer to 128 bits.

Do we need to extend this as well?:
https://github.com/kraflab/dsda-doom/blob/ba9809ad20947bb6bca362297f7d402eaef00ca5/prboom2/src/dsda/features.c#L94-L103

Pedro-Beirao:
Again to repeat what we talked on discord.

Yes the bitmap needs to be increased to 128 to accommodate the new and future features.
Since there is no uint128 in 32-bit targets, I'll be using an array of uint64. So this PR needs to wait until then.

As for how these features work, before uf_unknown, it tracks the features allowed in strict mode, after that, are the ones that are not allowed.

andrikpowell:
I'm closing this because it's a mess of conflicts at this point.

Issue #588, opened 2025-01-25
Inaccurate OpenGL Software render emulation


Kappa971:
Hi. As specified in the title, currently the emulation of the Software render color palette in OpenGL is not perfectly accurate.

**Software**

![Image](https://github.com/user-attachments/assets/5896d87a-de32-41ee-ab7d-f854c7920d5a)

**OpenGL**

![Image](https://github.com/user-attachments/assets/3b44fcad-7326-4c76-ba24-a90243e733e8)

You can notice that in OpenGL the image is brighter, the darker color tone seems to be completely missing.

Kappa971:
It should be partially solved with https://github.com/kraflab/dsda-doom/pull/590

PR #590, opened 2025-01-27
Fix walls being too bright in OpenGL

jackrjli:
This reverts some of the changes that were made to the indexed lightmode shader when it was refactored for smooth fade mode (#409). In my testing these end up making the indexed lightmode less accurate to software.

Note that this does not address floors and ceilings being different between software and OpenGL, but walls should be pretty close.

Comparison screenshots:
Software
sw

OpenGL, without this change
gl-before

OpenGL, with this change
gl-after

### DIFF
--- a/prboom2/data/lumps/gls_main.lmp
+++ b/prboom2/data/lumps/gls_main.lmp
@@ -9,6 +9,11 @@
 // and publishing an article explaining how it all works.
 // https://medium.com/@jmickle_/writing-a-doom-style-shader-for-unity-63fa13678634
 
+// Also thanks to Gooberman, since I borrowed his method for
+// fixing rounding issues for the colorIndex variable below.
+// A little bit of fudgery goes a long way... :P
+// https://github.com/GooberMan/rum-and-raisin-doom/commit/91500a86b02243dd90958aae4427051ad5c63b0d
+
 // [XA] adjusted this value a bit from the truecolor shader,
 // to taste -- seems to match software pretty damn closely
 #define DOOMLIGHTFACTOR 200.0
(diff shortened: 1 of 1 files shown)


Kappa971:
If you look at this https://github.com/kraflab/dsda-doom/issues/588, you'll notice that I'm getting a different result than shown here.
My screenshots are in 1920x1080, maybe the lighting changes based on the resolution?

jackrjli:
@Kappa971 Are you autoloading the WAD shared in Discord? That's an older version of what I have here.

Kappa971:
> @Kappa971 Are you autoloading the WAD shared in Discord? That's an older version of what I have here.

I'm not autoloading any WADs, except widescreen STBAR.

jackrjli:
This is what I get at 1920x1080 resolution:

Software
sw

OpenGL (no fix)
gl-old

OpenGL (with fix)
gl-new

Could you test this WAD (or build dsda-doom.wad with this change) to see how it looks?
[gllightfix_new.zip](https://github.com/user-attachments/files/18581650/gllightfix_new.zip)

Kappa971:
As you can see by comparing your `OpenGL (no fix)` image with the one in the issue I opened, they are equally different (as before), so I think the resolution has nothing to do with it.
I am using DSDA-Doom compiled on January 27th and not version 0.28.3. I'm on Windows 11 and I'm using an AMD GPU.

With "gllightfix_new.wad" the result seems the same instead:
dsda-doom fix
but still different from software rendering, so it's not really a fix but just a slight improvement. Someone with OpenGL knowledge would be needed to dig into this (In my opinion).

Pedro-Beirao:
@Kappa971 so your concern is that it isnt a 1:1 match with software? Well it never will be. OpenGL is just a completely different way to render.
But this is a good change that shortens the difference.

Kappa971:
Just to know, why are @jackrjli and I experiencing different behavior?
@jackrjli OpenGL (no fix)
image

@Kappa971 OpenGL (no fix)
image

Pedro-Beirao:
I dont know how to answer that. The no-fix behaviour is also different for me, but the fixed one is the same as the screenshots you guys sent.
Maybe in the old code there was something that was calculated differently in each GPU?

Interestingly, the no-fix looks very similar, just slightly different, to the new one in my machine (M1 Mac). 

OpenGL (no-fix)
1-old-gl

OpenGL (fix)

2-new-gl

Pedro-Beirao (approved):
Seems right, and definitely gives better results

jackrjli:
No idea why we all get different "no fix" results, but at least the "with fix" results are the same :sweat_smile:
For the record, I'm on Linux (Fedora 41), with an Nvidia GPU.
Just to throw one more into the mix, I gave it a try on my laptop with Intel integrated graphics:

OpenGL (no fix)
gl-before

OpenGL (with fix)
gl-after

Pedro-Beirao:
Merging. Thank you!

PR #592, opened 2025-01-28
Menu Rework

Pedro-Beirao:
- [x] Add new menus
        Display, Demos, Compatibility
        Officially dropping the boom menu graphics in favor of future DBIGFONT support
        
- [x] Reorganize all menus
        Some thing may still warrant improvement, but Im happy with this.
        
- [x] Display page names
        Its also possible to display them using a carousel. Used for keybindings
        
- [x] Vertical scrolling now has a graphic indicating it can be scrolled down/up

- [x] Rewrite Thermo code to allow using it for options
        It is still not very modular, only really suited for setting the volume
        Doesnt look amazing on raven, but good enough imo

- [x] Raven fixes
        New graphics for < and >
        Use FONTA for options and FONTB for main title

Ideas for another time:
- "Disabled" options
- DBIGINFO support
- Highlight the words "opengl" and "software" in options that only support on of them


andrikpowell:
So this looks pretty nice...

My only comment is that I think the scrollbar colour should be customisable in DSDATC (`menu_scrollbar`).

Also out of curiosity, what do you think of using `CR_BROWN` instead of `CR_TAN` for the scrollbar?

rfomin:
Found a little issue, MIDI player owerwrite help string in Ancient Aliens:
doom02

prboom2/src/m_menu.c
```diff
@@ -168,15 +168,17 @@ extern dboolean  message_dontfuckwithme;
...
 dboolean setup_active      = false; // in one of the setup screens
+dboolean set_general_active = false;
 dboolean set_keybnd_active = false; // in key binding setup screens
+dboolean set_display_active = false;
+dboolean set_demos_active = false; // in demos setup screen
+dboolean set_compatibility_active = false;
 dboolean set_weapon_active = false; // in weapons setup screen
-dboolean set_status_active = false; // in status bar/hud setup screen
 dboolean set_auto_active   = false; // in automap setup screen
+dboolean level_table_active = false;
```
rfomin: We can make these variables `static` so that they are only visible in the m_menu.c file. We can consider making most of the variables and function declarations `static`, since there are a lot of changes in m_menu.c anyway.

By the way, how do you want to merge this PR? Squash or history editing? There are a lot of commits.

Pedro-Beirao:
> My only comment is that I think the scrollbar colour should be customisable in DSDATC (`menu_scrollbar`).

Done. But I dont like how CR_BROWN looks

> Found a little issue, MIDI player owerwrite help string in Ancient Aliens:

Fixed by limiting the font max height to 9 (AA has an already big font, then it even adds an empty row in the bottom, making it bigger).
The issue itself is that an offset is applied to account for the Thermo, but isnt accounted by the scrolling logic. When I get back to possibly use Thermo in places other than Volume, I will figure out how to properly handle it.

prboom2/src/m_menu.c
Pedro-Beirao: I agree. Did a quick pass to make most of the variables and functions static in https://github.com/kraflab/dsda-doom/pull/592/commits/e835bb85f3831bf74eed9f735301da0886e90936

I wanted to simply squash, but since there's this big change, went with history editing

rfomin (approved):
Looks good, thanks.

PR #593, opened 2025-01-28
Separate "Player Trail Mode" Option

andrikpowell:
This setting in the menu is long and formatted in an awkward way.
As a result the option is hard to fit well in the menus, and can be a bit confusing being two options.

I've separated the option "Player Trail Mode" into 2 separate options:
- "Player Trail" (off/on)
- "Include Collisions" (off/on)

I feel this is easier to understand and easier to fit in the menus compared to its current iteration.

The main downside is there isn't an easy way to convert the current setting a user has into these two options.


Pedro-Beirao (changes requested):
I do prefer this since the menu wont need that weird offset no more (and its such a rare feature to use, that no one will notice it being reset).
But the logic here is wrong.

Try setting Player Trail to On and Collisions Off.

You can work with collisions as a separate bool, or go back to `map_trail_mode_off` into the enum. But the way its done right now is wrong.

andrikpowell:
> I do prefer this since the menu wont need that weird offset no more (and its such a rare feature to use, that no one will notice it being reset). But the logic here is wrong.
> 
> Try setting Player Trail to On and Collisions Off.
> 
> You can work with collisions as a separate bool, or go back to `map_trail_mode_off` into the enum. But the way its done right now is wrong.

Thanks, you are correct. My logic was faulty in `AM_initPlayerTrail()`. Check out the new force pushed code. Tested it and it now works correctly. Perhaps I'm dumb, but I can't really tell the difference between "Collisions" on/off.

Pedro-Beirao (approved):

Pedro-Beirao:
> Perhaps I'm dumb, but I can't really tell the difference between "Collisions" on/off.

Only a select number of trails squares are allowed on screen at a time (map trail size). When you move, a new square is made, and possibly an older square is removed.

So with collisions on, instead of creating new square when you move a few units, it creates new squares when attempting to move. So even if you attempt to move into a wall and getting stopped by the collision, it will make a new square.

Pedro-Beirao:
Ive already approved this, but before I merge, could you change `false` to `map_trail_mode_off`? Its more correct.
Sry 😅 
```
map_trail_mode = dsda_IntConfig(dsda_config_map_trail) ? trail_collisions : false;
```

andrikpowell:
> Ive already approved this, but before I merge, could you change `false` to `map_trail_mode_off`? Its more correct. Sry 😅
> 
> ```
> map_trail_mode = dsda_IntConfig(dsda_config_map_trail) ? trail_collisions : false;
> ```

done :)

Pedro-Beirao:
Thank you!

PR #594, opened 2025-01-29
Tweak "Parallel Same-Sound Limit" Option

andrikpowell:
So there are a couple things I'd like to pitch here... But first, I need to explain what my main issue with this option is in its current form.

I will admit I didn't understand what this option was supposed to do at first, and I don't think that the option's current wording helped. But I need to explain how this option works, to show why I think it should be tweaked.

So there are two options:
- Parallel Same-Sound Limit
- Parallel Same-Sound Window

First off, we need to understand what "Parallel Same-Sound Limit" does. Essentially, what this option tries to limit the amount of sounds that are played. However it only limits the sounds with two criteria: 1) if it is the same sound, and 2) if it is playing at the same time.

The best example of this would be a group of Revenants, as they are very loud and are very noisy. This option can be used to limit how many Revenant "wake" sounds happen under a certain "window". This is where the second option comes in called "Parallel Same-Sound Window". Essentially this value is within how many tics does it limit the same-sounds.

Now that we have an understanding of the option, I'm going to explain my issue with the setting. The value of "Parallel Same-Sound Limit" can be set from 0-32. 32 translates to 32 iterations of the same-sound can be played at once, whereas 1 translates to only 1 of the same-sound can be played at once. Following this logic, the higher the value, the more sounds can play...

However, what happens when you set the "Parallel Same-Sound Limit" to 0? Well apparently the limit is removed altogether. The problem here is that the logic is supposed to be the lower the value, the more sounds are being limited. The higher the value, the less sounds limited. However 0 breaks the logic, as the limit is removed, acting far closer to the value of 32 for the limit.

My suggestion is to separate the on/off functionality of 0 currently present the "Parallel Same-Sound Limit" option, into a new boolean option acting as an on/off switch. With the current option becoming a "Parallel Same-Sound Number" option to specify how many sounds to limit instead.

I will mention that I have these options renamed in Nyan Doom as the following:
- Overlapping Same-Sound Limit
- Number of Overlapping Sounds
- Sound Replay Window (tics)

Although I haven't renamed these options in this PR, as I'm not sure if people would like them renamed. These names just make slightly more sense to me.


jackrjli:
An alternative I just thought of is to display 0 as "disabled" instead of the number. Not sure how much reworking of the menu code that would take, but there are some other settings (FPS limit, rewind timeout) that could benefit from the same.

Pedro-Beirao:
I understand that you prefer separated options, but for some things (like this for example), its just more practical to keep this behaviour.

Having "disabled" instead of "0" seems like a good idea in this case, as well as in FPS limit and rewind timeout.

andrikpowell:
> I understand that you prefer separated options, but for some things (like this for example), its just more practical to keep this behaviour.
> 
> Having "disabled" instead of "0" seems like a good idea in this case, as well as in FPS limit and rewind timeout.

While I agree that having 0 mean "disabled" is somewhat of a good idea, it's terrible in terms of user experience and UI. If we did go with something like that I'd prefer to use "0 - disabled".

The reason this is terrible is this: "a user sees an option called disabled on the menu. Due to how all the other "word based" options, they would infer it'd be a multiple choice option. If they went in to change it, I could see them getting confused in learning that the option actually requires a number input."

This is why I think this is terrible... You've turned a simple to understand "number" option into not showing a number when it's "off". Now I'm sure most people may figure out how this works *eventually*, but it doesn't change the fact that you've overcomplicated how the option works at the cost of readability.

My other comment on this would be if disabled options ever get added to the menu, it'd make a lot more sense for there to be an off/on switch that would then disable the 2 below options.

andrikpowell:
> An alternative I just thought of is to display 0 as "disabled" instead of the number. Not sure how much reworking of the menu code that would take, but there are some other settings (FPS limit, rewind timeout) that could benefit from the same.

Regarding these 2 examples, I have some thoughts. They do have a similar problem to the option this PR is about, but that just means I feel a similar way about these options as well. Disabling a number option set to 0, is such a programmer move, it's not even funny. Like since I work with code, it does make sense... But to a normal user, it's not very clear. It's like the difference between how a programmer sees 0 as the first value, whereas a user sees 1 as the first. I think it's best to leave programming stuff like this in the actual code and not for users to have to deal with

- FPS limit - it doesn't exactly seem intuitive to set this to 0 to disable it, although I do think in this context it makes a tiny bit of sense. I very much doubt most people would come to conclusion that to disable this limit, they'd have to set it to 0... Although idk if I see that as a bad thing. You can do a similar affect to "unlimited" by just raising the FPS number super high (aka 600), and it'd prob be better to avoid GPU strain. So in a way it's actually better for the user to cap the framerate to a specific number instead of disabling it altogether.

- Rewind Timeout - I'm kinda confused why you can set this to 0, or what setting this to 0 actually does. Like is there a benefit to disabling this feature, if you can just make sure the rewind key is unbound. If there is a benefit to disabling this, then I definitely think there should be an on/off switch, as it could help older machines.

jackrjli:
> * Rewind Timeout - I'm kinda confused why you can set this to 0, or what setting this to 0 actually does. Like is there a benefit to disabling this feature, if you can just make sure the rewind key is unbound. If there is a benefit to disabling this, then I definitely think there should be an on/off switch, as it could help older machines.

The rewind feature can be disabled by setting "rewind depth" (`dsda_auto_key_frame_depth`) to 0. This is mainly done for performance, as saving keyframes can cause the game to stutter on large maps. But that's different from rewind _timeout_ - "rewind timeout" is a feature that automatically disables rewinds if dsda-doom detects that saving keyframes is taking longer than the configured timeout, again for performance. Setting _this_ to 0 disables the timeout feature, which is done if you want to guarantee that rewinds are available, whatever the performance cost.

(Honestly, that hardly anyone seems to understand this setting is probably a bigger problem than how 0 for the setting is treated...)

andrikpowell:
> The rewind feature can be disabled by setting "rewind depth" (`dsda_auto_key_frame_depth`) to 0. This is mainly done for performance, as saving keyframes can cause the game to stutter on large maps.
Now I'm thinking that the overall "rewind" feature should have an on/off switch. :/

It's actually kinda weird that it's just always on by default, and that older computers are just kinda forced to run it this way... Like I guess the default "on" is fine, but I don't think people exactly understand that "rewind depth" is the on/off switch.

I have a mapper friend who has an ancient CPU from around 2012, and they tend to have issues running some maps. I do wonder if they knew about this rewind on/off switch and if it would improve performance for them.

If there is an on/off switch needed here, it's actually probably less needed for the "rewind timeout" and more for the actual "rewind" feature as a whole :P

To clarify (for my sanity mostly):
- Rewind Interval (s) - the amount of seconds to go back when "rewind" key is pressed
- Rewind Depth - I'm guessing this is the amount of keyframes that are saved? So the default is 60... but I kinda don't know what that means... does that mean it's saving up to 60 instances of the "rewind interval" above? or is it saving 60 tics?
- Rewind Timeout (ms) - So if keyframes are taking longer than the default set "10 milliseconds", then it disabled rewind altogether

I guess my next question is what causes a rewind timeout to happen? Like I assume it bombs out if too much info is saved from the map, causing keyframes to take longer than the set value. But what affects that? The amount of monsters, or the amount of tags in a map?

(I realise we are going a bit off topic here, but tbh I've never actually understood what these options did, nor how I could use them to my benefit).

rfomin:
> It's actually kinda weird that it's just always on by default, and that older computers are just kinda forced to run it this way... 

It is automatically disabled when saving takes a long time.

> I guess my next question is what causes a rewind timeout to happen? Like I assume it bombs out if too much info is saved from the map, causing keyframes to take longer than the set value. But what affects that? The amount of monsters, or the amount of tags in a map?

Keyframes are like normal savegames, they contain all mobjs and level geometry. It will be slow if the level contains a lot of monsters and lines. I guess this option was added because people have different tolerance to lag. If you ask me, I would remove this option and hardcode some value.

Pedro-Beirao:
Again on this, I vote to not merge. Some things make sense to separate, but I dont think this does

In the future we should look into the "0 - Disabled" suggestion

thecnoNSMB:
> I understand that you prefer separated options, but for some things (like this for example), its just more practical to keep this behaviour.

Forgive me, but could you explain the rationale here?

Pedro-Beirao:
> Forgive me, but could you explain the rationale here?

We have limited space in the menu to add options. So in some pages that already have a lot of options, and will possibly have more in the future, I think we should try to not add useless (not from a design perspective ig) options.
In this case its for the Audio page, so I think this applies.

For https://github.com/kraflab/dsda-doom/pull/593, separate options make sense.

thecnoNSMB:
> We have limited space in the menu to add options. So in some pages that already have a lot of options, and will possibly have more in the future, I think we should try to not add useless (not from a design perspective ig) options. In this case its for the Audio page, so I think this applies.
> 
> For #593, separate options make sense.

Ah, so it's a tradeoff against design effort required to organize and paginate the options menu, that's fair. I personally think this options change might be worth it for the improved clarity, but I'm not familiar with the codebase. If not, since this particular option only goes from 1-32, maybe 0 could display as "Unlimited", and the value could wrap when clicking through it with the arrow keys, so that "Unlimited" appears after 32 instead of before 1? That might be the best of both worlds.

andrikpowell:
> If not, since this particular option only goes from 1-32, maybe 0 could display as "Unlimited", and the value could wrap when clicking through it with the arrow keys, so that "Unlimited" appears after 32 instead of before 1?

Um, this is really confusing and kinda overcomplicating things. the value that means 0 shouldn't come after 32... even if it means "unlimited".

I'd much rather type in a number than have to cycle through a bunch of options. Plus this logic wouldn't work for something like "fps limit" which "unlimited" would have to go all the way past 1000, which would be absurd to cycle through.

While I'm still not super into the "0 - Disabled" idea, it's the best idea I've seen besides just separating the option to have an off/on switch.

You still have the slight issue of players not knowing that an option can be disabled though, because it's not printed anywhere. They just have to know that an option can be disabled by typing 0, which is a little weird. It sorta makes sense in the "fps limit" context, but in other contexts, it's not as clear (such as "Rewind Depth").

Personally I think Rewind should have an off/on switch, even more than the option this PR is about. My mapper friend with an old 2012 CPU would have that "rewind disabled" message come up on almost every map they played... so I think it may actually be worth having an on/off switch for that at least.

thecnoNSMB:
Personally, my ideal solution would be to add a separate option like you originally suggested, I was just brainstorming ideas just in case adding a new option is absolutely 100% off the table.

Pedro-Beirao:
Closing. Sorry, I prefer the way it currently is.

PR #597, opened 2025-01-30
Save button list

rfomin:


prboom2/src/p_saveg.c
```diff
@@ -741,6 +741,7 @@ typedef enum {
   tc_poly_door,
   tc_quake,
   tc_ambient_source,
+  tc_button,
```
fabiangreffrath, line 744: If we put this past `tc_end`, we'd stay at least forward compatible.

rfomin: DSDA-Doom does not guarantee backwards/forwards compatibility, I think we should not compromise here.

PR #601, opened 2025-02-04
Improve looping for older midi hardware devices

### LINKED ISSUE #600: Music track loop issues with Nuked SC55 (and real hardware?)
Hi. I'm using DSDA-Doom with Nuked-SC55. I notice a small problem with midi playback.
Basically the "mus_portmidi_reset_delay" option seems to work when the music track starts (`0` as default is not good, maybe a solution like in Woof should be adopted which applies an automatic delay if set to `-1`), but it doesn't seem to work during the track loop: the track ends, starts again but no delay is added to get the emulator ready to start playing again, and this causes some initial notes to be skipped or played too fast. For example, this happens in the music track of the E2M1 level of The Ultimate Doom.

I don't know what solution they adopted in the DOS version, but the loop seems to work correctly there with Nuked-SC55.
Assuming that Nuked SC55 behaves like real hardware, I would imagine that in the source ports of the last 20 years, midi playback has never been tested on SC-55, probably because people used software solutions like the built-in Windows synthesizer, Edirol Virtual Sound Canvas, etc.

ceski-1:
Fixes https://github.com/kraflab/dsda-doom/issues/600

### DIFF
--- a/prboom2/src/MUSIC/portmidiplayer.c
+++ b/prboom2/src/MUSIC/portmidiplayer.c
@@ -121,6 +121,8 @@ static const char *pm_name (void)
 #include <delayimp.h>
 #endif
 
+static dboolean channel_used[16];
+
 #define DEFAULT_VOLUME 100
 static int channel_volume[16];
 static float volume_scale;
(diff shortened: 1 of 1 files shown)


rfomin:
Thank you!

PR #602, opened 2025-02-06
Remove SysEx support from FluidSynth midi player

ceski-1:
There are broken songs like DBP37_AUGZEN.wad MAP22 that don't work correctly when SysEx is enabled. Most songs don't use these features and FluidSynth's support is incomplete. This also cuts down on console spam:

```
fluidplayer: SYSEX message received but not understood
fluidplayer: SYSEX message received but not understood
fluidplayer: SYSEX message received but not understood
fluidplayer: SYSEX message received but not understood
```

### DIFF
--- a/prboom2/src/MUSIC/flplayer.c
+++ b/prboom2/src/MUSIC/flplayer.c
@@ -102,10 +102,6 @@ static double spmc;
 static double f_delta;
 static int f_soundrate;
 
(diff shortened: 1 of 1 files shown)


rfomin (approved):

Pedro-Beirao:
Thank you!

PR #603, opened 2025-02-06
Add "GAMEINFO" / Episode IWAD load support

andrikpowell:
So this is going to be a fun one...

This solution is pretty jank and probably not the best implementation... but it does work!

This is similar to the `GAMEINFO` lump that GZDoom uses that can specify an IWAD to load via a PWAD lump.

Essentially to search the PWADs for an "IWAD" lump, I needed to cache the lumps before choosing the IWAD. So what ends up happening is that I cache it once (as a sort of quick cache), read the IWAD lump and then clear the cache for the "Real" lump cache.

Currently the main thing I don't like is, I haven't been able to combine the zip functions together as one with the variable "MainLumpCache".

If you're wondering what this is useful for, it's so I can drag-and-drop a plutonia based wad on dsda-doom, and it can load `plutonia.wad` if it finds it in the directory. I have a clause that defaults to normal behaviour if the wad specified in the "IWAD" lump isn't found.

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -124,6 +124,8 @@
 static void D_PageDrawer(void);
 
 char* iwadlump;
+char* iwadver;
+int EpisodeStructure = false;
 
 // jff 1/24/98 add new versions of these variables to remember command line
 dboolean clnomonsters;   // checkparm of -nomonsters
@@ -183,8 +185,27 @@ const char *const standard_iwads[]=
 
   "heretic1.wad"
 };
+
+// list of episode-formatted IWAD names
+const char *const episode_iwads[]=
+{
+  "doom.wad",
+  "doom1.wad",
+  "doomu.wad", /* CPhipps - alow doomu.wad */
+
+  "freedoom1.wad",
+
+  "chex.wad",
+  "rekkrsa.wad",
+
+  "heretic.wad",
+
+  "heretic1.wad"
+};
+
 //e6y static
 const int nstandard_iwads = sizeof standard_iwads/sizeof*standard_iwads;
+const int nepisode_iwads = sizeof episode_iwads/sizeof*episode_iwads;
 
 /*
  * D_PostEvent - Event handling
@@ -1079,6 +1100,7 @@ static char *FindIWADFile(void)
   int   i;
   dsda_arg_t* arg;
   char  * iwad  = NULL;
+  int iwadnum = EpisodeStructure ? nepisode_iwads : nstandard_iwads;
 
   if (CheckExeSuffix("-heretic"))
   {
@@ -1106,8 +1128,8 @@ static char *FindIWADFile(void)
     if (iwadlump != NULL)
       return I_FindWad(iwadlump);
 
-    for (i=0; !iwad && i<nstandard_iwads; i++)
-      iwad = I_FindWad(standard_iwads[i]);
+    for (i=0; !iwad && i<iwadnum; i++)
+      iwad = EpisodeStructure ? I_FindWad(episode_iwads[i]) : I_FindWad(standard_iwads[i]);
   }
   return iwad;
 }
@@ -1694,6 +1716,38 @@ static void dsda_Loadfiles(void)
 }
 
 //
+// DetectEpisodeStructure
+//
+// If GAMEINFO is not present but Doom 1 maps are found, autoload episode IWAD.
+//
+
+static void dsda_DetectEpisodeStructure(void)
+{
+  int i, ii;
+  char lump[5];
+  char *iwad;
+
+  for (i=0;i<10;i++)
+  {
+    for (ii=0;ii<10;ii++)
+    {
+      sprintf(lump, "E%dM%d", i, ii);
+      if (W_LumpNameExists(lump))
+      {
+        EpisodeStructure = true;
+        iwadver = Z_Strdup(lump);
+        iwad = FindIWADFile();
+
+        if (iwad)
+          iwadlump = Z_Strdup(iwad);
+
+        return;
+      }
+    }
+  }
+}
+
+//
 // IdentifyVersion
 //
 // Set the location of the defaults file and the savegame root
@@ -1731,6 +1785,10 @@ static void IdentifyVersion (void)
     // Parse GAMEINFO lump
     dsda_LoadGameInfo();
 
+    // Autodetect Doom 1 maps
+    if (iwadlump == NULL)
+      dsda_DetectEpisodeStructure();
+
     // Reset lump cache
     dsda_ResetInitLumpCache();
 
@@ -1743,15 +1801,18 @@ static void IdentifyVersion (void)
       if (!(iwad && *iwad))
       {
         Z_Free(iwadlump);
+        Z_Free(iwadver);
         iwadlump = NULL;
+        iwadver = NULL;
       }
     }
   }
 
-  // If GAMEINFO IWAD not found,
+  // If GAMEINFO / Episode IWAD not found,
   // locate IWAD the traditional way
   if (iwadlump == NULL)
   {
+    EpisodeStructure = false;
     iwad = FindIWADFile();
   }
 
@@ -1922,8 +1983,11 @@ static void D_DoomMainSetup(void)
 
   if (iwadlump != NULL)
   {
-    lprintf(LO_INFO, "Detected GAMEINFO lump: %s\n", iwadlump);
+    lprintf(LO_INFO, "Detected %s lump: %s\n", iwadver ? iwadver : "GAMEINFO", iwadlump);
     Z_Free(iwadlump);
+
+    if (iwadver)
+      Z_Free(iwadver);
   }
 
   G_ReloadDefaults();
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -93,6 +93,7 @@
 #include "dsda/data_organizer.h"
 #include "dsda/map_format.h"
 #include "dsda/mapinfo.h"
+#include "dsda/gameinfo.h"
 #include "dsda/mobjinfo.h"
 #include "dsda/options.h"
 #include "dsda/pause.h"
@@ -121,6 +122,8 @@
 #include "i_glob.h"
 
 static void D_PageDrawer(void);
+
+char* iwadlump;
 
 // jff 1/24/98 add new versions of these variables to remember command line
 dboolean clnomonsters;   // checkparm of -nomonsters
@@ -1100,6 +1103,9 @@ static char *FindIWADFile(void)
     else if (dsda_Flag(dsda_arg_hexen))
       return I_FindWad("hexen.wad");
 
+    if (iwadlump != NULL)
+      return I_FindWad(iwadlump);
+
     for (i=0; !iwad && i<nstandard_iwads; i++)
       iwad = I_FindWad(standard_iwads[i]);
   }
@@ -1722,11 +1728,32 @@ static void IdentifyVersion (void)
     if (!dsda_Flag(dsda_arg_noautoload)) D_AutoloadPWadDir(); // Load autoload PWAD files for GAMEINFO lump
     W_Init(); // Quick cache to search for GAMEINFO lump
 
+    // Parse GAMEINFO lump
+    dsda_LoadGameInfo();
+
     // Reset lump cache
     dsda_ResetInitLumpCache();
+
+    // If IWAD found, check if it exists
+    if (iwadlump != NULL)
+    {
+      iwad = FindIWADFile();
+
+      // Clear data if IWAD not found
+      if (!(iwad && *iwad))
+      {
+        Z_Free(iwadlump);
+        iwadlump = NULL;
+      }
+    }
   }
 
-  iwad = FindIWADFile();
+  // If GAMEINFO IWAD not found,
+  // locate IWAD the traditional way
+  if (iwadlump == NULL)
+  {
+    iwad = FindIWADFile();
+  }
 
   // It is now ok to load dehacked / unzip files
   MainLumpCache = true;
@@ -1892,6 +1919,13 @@ static void D_DoomMainSetup(void)
 
   lprintf(LO_DEBUG, "G_ReloadDefaults: Checking OPTIONS.\n");
   dsda_ParseOptionsLump();
+
+  if (iwadlump != NULL)
+  {
+    lprintf(LO_INFO, "Detected GAMEINFO lump: %s\n", iwadlump);
+    Z_Free(iwadlump);
+  }
+
   G_ReloadDefaults();
 
   // e6y
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1128,52 +1128,6 @@ static dboolean FileMatchesIWAD(const char *name)
 }
 
 //
-// IdentifyVersion
-//
-// Set the location of the defaults file and the savegame root
-// Locate and validate an IWAD file
-// Determine gamemode from the IWAD
-//
-// supports IWADs with custom names. Also allows the -iwad parameter to
-// specify which iwad is being searched for if several exist in one dir.
-// The -iwad parm may specify:
-//
-// 1) a specific pathname, which must exist (.wad optional)
-// 2) or a directory, which must contain a standard IWAD,
-// 3) or a filename, which must be found in one of the standard places:
-//   a) current dir,
-//   b) exe dir
-//   c) $DOOMWADDIR
-//   d) or $HOME
-//
-// jff 4/19/98 rewritten to use a more advanced search algorithm
-
-static void IdentifyVersion (void)
-{
-  char *iwad;
-
-  // why is this here?
-  dsda_InitDataDir();
-  dsda_InitSaveDir();
-
-  // locate the IWAD and determine game mode from it
-
-  iwad = FindIWADFile();
-
-  if (iwad && *iwad)
-  {
-    AddIWAD(iwad);
-    Z_Free(iwad);
-  }
-  else
-  {
-    I_Error("IdentifyVersion: IWAD not found\n\n"
-            "Make sure your IWADs are in a folder that dsda-doom searches on\n"
-            "For example: %s", I_ConfigDir());
-  }
-}
-
-//
 // DoLooseFiles
 //
 // Take any file names on the command line before the first switch parm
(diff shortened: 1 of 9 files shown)


rfomin:
> This is similar to the `GAMEINFO` lump that GZDoom uses that can specify an IWAD to load via a PWAD lump.

Why not support `GAMEINFO` lump or ID24 [`GAMECONF`](https://docs.google.com/document/d/1sOI826AlCS9e85mIq4wiVoAdn0t_XV7oVq6IF98ADak/edit?usp=drive_link)? We don't have to support all fields.

I don't think we should add another lump, many WADs already contain GAMEINFO/GAMECONF.

andrikpowell:
> > This is similar to the `GAMEINFO` lump that GZDoom uses that can specify an IWAD to load via a PWAD lump.
> 
> Why not support `GAMEINFO` lump or ID24 [`GAMECONF`](https://docs.google.com/document/d/1sOI826AlCS9e85mIq4wiVoAdn0t_XV7oVq6IF98ADak/edit?usp=drive_link)? We don't have to support all fields.
> 
> I don't think we should add another lump, many WADs already contain GAMEINFO/GAMECONF.

I agree. The IWAD lump was more an example about how it would work. This PR was mostly for the functionality of choosing an IWAD based on a PWAD lump.

Personally I'd rather support `GAMEINFO`. It's simpler and it could just look for the line `iwad = "plutonia.wad"` (example wad). Although I can't say I'm that knowledgeable when it comes to parsing text lumps in general.

rfomin:
> Personally I'd rather support `GAMEINFO`. It's simpler and it could just look for the line `iwad = "plutonia.wad"` (example wad). Although I can't say I'm that knowledgeable when it comes to parsing text lumps in general.

Basically, DSDA-Doom already has a ZDoom scanner, see `src/scanner.h`. It used to parse `*MAPINFO` and `UDMF`.

andrikpowell:
> > Personally I'd rather support `GAMEINFO`. It's simpler and it could just look for the line `iwad = "plutonia.wad"` (example wad). Although I can't say I'm that knowledgeable when it comes to parsing text lumps in general.
> 
> Basically, DSDA-Doom already has a ZDoom scanner, see `src/scanner.h`. It used to parse `*MAPINFO` and `UDMF`.

Thanks! This took a little longer because I was looking at `UMAPINFO`, `UDMF` .cpp files... I really should've been looking at a more simple example like `ambient.cpp` instead :P

Your code really helped... Although I'd like to get better at understand C++ and learn how to get C and C++ to work together :)

Now that I see the "GAMEINFO" support in action, I really like it!

Pedro-Beirao:
I like the idea of supporting GAMEINFO. But lets also support the LOAD field yeah?

And woof should probably do the same

prboom2/src/d_main.c
```diff
@@ -1398,10 +1419,11 @@ static void D_AddZip(const char* zipped_file_name, wad_source_t source, deh_queu
   const char* temporary_directory;
 
   full_zip_path = I_RequireZip(zipped_file_name);
-  temporary_directory = dsda_UnzipFile(full_zip_path);
+  temporary_directory = MainLumpCache ? dsda_UnzipFile(full_zip_path) : dsda_UnzipFileInit(full_zip_path);
```
rfomin: Why do we have to unzip WADs twice?

prboom2/src/dsda/gameinfo.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  if (!scanner.CheckString()) {
+    scanner.GetNextToken();
+    scanner.SkipLine();
+    return;
+  }
+
+  if (!stricmp(scanner.string, "IWAD"))
+  {
+    scanner.CheckToken('=');
```
rfomin: We need to check `=` token like GZDoom does. Try to load this GAMEINFO with GZDoom:
```
IWAD TNT.WAD
```
So I suggest these changes:
```diff
diff --git a/prboom2/src/dsda/gameinfo.cpp b/prboom2/src/dsda/gameinfo.cpp
index 0763e3e70..19fda7be3 100644
--- a/prboom2/src/dsda/gameinfo.cpp
+++ b/prboom2/src/dsda/gameinfo.cpp
@@ -15,16 +15,13 @@
 //  DSDA GAMEINFO
 //
 
-#include <string>
-#include <cstring>
+#include <string.h>
 
 extern "C" {
 #include "d_main.h"
 #include "w_wad.h"
 #include "lprintf.h"
 #include "z_zone.h"
-
-char *Z_Strdup(const char *s);
 }
 
 #include "scanner.h"
@@ -39,21 +36,13 @@ void dsda_ParseGameInfoLine(Scanner &scanner) {
     return;
   }
 
-  if (!stricmp(scanner.string, "IWAD"))
-  {
-    scanner.CheckToken('=');
-
-    if (!scanner.CheckString()) {
-      lprintf(LO_WARN, "Invalid GAMEINFO: expects string\n");
-
-      scanner.GetNextToken();
-      scanner.SkipLine();
-      return;
+  if (!stricmp(scanner.string, "IWAD")) {
+    scanner.MustGetToken('=');
+    scanner.MustGetString();
+    if (iwadlump) {
+      Z_Free(iwadlump);
     }
-
-    std::string name(scanner.string);
-
-    iwadlump = Z_Strdup(name.c_str());
+    iwadlump = Z_Strdup(scanner.string);
   }
 }
 
```

prboom2/src/w_wad.c
```diff
@@ -503,11 +510,14 @@ void W_Init(void)
...
   // killough 4/17/98: Add namespace tags to each entry
   // killough 4/4/98: add colormap markers
-  W_CoalesceMarkedResource("S_START", "S_END", ns_sprites);
-  W_CoalesceMarkedResource("F_START", "F_END", ns_flats);
-  W_CoalesceMarkedResource("C_START", "C_END", ns_colormaps);
-  W_CoalesceMarkedResource("B_START", "B_END", ns_prboom);
-  W_CoalesceMarkedResource("HI_START", "HI_END", ns_hires);
+  if (MainLumpCache)
+  {
+    W_CoalesceMarkedResource("S_START", "S_END", ns_sprites);
```
rfomin: Why not call `W_CoalesceMarkedResource`? Unlikely, but name collisions are possible.

rfomin (commented):
Autoload folders doesn't work. For example, I created GAMEINFO for PL2.WAD (Plutonia 2) and put `gameinfo.lmp` in `autoload/pl2.wad`, but it was ignored.

In general, I think it's the right approach. Thank you!

rfomin:
> But lets also support the LOAD field yeah?

Is it practical? I have never seen a PWAD that uses it.

> And woof should probably do the same

Yes, agreed.

Pedro-Beirao:
> Is it practical? I have never seen a PWAD that uses it.

I also dont know of any PWAD that uses it, but I can imagine reasons to want it.
For example a vanilla wad that has widescreen assets distributed in a separate pwad

prboom2/src/w_wad.c
andrikpowell: I was trying to reduce the amount of lump information to cache during the first round, since it needs to be dropped for the re-cache later. But if you think we need this, I can allow it for both caches.

prboom2/src/dsda/gameinfo.cpp
andrikpowell: Good idea, I'll tweak the code so it works closer to GZDoom 👍

prboom2/src/d_main.c
andrikpowell: Mostly because I can't seem to figure out how to avoid doing that. I'm aware it's messy, and I'm not a fan of how it currently works.

The thing is that we need to recache the files that have been unzipped during the lump recache, so it comes after the IWAD in the lump order. But in order to check GAMEINFO, we need to do the first cache to check the unzipped files first.

The reason this is a pain, is because the `dsda_UnzipFile()` function doesn't just unzip files. It also adds them to the PWAD load list, and during the recache, we need to re-add these unzipped files to be in the correct load order.

What we need here is to replace `dsda_UnzipFile()` during the recache. We need a function similar to `dsda_CleanZipTempDirs()` but called `dsda_GetZipTempDirs()` that can get the files already unzipped during the recache.

Problem is I haven't been able to figure this out, and so my temp solution is just to delete the zipped files during the first cache and unzip them again for the recache :p

andrikpowell:
> > Is it practical? I have never seen a PWAD that uses it.
> 
> I also dont know of any PWAD that uses it, but I can imagine reasons to want it. For example a vanilla wad that has widescreen assets distributed in a separate pwad

I'm not sure this works the way you think it works exactly. According to the ZDoom wiki, the LOAD field loads certain wad files *BEFORE* the wad (usually iwad) that I includes the GAMEINFO lump.

In the example of a wide-screen asset wad, the LOAD lump wouldn't allow for wide-screen assets to be used, because they would be overwritten by the iwad.

prboom2/src/d_main.c
andrikpowell: Actually I feel I'm really close to getting this to work. Check out [0eace88](https://github.com/kraflab/dsda-doom/pull/603/commits/0eace88212124ff6786a987a5e022e8ea8b161c0). Main function that loads the already unzipped files is `dsda_ReadUnzippedFile()`

Only problem with this, is when loading 2+ zip files, it only seems to clear out the first temp directory.

Pedro-Beirao:
My bad then, I didnt read the "before" in the wiki. Well that sucks... what uses does loading before even have?

The simple "IWAD" lump you suggested before is looking cleaner now tbh. With the drawback that no current wads support this lump yet, unlike GAMEINFO.

I think its worth thinking more about this, as supporting a lump just for one field seems wrong. Wouldve prefered if GAMEINFO had fields for iwad, pwads and complvl.
The name is also confusing, as it is part of Z***INFO lumps. We should either get a new UGAMEINFO with the fields we need, or just support the "loose" lumps of COMPLVL and IWAD.
(Also mentioning your suggested MODINFO/UMODINFO lump on the discord)

andrikpowell:
> My bad then, I didnt read the "before" in the wiki. Well that sucks... what uses does loading before even have?

I think it's more focused around zdoom projects separating resources into different pk3s (like graphics, sounds, etc)... Perhaps even making some resources optional, which the iwad would then check if those existed.

> The simple "IWAD" lump you suggested before is looking cleaner now tbh. With the drawback that no current wads support this lump yet, unlike GAMEINFO.
> 
> I think its worth thinking more about this, as supporting a lump just for one field seems wrong. Wouldve prefered if GAMEINFO had fields for iwad, pwads and complvl. The name is also confusing, as it is part of Z***INFO lumps. We should either get a new UGAMEINFO with the fields we need, or just support the "loose" lumps of COMPLVL and IWAD. (Also mentioning your suggested MODINFO/UMODINFO lump on the discord)

I actually disagree here. The main reason I took the "IWAD" lump approach before is because I thought it'd be too difficult to parse the "GAMEINFO" lump. If I knew it was as easy as this, I would've gone with "GAMEINFO" originally.

There's a couple points I want to point out here. GAMEINFO is a GZDoom lump.. so it's not that weird to only parse some of the lump. In fact, it doesn't make any sense for it to parse complevel, since that isn't even zdoom related.

I actually am highly against having a new lump for multiple reasons.
1. GAMEINFO is actually a very common lump in Plutonia/TNT/Doom1 wads. For example, I used Plerbtonia for testing, and it already had the lump in the wad. I was surprised to learn that my own RUST doom1 wad also had a GAMEINFO with and iwad specification there. It's not uncommon nowadays for modders to also target zdoom ports, and thus most wads for specific IWADs have GAMEINFO nowadays.
2. I think the main thing you are overlooking is that GAMEINFO has been around for over a decade, and modders have been using it to define exact IWADs. This basically means that adding this feature, we would support lots of legacy wads. The problem with adding a new wad is that modders would have to add yet another lump to their wad. Plus it really makes no sense to create a new lump when many doom wads already have a lump that does the same purpose.
3. I'm not sure what you mean by the "IWAD" lump approach being cleaner... Both approaches use around the same amount of code. Only difference is that "GAMEINFO" is using a parser similar to umapinfo or udmf. When we parse those lumps, we already don't parse things that aren't applicable to dsda-doom (MAPINFO being a big example).

rfomin:
> I think its worth thinking more about this, as supporting a lump just for one field seems wrong. Wouldve prefered if GAMEINFO had fields for iwad, pwads and complvl.

We can add new fields to GAMEINFO, GZDoom would ignore them.

GAMECONF has `pwads` field, `executable` for complevel (although I disagree with it) and interesting `wadtranslation` (converting a WAD's graphics from its original PLAYPAL to a new PLAYPAL included by a PWAD). GAMECONF has also been added to many WADs for compatibility with the KEX port, and we can also add new fields to it if we want.

prboom2/src/dsda/gameinfo.cpp
rfomin: Cosmetic fix:
```diff
diff --git a/prboom2/src/dsda/gameinfo.cpp b/prboom2/src/dsda/gameinfo.cpp
index 1020bae72..afb9e72ce 100644
--- a/prboom2/src/dsda/gameinfo.cpp
+++ b/prboom2/src/dsda/gameinfo.cpp
@@ -39,10 +39,9 @@ void dsda_ParseGameInfoLine(Scanner &scanner) {
   if (!stricmp(scanner.string, "IWAD")) {
     scanner.MustGetToken('=');
     scanner.MustGetString();
-    if (iwadlump) {
+    if (iwadlump)
       Z_Free(iwadlump);
-    }
-  iwadlump = Z_Strdup(scanner.string);
+    iwadlump = Z_Strdup(scanner.string);
   }
 }
 
```

prboom2/src/w_wad.c
rfomin: There are several "namespaces" in WADs, so we can have multiple files with the same name. It is unlikely that someone would create a sprite named `GAMEINFO`, but it's possible.

andrikpowell:
> We can add new fields to GAMEINFO, GZDoom would ignore them.

We could do this... but honestly it feels kinda useless and undermines the whole point of the COMPLVL lump. The main point of the COMPLVL lump was that it was supported by a majority of source ports. So most wads would keep this lump regardless, unless you are expecting other source ports to magically support GAMEINFO as well.

The main appeal for me was that old wads already had GAMEINFO to pick the IWAD from.

> GAMECONF has `pwads` field, `executable` for complevel (although I disagree with it) and interesting `wadtranslation` (converting a WAD's graphics from its original PLAYPAL to a new PLAYPAL included by a PWAD). GAMECONF has also been added to many WADs for compatibility with the KEX port, and we can also add new fields to it if we want.

I realise this is probably just a "me" thing, but I very much dislike GAMECONF. It would be much more code to support it, due to it being in JSON format. I recently asked my discord server if they preferred GAMEINFO or GAMECONF, and they said GAMEINFO due to it being easier to write from scratch.

I think it may possibly be worth supporting at some point... maybe... Although me and many other modders are quite unhappy about how ID24 is formatted and how not user-friendly it is. So I'll admit I'm a bit biased against GAMECONF, but if we eventually support GAMECONF, we definitely should support GAMEINFO.

rfomin:
> We could do this... but honestly it feels kinda useless and undermines the whole point of the COMPLVL lump. 

But we can add an `ALTLOAD` field for additional WADs that will work as we want.

prboom2/src/d_main.c
andrikpowell: I just fixed this in [51d33ec](https://github.com/kraflab/dsda-doom/pull/603/commits/51d33ecf8b772cb558d0c1fc7bdabf4c574b10f7)... I feel kinda dumb :P

andrikpowell:
Figured I'd mention this is ready for an other review.

Now zip files have been fixed, as well GAMEINFO reading via autoload.

This PR should be fully complete now, in regards to the functionality of allowing PWADs to define an IWAD and GAMEINFO support.

IMO any other lumps defining IWADs should be a separate PR.

prboom2/src/d_main.c
```diff
@@ -1839,6 +1882,15 @@ static void D_DoomMainSetup(void)
   lprintf(LO_DEBUG, "W_Init: Init WADfiles.\n");
   W_Init(); // CPhipps - handling of wadfiles init changed
 
+  if (iwadlump != NULL)
+  {
+    lprintf(LO_INFO, "Detected GAMEINFO lump: %s\n", iwadlump);
+    Z_Free(iwadlump);
+  }
+
+  G_ReloadDefaults();    // killough 3/4/98: set defaults just loaded.
```
rfomin: Why did you move `G_ReloadDefault()` here? `HandlePlayback()` may be affected by this. I'd rather not move `G_ReloadDefault()` if possible.

prboom2/src/dsda/gameinfo.cpp
```diff
@@ -0,0 +1,64 @@
...
+  if (lump == LUMP_NOT_FOUND)
+    return;
+
+  Scanner scanner((const char*) W_LumpByNum(lump), W_LumpLength(lump));
+
+  scanner.SetErrorCallback(I_Error);
+
+  while (scanner.TokensLeft())
+    dsda_ParseGameInfoLine(scanner);
+}
```
rfomin: Missing newline at the end of the file. See https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

rfomin (commented):
Seems to work fine, thank you.

prboom2/src/d_main.c
andrikpowell: The main reason why I moved `G_ReloadDefault()` is because the game crashes if it doesn't come after the second `W_Init()` call.

I haven't figured out the reason why. I'd prefer it to be above `Handleplayback()` like before, but currently it's there to avoid the crash.

andrikpowell: Ah I figured it out! See [37979ea](https://github.com/kraflab/dsda-doom/pull/603/commits/37979ea147efcacb4a14516f0bb529ff5575e585).

The crash was caused because `lumpnum` and `lumpinfo` weren't being cleared... Which is a little weird because I know at some point I had that code there.

Anyway, `G_ReloadDefaults()` should be in the same place as it was before now, and I moved the `iwadlump lprintf` to happen a bit later so that it shows just before when the "Detected COMPLVL" text shows.

prboom2/src/dsda/gameinfo.cpp
andrikpowell: Just fixed this through a force push on [ae6be10](https://github.com/kraflab/dsda-doom/pull/603/commits/ae6be10d45eaae6386a24f83837531b160c6d897).

Usually I tend to leave a newline, but I guess I missed it on this one.

Also thanks for the explanation, because I actually didn't know how important it was. I just tended to leave a space at the end, since that's how the other files were formatted.

rfomin (approved):
LGTM

andrikpowell:
Figured I'd mention that I added the "Doom 1 lump detection" code into this PR. Hopefully that's fine...

andrikpowell:
I have reworked how the `IdentifyVersion()` function works based on @Pedro-Beirao 's feedback on Discord. The IWAD detection process should be much clearer now.

PR #604, opened 2025-02-07
Better TC game support + Autoload folders

andrikpowell:
This PR adds better TC game support for:
- Chex Quest (original DOS version)
- Chex Quest 3: Vanilla / Modding Edition
- REKKR
- Freedoom (Phase 1, Phase 2, FreeDM)

It also adds autoload folders for the above TCs.

I haven't added "REKKR: Sunken Land" here because the IWAD is badly formatted, and to get it to work in DSDA-Doom, it needs its own draft PR.

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1450,7 +1450,10 @@ static const char *D_AutoLoadGameBase()
 {
   return hexen ? "hexen-all" :
          heretic ? "heretic-all" :
-         "doom-all";
+         chex ? "chex-all" :
+         freedoom ? "freedoom-all" :
+         !tc_game ? "doom-all":
+         NULL;
 }
 
 #define ALL_AUTOLOAD "all-all"
@@ -1467,11 +1470,14 @@ void D_AutoloadIWadDir()
   LoadZIPsAtPath(autoload_dir, source_auto_load, &autoload_deh_all_queue);
   Z_Free(autoload_dir);
 
-  // common auto-loaded files for the game
-  autoload_dir = GetAutoloadDir(D_AutoLoadGameBase(), true);
-  LoadWADsAtPath(autoload_dir, source_auto_load);
-  LoadZIPsAtPath(autoload_dir, source_auto_load, &autoload_deh_game_queue);
-  Z_Free(autoload_dir);
+  if (D_AutoLoadGameBase())
+  {
+    // common auto-loaded files for the game
+    autoload_dir = GetAutoloadDir(D_AutoLoadGameBase(), true);
+    LoadWADsAtPath(autoload_dir, source_auto_load);
+    LoadZIPsAtPath(autoload_dir, source_auto_load, &autoload_deh_game_queue);
+    Z_Free(autoload_dir);
+  }
 
   // auto-loaded files per IWAD
   autoload_dir = GetAutoloadDir(IWADBaseName(), true);
@@ -1510,11 +1516,14 @@ static void D_AutoloadDehIWadDir()
   D_ProcessDehAutoloadQueue(&autoload_deh_all_queue);
   Z_Free(autoload_dir);
 
-  // common auto-loaded files for the game
-  autoload_dir = GetAutoloadDir(D_AutoLoadGameBase(), true);
-  LoadDehackedFilesAtPath(autoload_dir, false, NULL);
-  D_ProcessDehAutoloadQueue(&autoload_deh_game_queue);
-  Z_Free(autoload_dir);
+  if (D_AutoLoadGameBase())
+  {
+    // common auto-loaded files for the game
+    autoload_dir = GetAutoloadDir(D_AutoLoadGameBase(), true);
+    LoadDehackedFilesAtPath(autoload_dir, false, NULL);
+    D_ProcessDehAutoloadQueue(&autoload_deh_game_queue);
+    Z_Free(autoload_dir);
+  }
 
   // auto-loaded files per IWAD
   autoload_dir = GetAutoloadDir(IWADBaseName(), true);
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -168,8 +168,9 @@ const char *const standard_iwads[]=
   "freedoom1.wad",
   "freedm.wad",
 
-  "hacx.wad",
   "chex.wad",
+
+  "hacx.wad",
   "rekkrsa.wad",
 
   "bfgdoom2.wad",
@@ -1028,7 +1029,7 @@ void AddIWAD(const char *iwad)
     case shareware:
       gamemission = doom;
       if (i>=8 && !strnicmp(iwad+i-8,"chex.wad",8))
-        gamemission = chex;
+        gamemission = tc_chex;
       break;
     case commercial:
       gamemission = doom2;
@@ -1039,7 +1040,7 @@ void AddIWAD(const char *iwad)
       else if (i>=12 && !strnicmp(iwad+i-12,"plutonia.wad",12))
         gamemission = pack_plut;
       else if (i>=8 && !strnicmp(iwad+i-8,"hacx.wad",8))
-        gamemission = hacx;
+        gamemission = tc_hacx;
       break;
     default:
       gamemission = none;
@@ -1605,7 +1606,7 @@ static void EvaluateDoomVerStr(void)
       case retail:
         switch (gamemission)
         {
-          case chex:
+          case tc_chex:
             doomverstr = "Chex(R) Quest";
             break;
           default:
@@ -1628,7 +1629,7 @@ static void EvaluateDoomVerStr(void)
           case pack_tnt:
             doomverstr = "Final DOOM - TNT: Evilution";
             break;
-          case hacx:
+          case tc_hacx:
             doomverstr = "HACX - Twitch 'n Kill";
             break;
           default:
@@ -1885,7 +1886,7 @@ static void D_DoomMainSetup(void)
         ProcessDehFile(NULL, D_dehout(), lump);
       }
     }
-    if (gamemission == chex)
+    if (gamemission == tc_chex)
     {
       int lump = W_CheckNumForName2("CHEXDEH", ns_prboom);
       if (lump != LUMP_NOT_FOUND)
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -169,6 +169,8 @@ const char *const standard_iwads[]=
   "freedm.wad",
 
   "chex.wad",
+  "chex3v.wad",
+  "chex3d2.wad",
 
   "hacx.wad",
   "rekkrsa.wad",
@@ -1036,6 +1038,8 @@ void AddIWAD(const char *iwad)
       gamemission = doom;
       if (i>=8 && !strnicmp(iwad+i-8,"chex.wad",8))
         gamemission = tc_chex;
+      else if (i>=10 && !strnicmp(iwad+i-10,"chex3v.wad",10))
+        gamemission = tc_chex3v;
       else if (i>=11 && !strnicmp(iwad+i-11,"rekkrsa.wad",11))
         gamemission = tc_rekkr;
       else if (i>=13 && !strnicmp(iwad+i-13,"freedoom1.wad",13))
@@ -1049,6 +1053,8 @@ void AddIWAD(const char *iwad)
         gamemission = pack_tnt;
       else if (i>=12 && !strnicmp(iwad+i-12,"plutonia.wad",12))
         gamemission = pack_plut;
+      else if (i>=11 && !strnicmp(iwad+i-11,"chex3d2.wad",11))
+        gamemission = tc_chex3v;
       else if (i>=8 && !strnicmp(iwad+i-8,"hacx.wad",8))
         gamemission = tc_hacx;
       else if ((i>=13 && !strnicmp(iwad+i-13,"freedoom2.wad",13))
@@ -1622,6 +1628,9 @@ static void EvaluateDoomVerStr(void)
           case tc_chex:
             doomverstr = "Chex(R) Quest";
             break;
+          case tc_chex3v:
+            doomverstr = "Chex(R) Quest 3: Vanilla Edition";
+            break;
           case tc_rekkr:
             doomverstr = "REKKR";
             break;
@@ -1647,6 +1656,9 @@ static void EvaluateDoomVerStr(void)
             break;
           case pack_tnt:
             doomverstr = "Final DOOM - TNT: Evilution";
+            break;
+          case tc_chex3v:
+            doomverstr = "Chex(R) Quest 3: Modding Edition";
             break;
           case tc_hacx:
             doomverstr = "HACX - Twitch 'n Kill";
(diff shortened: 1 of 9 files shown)


rfomin:
> It also adds autoload folders for the above TCs.

I think autoload folders for REKKR is too much. Are there any mods for REKKR? Also there are some other IWADs like Hacx, Harmony, etc. then why are we ignoring them?

andrikpowell:
> > It also adds autoload folders for the above TCs.
> 
> I think autoload folders for REKKR is too much. Are there any mods for REKKR? Also there are some other IWADs like Hacx, Harmony, etc. then why are we ignoring them?

I guess the main thing I'd say about this is that I think it's annoying when I put a widescreen asset wad in doom-all and it loads with Chex or REKKR.

Regarding REKKR, I plan on adding Sunken Land later on, which would use the same autoload. But also this makes our lives easier in the future if REKKR pwads start coming out.

Also I wasn't ignoring Hacx, that was added in the autoload as well.
Currently dsda doesn't support Harmony, but we could add support if it's iwad.

andrikpowell:
I guess if you're really concerned about the amount of autoload folders, we could have like a "tcgame-all" folder, although I think that defeats the purpose of autoload folders in general (since I can't think of a reason why you'd want to apply something for all tc games, that just wasn't already in "all-all")

rfomin:
> Currently dsda doesn't support Harmony, but we could add support if it's iwad.

The newer version is working as IWAD: https://www.doomworld.com/idgames/levels/doom2/Ports/g-i/harmonyc But we don't have to do anything to support it. There are probably other IWADs that I don't know about.

andrikpowell:
I have reworked the autoload folders to only include "doom-all", "hexen-all", "heretic-all", "chex-all", and "freedoom-all". Any other iwad such as HACX or REKKR will just not have an "*-all" folder.

> > Currently dsda doesn't support Harmony, but we could add support if it's iwad.
> 
> The newer version is working as IWAD: https://www.doomworld.com/idgames/levels/doom2/Ports/g-i/harmonyc But we don't have to do anything to support it. There are probably other IWADs that I don't know about.
Circling back to Harmony, we can't currently support it due to an error thrown in "R_InitTextures()". Apparently there are some bad/missing textures in it... Although not as bad as REKKR: Sunken Land.

I tested disabling the abort message about textures, but it's too unstable at the moment.

Testing REKKR: Sunken Land, there's a 50/50 chance that the game crashes when entering a demo or starting a level in software. In OpenGL, it always crashes.

I'm going to assume the cause of the crash has something to do with OpenGL caching the textures, which due to the 900+ missing textures just causes DSDA to crash.

Either way, adding support for any IWAD (and IWADs with bad texture defs) should be a separate PR from this one.

prboom2/src/doomdef.h
```diff
@@ -76,11 +76,20 @@ typedef enum {
...
-  chex,         // Chex Quest
+  tc_hacx,      // HACX - Twitch 'n Kill
+  tc_chex,      // Chex Quest
+  tc_chex3v,    // Chex Quest 3: Vanilla/Modding Edition
+  tc_rekkr,     // REKKR
+  tc_freedoom,  // Freedoom
   none
 } GameMission_t;
 
+#define tc_game   !((gamemission == doom) || (gamemission == doom2) || (gamemission == pack_tnt) || (gamemission == pack_plut) || (gamemission == pack_nerve))
```
rfomin: You can just write `gamemission >= tc_hacx && gamemission <= tc_freedoom`.
Why do we need these macros? I think it's more readable without them.

andrikpowell: Funny enough, I didn't think of doing `gamemission <= pack_nerve` for `tc_game`... which is actually a bit more simpler. I plan on using the macro `tc_game` for more things, and I think writing out 5 game modes everytime would be very annoying.

As for why I don't write it `gamemission >= tc_hacx && gamemission <= tc_freedoom`, I prefer to specifically target the original gamemissions as they won't ever change order. Depending on tc support added, your method would always require `tc_freedoom` being at the end, which I just would rather not deal with.

That being said, some of the macros under are probably unneeded (specifically `hacx`, `freedoom`, and `chex`). I do more stuff with Chex and in the Freedoom in Nyan Doom, but for DSDA it's not required. I just really don't want to have to write `gamemission == [insert game mission here]` every single time I want to refer to `chex` or `chex_exe`. I find the macros much easier to understand, and less code to type for tc game support.

rfomin: > As for why I don't write it `gamemission >= tc_hacx && gamemission <= tc_freedoom`, I prefer to specifically target the original gamemissions as they won't ever change order. Depending on tc support added, your method would always require `tc_freedoom` being at the end, which I just would rather not deal with.

Another solution:
```c
typedef enum {
  none = -1,
  doom,         // DOOM 1
  doom2,        // DOOM 2
  pack_tnt,     // TNT mission pack
  pack_plut,    // Plutonia pack
  pack_nerve,   // No Rest For The Living
  tc_game,
  tc_hacx,      // HACX - Twitch 'n Kill
  tc_chex,      // Chex Quest
  tc_chex3v,    // Chex Quest 3: Vanilla/Modding Edition
  tc_rekkr,     // REKKR
  tc_freedoom,  // Freedoom
} GameMission_t;
...
if (gamemission > tc_game)
```

In general, I think we should avoid macros unless absolutely necessary. Macros in C are not type safe and have side effects https://stackoverflow.com/questions/14041453/why-are-preprocessor-macros-evil-and-what-are-the-alternatives

andrikpowell: Wouldn't it be really confusing to distinguish between what is TC and what is not between `(gamemission > tc_game)` and `(gamemission < tc_game)`.

> In general, I think we should avoid macros unless absolutely necessary. Macros in C are not type safe and have side effects https://stackoverflow.com/questions/14041453/why-are-preprocessor-macros-evil-and-what-are-the-alternatives

Ok, instead of macros can we just use `dboolean`s then?

rfomin: > Wouldn't it be really confusing to distinguish between what is TC and what is not between `(gamemission > tc_game)` and `(gamemission < tc_game)`.

I think it's okay.

> Ok, instead of macros can we just use `dboolean`s then?

Yeah, sure.

andrikpowell: Converted macros to dbooleans

prboom2/src/d_main.c
```diff
@@ -1068,6 +1068,14 @@ void AddIWAD(const char *iwad)
   if (gamemode == indetermined)
     //jff 9/3/98 use logical output routine
     lprintf(LO_WARN,"Unknown Game Version, may not work\n");
+
+  // Set up gamemission booleans
+  tc_game = (gamemission > pack_nerve) ? true : false;
+  freedoom = (gamemission == tc_freedoom) ? true : false;
+  chex_exe = (gamemission == tc_chex) ? true : false;
+  chex = ((gamemission == tc_chex)
+       || (gamemission == tc_chex3v)) ? true : false;
```
ceski-1: No need for `? true : false` right?

andrikpowell: > No need for `? true : false` right?

Yeah, I guess I got a little confused moving away from macros. the `? true : false` is unneeded. Not sure why I put that in :/ Fixed.

andrikpowell:
One thing to mention here, I was experimenting with including an "Iwads" folder to help "clean" up the autoload folder.

I mention this because although the autoload folder can be a bit simple, it can also get a bit crowded. Right now you get all of the `*-all` folders (all, doom-all, hexen-all, chex-all, etc) mixed in along with the specific iwad folders (doom.wad, doom2.wad, rekkrsa.wad, chex.wad, etc).

The result is a folder that can look like this:
```
all-all
chex.wad
chex3.wad
chex-all
doom.wad
doom2.wad
doom-all
freedoom1.wad
freedoom2.wad
freedoom-all
heretic.wad
heretic1.wad
heretic-all
hexdd.wad
hexen.wad
hexen-all
plutonia.wad
rekkrsa.wad
tnt.wad
```

Which can be a bit of a mess... It may be worth including an `Iwads` folder, so that in the main folder we get something like:
```
all-all
chex-all
doom-all
freedoom-all
heretic-all
hexen-all
iwads
```

and then the `iwads` folder gives us all the specific iwads:
```
chex.wad
chex3.wad
doom.wad
doom2.wad
freedoom1.wad
freedoom2.wad
freedoom-all
heretic.wad
heretic1.wad
hexdd.wad
hexen.wad
plutonia.wad
rekkrsa.wad
tnt.wad
```

This is mostly just a food-for-thought.

prboom2/src/doomdef.h
```diff
@@ -76,11 +76,19 @@ typedef enum {
...
   pack_plut,    // Plutonia pack
   pack_nerve,   // No Rest For The Living
-  hacx,         // HACX - Twitch 'n Kill
-  chex,         // Chex Quest
+  tc_hacx,      // HACX - Twitch 'n Kill
+  tc_chex,      // Chex Quest
+  tc_chex3v,    // Chex Quest 3: Vanilla/Modding Edition
+  tc_rekkr,     // REKKR
+  tc_freedoom,  // Freedoom
   none
```
rfomin, line 84: Move `none` to the top of the `enum`:
```c
enum
{
    none = -1,
    ...
}
```
Otherwise `tc_game` is also true in the case `none`.

prboom2/src/d_main.c
```diff
@@ -1048,6 +1068,14 @@ void AddIWAD(const char *iwad)
   if (gamemode == indetermined)
     //jff 9/3/98 use logical output routine
     lprintf(LO_WARN,"Unknown Game Version, may not work\n");
+
+  // Set up gamemission booleans
+  tc_game = (gamemission > pack_nerve);
+  freedoom = (gamemission == tc_freedoom);
+  chex_exe = (gamemission == tc_chex);
+  chex = ((gamemission == tc_chex)
+       || (gamemission == tc_chex3v));
```
rfomin: Is it not too late to convince you to remove those variables? 😄 For me `gamemission == tc_freedom` is more readable than `freedoom`.

rfomin (commented):
> One thing to mention here, I was experimenting with including an "Iwads" folder to help "clean" up the autoload folder.

Autoloading is also supported in Chocolate/Crispy Doom and Woof (and maybe some other ports). I don't think it's worth changing this.

prboom2/src/doomdef.h
andrikpowell: > Move `none` to the top of the `enum`:
> 
> ```c
> enum
> {
>     none = -1,
>     ...
> }
> ```
> 
> Otherwise `tc_game` is also true in the case `none`.

I consider any game that isn't doom (1, 2, plutonia, tnt, nerve) a TC, so don't we want `none` to be considered a TC.

I'm actually not sure what an IWAD would have to do to be considered `none` anyway...

prboom2/src/d_main.c
andrikpowell: I don't think you can convince me on the `tc_game` variable... As `(gamemission > pack_nerve)` is definitely not very clear... but the other variables are mostly only used for the autoload folders, so it wouldn't be too difficult I guess.

Again, in Nyan Doom I do more with this, so for Nyan I intend to keep the variables, as it's easier to type `chex` over `((gamemission == tc_chex) || (gamemission == tc_chex3v))`

I still think this looks way uglier:
```
  return hexen ? "hexen-all" :
         heretic ? "heretic-all" :
         (gamemission == tc_chex ||
         gamemission == tc_chex3v) ? "chex-all" :
         (gamemission == tc_freedoom) ? "freedoom-all" :
         !tc_game ? "doom-all":
         NULL;
```

Compared to:
```
  return hexen ? "hexen-all" :
         heretic ? "heretic-all" :
         chex ? "chex-all" :
         freedoom ? "freedoom-all" :
         !tc_game ? "doom-all":
         NULL;
```

Most of the chex code in DSDA only applies to the classic DOS version of chex aka `chex_exe`... I do think the code could get possibly more confusing using `(gamemission == tc_chex)` as you'd have to remember this is only for the DOS chex and not chex3v. Personally `chex` vs `chex_exe` seems way more clear to me... But I guess I could begrudgingly cave on that part if you really want to axe those variables.

prboom2/src/doomdef.h
andrikpowell: If `none` is what I think it means, if later down the line we add support for any iwad to be loaded, I want Harmony or REKKR Sunken Land (which may not be in the gamemission list) to be considered a `tc_game`, and they would be categorised `none`.

rfomin: > I'm actually not sure what an IWAD would have to do to be considered none anyway...

I haven't tested it, but it will probably exit with an error. Which is correct behavior in my opinion.

prboom2/src/d_main.c
rfomin: > I still think this looks way uglier:

Looks better to me because I don't have to look up variables.

>  Personally `chex` vs `chex_exe` seems way more clear to me...

I think it's the same, readers still need to look up how these variables are set.

prboom2/src/doomdef.h
andrikpowell: > I haven't tested it, but it will probably exit with an error. Which is correct behavior in my opinion.

`gamemission == none` currently doesn't exit with an error. The only thing that stops a `gamemission == none` is if the texture lump is incorrectly formatted.

I disagree with this assessment, as if we ever want to load IWADs that are not explicitly defined in the IWAD list, they should be "unknown" or essentially "none". We don't want unknown TCs pulling from `doom-all` autoload or `!tc_game` logic. Only Doom gametypes should pull from doom logic.

Keep in mind `gamemission` is different from `gamemode`, which makes sense to bomb out if it's `indetermined`.

It seems that currently most doom `gamemission` is only used for plutonia, tnt, nerve specific code. Only line that actually affects something is this SSG code:
```
  // Can't select the super shotgun in Doom 1.
  if (weapon == wp_supershotgun && gamemission == doom)
  {
    return false;
  }
```
  
Which in my opinion should be `gamemode != commercial` instead.

andrikpowell: Actually at the moment I think it's actually impossible for `gamemission == none`, since it's automatically set to `doom` or `doom2` in `AddIWAD()`, and is only changed if other conditions are found.

Although since doom gametypes are tied to autoload, I still think it's worth rethinking this if we end up adding support for any IWAD. Because it makes no sense to use `doom-all` autoload if the game isn't doom.

Aka `gamemission == none` is only possible if `gamemode == indetermined`.

andrikpowell: For now, I'll prob do the `none = -1` treatment, but I think if we do allow custom IWADs, we should add a `tc_unknown` for that case

andrikpowell: Actually I'm gonna back down on `none` in that it should be considered a TC, so `none` being considered a `tc_game` is correct logic.

For example, the only thing `tc_game` is being used for right now is autoload and not showing the dynamic help screen if it's a TC.

`none` should not use `doom-all` (which would be the case if we follow your logic), nor should it show the dynamic help screen. The main reason I set that condition is because TCs can have custom weapon names which the dynamic help screen prints. It makes more sense to just use the TC's HELP screen instead in that case.

rfomin: Okay, thanks.

rfomin (approved):
LGTM

Pedro-Beirao:
Thank you both!

PR #605, opened 2025-02-09
Fix Chainsaw Vanilla Weapon Switch Behaviour

andrikpowell:
So this is a vanilla demo-breaking issue... But it only affects Vanilla WADs with a custom dehacked weapon in the chainsaw slot that uses ammo.

The issue is in regards to the weapon switch behaviour when the weapon in the chainsaw slot uses ammo, and all other ammo types have been used up.

Currently in DSDA-Doom, if you run out of ammo and use up the ammo in the chainsaw slot, you will switch to the fist after trying to fire... This is incorrect, as in Vanilla it should be impossible to access the fist if the player has the chainsaw (berserk allows access to the fist... however it will never auto-switch to the fist when running out of ammo).

Test the same behaviour in DOS or Chocolate Doom, and you will see that holding the fire button, it will constantly try and switch to the chainsaw weapon, and never the fist.

The example WAD that can best showcase this behaviour is [RUST](https://www.doomworld.com/forum/topic/144559). Just warp to the first map, type IDCHOPPERS, and see the behavour.

_Note: It's worth noting that the Boom weapon switch behaviour is most likely correct, as Boom rewrote the weapon switch code. This is just a fix for vanilla complevels._

### DIFF
--- a/prboom2/src/p_pspr.c
+++ b/prboom2/src/p_pspr.c
@@ -303,6 +303,10 @@ int P_SwitchWeapon(player_t *player)
   newweapon = currentweapon;
   i = NUMWEAPONS + 1;   // killough 5/2/98
 
+  // Fix weapon switch logic in vanilla (example: chainsaw with ammo)
+  if (demo_compatibility)
+    currentweapon = newweapon = wp_nochange;
+
   // killough 2/8/98: follow preferences and fix BFG/SSG bugs
 
   do


andrikpowell:
Here is a [test wad](https://drive.google.com/file/d/10tTGfL3ux1ZK21-_1lBhdgKyQpSJscmb/view?usp=sharing) (with an internal dehacked lump) with a demo showing the desync. (I've also included a deh file for DOS testing)

The demo plays correctly in Choco, Crispy, and DOS. In current DSDA-Doom the demo will desync. However this PR allows the demo to work correctly.

This is just a standard Doom 2 (complevel 2) demo with a custom chainsaw weapon that uses ammo. The zombiemen have been dehacked to not drop clips.

Pedro-Beirao (approved):

Pedro-Beirao:
@fabiangreffrath need this on Woof as well

fabiangreffrath:
I'll check it, thanks!

fabiangreffrath:
I don't think this is entirely correct. In fact, there *is* a chance to switch to the fist in Vanilla Doom, but the check for the Chainsaw comes earlier:

https://github.com/chocolate-doom/chocolate-doom/blob/d9593fd7f834febabb2822106dbe15b7277ed2f7/src/doom/p_pspr.c#L220

However, the difference between Vanilla code and Boom code is that Vanilla explicitly switches to the chainsaw if it is available, whereas Boom switches to the fist:

https://github.com/chocolate-doom/chocolate-doom/blob/d9593fd7f834febabb2822106dbe15b7277ed2f7/src/doom/p_pspr.c#L202-L205

fabiangreffrath:
This breaks non-Vanilla complevels, as it now switches to the pistol once the "chainsaw" ammo runs out.

fabiangreffrath:
I suggest this change:
```patch
--- a/prboom2/src/p_pspr.c
+++ b/prboom2/src/p_pspr.c
@@ -314,6 +314,8 @@ int P_SwitchWeapon(player_t *player)
         // fallthrough
       case 0:
         newweapon = wp_fist;
+        if (player->weaponowned[wp_chainsaw] && demo_compatibility)
+          newweapon = wp_chainsaw;
         break;
       case 2:
         if (player->ammo[am_clip])
```

Pedro-Beirao:
> This breaks non-Vanilla complevels, as it now switches to the pistol once the "chainsaw" ammo runs out.

Unless Im misunderstanding something, it cant break non-Vanilla since the behaviour is the same as before when demo_compatibility is false.
But I do prefer your approach.

Note that this PR did pass regression testing.

andrikpowell:
> This breaks non-Vanilla complevels, as it now switches to the pistol once the "chainsaw" ammo runs out.

This is not true.

I will admit that the specific way that the condition is phrased is a little confusing, and that's why it took me a while to get to this. However I did test this behaviour with the same weapon in multiple complevels, over different situations (to avoid things like switching to the pistol). I prefer how I phrased the condition for multiple reasons.

Let's look at why this code works, and why I wrote it as such. In this case, we are checking the condition `!(demo_compatibility && player->weaponowned[wp_chainsaw])`.

I'll clarify that `demo_compatibility` is a macro for "complevel < boom". Because of how it's set up, the condition inside the `!` is "vanilla + has chainsaw". Since "vanilla" would always be false in higher complevels, and there is a `!` in front, the condition always returns true in complevels above vanilla (aka what it was before).

This works correctly because the behaviour we want for vanilla is to always skip the fist if the player has a chainsaw. And so if "vanilla + has chainsaw" is true, the `!` in front would turn the "true" to "false", allowing a fallthrough down to the chainsaw.

> I suggest this change:

I dislike this change here because 1) here we are choosing the fist and then the chainsaw (maybe this isn't an issue, but it feels bad to do it this way), and 2) I don't like forcing the chainsaw weapon switch in the "fist" block, and prefer to do it via another fallthrough to the "chainsaw" block, similar to how the berserk fallthrough works.

fabiangreffrath:
Would you please do `dsda-doom -iwad doom.wad -file RUST1.wad -warp 1 -complevel 9`, then type `IDCHOPPERS`, then push `1` and hit the fire button three times?

andrikpowell:
> Would you please do `dsda-doom -iwad doom.wad -file RUST1.wad -warp 1 -complevel 9`, then type `IDCHOPPERS`, then push `1` and hit the fire button three times?

This works the exact same way as `complevel 9` works in the current DSDA-Doom. On the third fire, the ammo goes negative (which I guess is a weird boom difference), and then it will try and switch to the pistol, and then back to the chainsaw. (trigger weaponswitch loop)

While the behaviour is weird, it is consistent with or without the new code in this PR. I'm unsure if this boom switch code logic is incorrect or not, but this PR isn't changing this logic and it is working as "intended" (aka this is how it worked before this PR).

fabiangreffrath:
Right, this is not this patch's fault. Sorry, I was mistaken.

The culprit is that Boom allows ammo to become negative, which is still non-zero, and thus triggers the `if (player->ammo[am_clip])` condition.

However, I think the code is still confusing, because it only selects a `newweapon`  before calling `break` if the newly introduced condition is met - which just feels odd.

fabiangreffrath:
> I don't like forcing the chainsaw weapon switch in the "fist" block, and prefer to do it via another fallthrough to the "chainsaw" block, similar to how the berserk fallthrough works.

Alright, maybe you could rearrange it that way.

andrikpowell:
> Alright, maybe you could rearrange it that way.

I think there's a misunderstanding here :/  This PR is set up as a passthrough, unless I'm mistaken.

fabiangreffrath:
> I think there's a misunderstanding here :/ This PR is set up as a passthrough, unless I'm mistaken.

No, the `break` is called unconditionally, even if no `newweapon` has been set. Even then, it would fall through to `newweapon = wp_pistol`.

andrikpowell:
> > I think there's a misunderstanding here :/ This PR is set up as a passthrough, unless I'm mistaken.
> 
> No, the `break` is called unconditionally, even if no `newweapon` has been set. Even then, it would fall through to `newweapon = wp_pistol`.

Ah ok, now I'm starting to get confused on how this PR works. Technically we want it to go through the weapon switching priority in the code here:
https://github.com/chocolate-doom/chocolate-doom/blob/d9593fd7f834febabb2822106dbe15b7277ed2f7/src/doom/p_pspr.c#L220

Thus the following weapon / ammo check:
1. plasmagun / cell ammo [6]
2. SSG / shell ammo + doom2 [9]
3. chaingun / bullet ammo [4]
4. shotgun / shell ammo [3]
5. pistol / bullet ammo [2]
6. chainsaw [8] (will loop here if above conditions are missing)
7. rocket launcher / rocket ammo [5]
8. BFG / cell ammo [7]
9. fist [0/1]

So in vanilla, the behaviour I was experiencing was getting stuck at 6 if the player runs out of ammo... but current DSDA-Doom behaviour goes to the fist if the chainsaw is out of ammo...

The main problem here is I don't understand how this prefer switch block works... And I'm even more confused about how the "case 1" for the berserk works.

According to the `weapon_preferences`, it is set correctly: `{6, 9, 4, 3, 2, 8, 5, 7, 1, 0}`

So then why does it not respect staying with the chainsaw (with no other ammo) in the weapon switch chain?

It seems like this PR does fix it, but I don't actually understand why it doesn't stay on the chainsaw, and just decides to go with the fist when it runs out of ammo.

fabiangreffrath:
The `while` loop continues because of the `newweapon == currentweapon` condition, which is `true` when attempting to switch from the chainsaw to the chainsaw. Thus it continues and ends up with the fist as its last resort.

andrikpowell:
> The `while` loop continues because of the `newweapon == currentweapon` condition, which is `true` when attempting to switch from the chainsaw to the chainsaw. Thus it continues and ends up with the fist as its last resort.

Yeah I've been experimenting with printing to console to figure out what's going on, and I see that it's really the `newweapon == currentweapon` that's causing the chainsaw to not stay on the chainsaw.

> I suggest this change:
> 
> ```diff
> --- a/prboom2/src/p_pspr.c
> +++ b/prboom2/src/p_pspr.c
> @@ -314,6 +314,8 @@ int P_SwitchWeapon(player_t *player)
>          // fallthrough
>        case 0:
>          newweapon = wp_fist;
> +        if (player->weaponowned[wp_chainsaw] && demo_compatibility)
> +          newweapon = wp_chainsaw;
>          break;
>        case 2:
>          if (player->ammo[am_clip])
> ```

This right here seems like the way to go... Although that begs the question on why the chainsaw was being chosen before through the current PR code?

I thought at first it was a fall through, but I was mistaken... So I'm a bit confused on how it ended up with the chainsaw. Was the function just bombing out?

One thing that I'm slightly concerned by is what happens if the fist and chainsaw weapon slots both use different ammo? What would the behaviour be in that instance? I hope this fix wouldn't mess that up in some way.

fabiangreffrath:
> This right here seems like the way to go... Although that begs the question on why the chainsaw was being chosen before through the current PR code?

Because of the `currentweapon = player->readyweapon` initialization at the beginning of that function.

andrikpowell:
> Because of the `currentweapon = player->readyweapon` initialization at the beginning of that function.

Oh I think I understand, so if it doesn't find a `newweapon` in the switch block, the switch block basically does nothing in that instance. and so it goes with the `newweapon = currentweapon` definition before the switch block.

And actually in a case that the "fist" weapon does have ammo (and the player had a chainsaw), it's actually better to do it your way, as the way I had it firing the fist would actually go the "fist" instead of the "chainsaw", which would actually be incorrect.

fabiangreffrath:
> This right here seems like the way to go... Although that begs the question on why the chainsaw was being chosen before through the current PR code?

Because of the `currentweapon = player->readyweapon` initialization at the beginning of that function.

fabiangreffrath:
Vanilla actually checks for `player->pendingweapon == wp_nochange` in the `while` loop, where `player->pendingweapon` coresponds to `newweapon` in the Boom code.

fabiangreffrath:
Next idea:

```patch
--- a/prboom2/src/p_pspr.c
+++ b/prboom2/src/p_pspr.c
@@ -303,6 +303,11 @@ int P_SwitchWeapon(player_t *player)
   newweapon = currentweapon;
   i = NUMWEAPONS + 1;   // killough 5/2/98
 
+  if (demo_compatibility)
+  {
+    currentweapon = newweapon = wp_nochange;
+  }
+
   // killough 2/8/98: follow preferences and fix BFG/SSG bugs
 
   do
```

andrikpowell:
> Next idea:
> 
> ```diff
> --- a/prboom2/src/p_pspr.c
> +++ b/prboom2/src/p_pspr.c
> @@ -303,6 +303,11 @@ int P_SwitchWeapon(player_t *player)
>    newweapon = currentweapon;
>    i = NUMWEAPONS + 1;   // killough 5/2/98
>  
> +  if (demo_compatibility)
> +  {
> +    currentweapon = newweapon = wp_nochange;
> +  }
> +
>    // killough 2/8/98: follow preferences and fix BFG/SSG bugs
>  
>    do
> ```

I like this solution alot better!

However, I'd like to create another test wad with the fist using another ammo type (in conjunction with a chainsaw ammo weapon), just to triple check that everything is working as intended.

fabiangreffrath:
One thing I noticed is that the raising state of the chainsaw-replacement weapon is a lot quicker in dsda-doom than in any other port. Is that somehow intended?

andrikpowell:
> One thing I noticed is that the raising state of the chainsaw-replacement weapon is a lot quicker in dsda-doom than in any other port. Is that somehow intended?

This is not intended. It's an issue with DSDA-Doom and how it deals with weapon switching.... Sometimes the raise state will skip the raising animation. Note that it's only cosmetic, and doesn't affect demos or gameplay... but it is something that should definitely be fixed.

I just haven't found what is causing the issue to do a fix yet.

andrikpowell:
Here is a second [test wad/demo](https://drive.google.com/file/d/1WIAyNWsTQgzlk15DZSJ-oV0E75d3meep/view?usp=sharing) utilising both the chainsaw and fist weapons with different ammo types. It's slightly more boring, but it does include alot of weapon switching and testing of the "CheckAmmo" behaviour.

I don't know if we want to use this for testing, but from a quick test, the `currentweapon = newweapon = wp_nochange` fix works just like Vanilla.

andrikpowell:
Just fyi, I did a force push with the new code for the previous test wad/demo... Probably should've updated the code first :P

fabiangreffrath:
> This is not intended. It's an issue with DSDA-Doom and how it deals with weapon switching.... Sometimes the raise state will skip the raising animation. Note that it's only cosmetic, and doesn't affect demos or gameplay... but it is something that should definitely be fixed.

It's mitigated when "Weapon Attack Alignment" is set to either "Off" or "Horizontal".

fabiangreffrath:
I guess the chainsaw-replacement-weapon's raise state is shorter in tics than its raising animation would take.

andrikpowell:
> > This is not intended. It's an issue with DSDA-Doom and how it deals with weapon switching.... Sometimes the raise state will skip the raising animation. Note that it's only cosmetic, and doesn't affect demos or gameplay... but it is something that should definitely be fixed.
> 
> It's mitigated when "Weapon Attack Alignment" is set to either "Off" or "Horizontal".

Yeah it seems that issue is with the "weapon centered / bobbing" settings with the '&psp_sy' condition: https://github.com/kraflab/dsda-doom/blob/4f57712bf815c1145dcc497f08499cf18a205a30/prboom2/src/r_things.c#L1101C1-L1106C8 .

The weapon does have a strange raise state:
`frame 1 - 0 duration - A_Punch`
`frame 2 - 2 duration - A_Raise (loop state until ready state)`

It's interesting that the weapon's raise state works correctly in Crispy with "weapon centered / bobbing" turned on... and apparently not with the code in DSDA (which was also based on the Crispy).

fabiangreffrath:
Yes, I think it's the `state != winfo->upstate` part.

andrikpowell:
> Yes, I think it's the `state != winfo->upstate` part.

This should be fixed in #609 . Further discussion on weapon alignment should probably move over there.

Pedro-Beirao (approved):
I find this solution more confusing than simply swapping to the chainsaw after `newweapon = wp_fist`, but it is more correct I guess.

Pedro-Beirao:
After further testing, 3 demos desynched. All from the same wad.
They also desync on vanilla.

https://dsdarchive.com/wads/mohawks2
```
2022-06-11|doom2|mohawks2|Map 10|UV Max|10:16.34|Peccatum Mihzamiz|-iwad doom2.wad -file mo2bonus.wad mohawks2.wad
2023-04-04|doom2|mohawks2|Map 10|UV Max|4:35.29|Phoenyx|-iwad doom2.wad -file mo2bonus.wad mohawks2.wad
2022-06-11|doom2|mohawks2|Map 13|UV Max|9:32.46|Peccatum Mihzamiz|-iwad doom2.wad -file mo2bonus.wad mohawks2.wad
```

PR #609, opened 2025-02-16
Weapon Alignment Fixes

andrikpowell:
So this took me a little while to figure out. Mostly because most of the original code here was taken from Crispy Doom, but obviously there was something wrong with some WADs when "centered" or "bobbing" was turned on for "Weapon Attack Alignment".

Essentially the issue was with the `A_Raise` and `A_Lower` state checking, in that it seemed that new custom weapons would seem to skip the `A_Raise` animation, or never play the `A_Lower` animation of death. (when the player is pressing down the fire button).

It seems the culprit was how the states were being checked (example: `state != winfo->downstate`). I think my best guess is that this works for vanilla weapons because it's checking the original states... but once custom dehacked is added in the mix, the states don't line up, and thus the code that allows `A_Raise` or `A_Lower` fails.

Examples you can check the behaviour in:
- [RUST](https://www.doomworld.com/forum/topic/144559) - the new chainsaw weapon will skip the `A_Raise` animation (before this fix)
- [200 Line Massacre](https://www.doomworld.com/idgames/levels/doom2/megawads/200lnm) - Upon death, the player weapon will stay on screen and never lower (before this fix)

I also added a fix for Hexen Melee Weapons from Crispy Hexen. This fixes melee weapons for both when weaponbob is disabled and when centered/bobbing alignment is turned on.

You can test the Hexen fix using the first weapon for the Fighter (see what happens after hitting an enemy 3 times)

### DIFF
--- a/prboom2/src/r_things.c
+++ b/prboom2/src/r_things.c
@@ -1063,7 +1063,12 @@ static void R_DrawPSprite (pspdef_t *psp)
   {
     int weapon_attack_alignment = dsda_IntConfig(dsda_config_weapon_attack_alignment);
 
-    if (!dsda_WeaponBob())
+    // [crispy] don't align swiping weapons
+    const dboolean swiping_weapon = hexen && (viewplayer->pclass == PCLASS_FIGHTER ||
+                                             (viewplayer->pclass == PCLASS_CLERIC &&
+                                             viewplayer->readyweapon == wp_first));
+
+    if (!dsda_WeaponBob() && !(swiping_weapon && viewplayer->attackdown))
     {
       static fixed_t last_sy = 32 * FRACUNIT;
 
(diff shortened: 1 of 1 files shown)


prboom2/src/r_things.c
```diff
@@ -1061,34 +1061,25 @@ static void R_DrawPSprite (pspdef_t *psp)
...
-    }
-    else
-    {
-      winfo = &weaponinfo[viewplayer->readyweapon];
-    }
-
-    state = viewplayer->psprites[ps_weapon].state - states;
-    weapon_attack_alignment = dsda_IntConfig(dsda_config_weapon_attack_alignment);
+    // [crispy] don't align swiping weapons
+    const dboolean swiping_weapon = viewplayer->pclass == PCLASS_FIGHTER ||
```
fabiangreffrath: Which value does `player->pclass` have in Doom? Could we rather guard this by a `if (raven && ...)` condition?

andrikpowell: `player->pclass` is exclusive to Hexen, but I added a `raven &&` just to be safe.

andrikpowell: or tbh it may be better to use `hexen &&` instead. Heretic doesn't use `pclass` either

Pedro-Beirao (approved):
Looks right

PR #613, opened 2025-02-24
Expand features bit array size to 128

Pedro-Beirao:
With https://github.com/kraflab/dsda-doom/pull/586, we are introducing 2 new features that need to be tracked. This will exceed the array size of 64, so we need to increase it to 128.

There is no unit128 in 32-bit targets, so this PR uses an array of uint64. It can easily be increased in the future as well.

### DIFF
--- a/prboom2/src/dsda/features.h
+++ b/prboom2/src/dsda/features.h
@@ -86,12 +86,16 @@ typedef enum {
   // 63
 } dsda_feature_flag_t;
 
-#define FEATURE_SIZE 8
+// Size of the uint64 array of features
+#define FEATURES_PARTS 2
+
+#define FOR_FEATURES_PART for (int f = 0; f < FEATURES_PARTS; f++) {
+#define END_FEATURES_PART }
 
 void dsda_TrackFeature(int feature);
 void dsda_ResetFeatures(void);
-uint64_t dsda_UsedFeatures(void);
-void dsda_MergeFeatures(uint64_t source);
+uint64_t *dsda_UsedFeatures(void);
+void dsda_MergeFeatures(uint64_t *source);
 void dsda_CopyFeatures(byte* result);
-void dsda_CopyFeatures2(byte* result, uint64_t source);
+void dsda_CopyFeatures2(byte* result, uint64_t *source);
 char* dsda_DescribeFeatures(void);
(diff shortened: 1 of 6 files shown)


rfomin:
"Bitmap" is an integer with flags?

How about just
```c
uint64_t features1;
uint64_t features2;
```

Pedro-Beirao:
That was the first obvious solution, but it would be more annoying to work with in my opinion (needing to pass 2 variables to all these functions instead of just an array) and would need a `featuress3` in the future if we ever go beyond 128.

Pedro-Beirao:
> "Bitmap" is an integer with flags?

Yes. Is it not the correct term for it? Seemed right to me.

mikeday0:
> 
> Yes. Is it not the correct term for it? Seemed right to me.

The name I have always heard for this kind of structure is [bit field](https://en.m.wikipedia.org/wiki/Bit_field).

rfomin:
Interesting set of macros for working with "bit arrays" in C: https://c-faq.com/misc/bitsets.html

Looks good, but I don't understand why the "features" code is so complicated. Maybe we can convert `uint64_t *features` to just a `char` array?

Pedro-Beirao:
I agree that its better to use an array of `8bits` than `64bits`.
That link does indeed have some nice macros.

So made a new commit.
And went with `byte` instead of `char` since some parts of this code were already using it.

prboom2/src/dsda/features.h
```diff
@@ -86,12 +86,22 @@ typedef enum {
...
+#define BITSET(a, b) ((a)[BITSLOT(b)] |= BITMASK(b))
+#define BITCLEAR(a, b) ((a)[BITSLOT(b)] &= ~BITMASK(b))
+#define BITTEST(a, b) ((a)[BITSLOT(b)] & BITMASK(b))
+#define BITNSLOTS(nb) ((nb + 8 - 1) / 8)
+
+#define FEATURE_SIZE 128
+#define FEATURE_SLOTS BITNSLOTS(FEATURE_SIZE)
+
+#define FOR_FEATURE_SLOT for (int f = 0; f < FEATURE_SLOTS; f++) {
+#define END_FEATURE_SLOT }
```
rfomin: Why do we need `FOR_FEATURE_SLOT` and `END_FEATURE_SLOT` macros? I think `for` loop is more readable.

Pedro-Beirao: I think its easier to understand using macros since we use this loop so many times. But sure, made a new commit to remove the macros.

rfomin (approved):
LGTM, thanks.

PR #614, opened 2025-02-27
Add Back "Classic Zoom Levels"

andrikpowell:
This adds back the functionality of being able to shrink/grow the screensize via "-"/"+".

I've also added a few improvements:
- fixed weapon interpolation causing the weapon to "wiggle" during screensize changes
- Added a menu option to enable or disable the feature. This means that players who don't like the classic zooms and prefer how dsda-doom is currently set up can continue to use dsda-doom without this feature (I've set the option to "no" by default)

There are two things I'd like to mention that I haven't been able to solve / figure out.
1. I'd like to scale the screensize border graphics similar to the grnrock flat
2. This is a carryover from PrBoom, but it's weird that the automap is forced in the screensize, instead of Vanilla where the automap is the same fullscreen size regardless of screen zoom size. I think it makes more sense to have the automap fullscreen at all times, as it's useless to use under smaller screen sizes.

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -617,15 +617,22 @@ void AM_SetPosition(void)
 {
   if (automap_active)
   {
-    f_x = f_y = 0;
-    f_w = SCREENWIDTH;
-
     if (automap_overlay > 0)
     {
-      f_h = viewheight;
+      f_x = 0;
+      f_y = 0;
+      f_w = SCREENWIDTH;
+      f_h = SCREENHEIGHT;
+
+      f_x = viewwindowx + f_x * viewwidth / SCREENWIDTH;
+      f_y = viewwindowy + f_y * viewheight / SCREENHEIGHT;
+      f_w = f_w * viewwidth / SCREENWIDTH;
+      f_h = f_h * viewheight / SCREENHEIGHT;
     }
     else
     {
+      f_x = f_y = 0;
+      f_w = SCREENWIDTH;
       f_h = SCREENHEIGHT - ST_SCALED_HEIGHT;
     }
   }
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -340,6 +340,7 @@ static void D_DrawPause(void)
 }
 
 static dboolean must_fill_back_screen;
+extern dboolean ScreenSize_Interpolate;
 
 void D_MustFillBackScreen(void)
 {
@@ -486,6 +487,10 @@ void D_Display (fixed_t frac)
     DSDA_REMOVE_CONTEXT(sf_status_bar);
 
     BorderNeedRefresh = false;
+
+    // Interpolate weapon only when not changing screensize
+    ScreenSize_Interpolate = true;
+
     if (V_IsSoftwareMode())
       R_DrawViewBorder();
 
--- a/prboom2/src/dsda/configuration.c
+++ b/prboom2/src/dsda/configuration.c
@@ -1144,7 +1144,7 @@ dsda_config_t dsda_config[dsda_config_count] = {
   },
   [dsda_config_screenblocks] = {
     "screenblocks", dsda_config_screenblocks,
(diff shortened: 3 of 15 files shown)


fabiangreffrath:
I think it's redundant to add a toggle for a feature that requies explicit key presses to take action anyway. If players are afraid of accidentally changing the screen size, they can just unbind the keys, or not?

andrikpowell:
> I think it's redundant to add a toggle for a feature that requies explicit key presses to take action anyway. If players are afraid of accidentally changing the screen size, they can just unbind the keys, or not?

I guess my logic was that the feature was removed for a reason, and perhaps the reason was that Kraflab didn't like the smaller screensizes? So I guess when I implemented the config option it was to allow players to have the screensize functionality as it currently works.

But if we think we don't need the option, I could scrap the option code pretty easily.

_Also ignore the force-push, I just cleaned up some of the code and removed some unneeded stuff (ex: the "paused" graphic had a "viewwindowy + 5" offset, but I tested it and it works the same way with or without, so I scrapped that edit)_

Pedro-Beirao:
I agree that we dont need an option for this. Just the keybinds.

andrikpowell:
Alright y'all's wishes are my command :P

I'd rather not have the option if we don't need it, so I removed it from the branch. I can easily add it back if need be.

rfomin:
> I guess my logic was that the feature was removed for a reason, and perhaps the reason was that Kraflab didn't like the smaller screensizes?

The reason was the same as for removing ENDOOM or PC Speaker emulation - remove useless code. Less code - less bugs and easier maintenance.

Have there been any complaints about the lack of this feature? Maybe we shouldn't restore it.

andrikpowell:
> The reason was the same as for removing ENDOOM or PC Speaker emulation - remove useless code. Less code - less bugs and easier maintenance.

I'm pretty sure the main reason the feature was cut was cuz at the time Kraflab completely redesigned the message / hud display... and alot of that logic was tied with the screensizes, for whatever reason.

So when I reimplemented the screensizes back, I didn't touch any of the message code that was removed... and it works fine with how the message code works currently.

> Have there been any complaints about the lack of this feature? Maybe we shouldn't restore it.

I'm definitely not a fan of just removing features that were present in the original exe. To be quite honest, I've thought about restoring PC Speaker emulation as well, but I'm sure that's an even less sought after feature. I realise I'm probably a bit weird about things like this, as one of the things that still bothers me is that there's no "end game" option in the main options menu.

But I think a feature like this is worth restoring as a majority of source ports already do, including even GZDoom. The only source port I can think of that doesn't currently support this is the KEX engine, which I personally disagree with, but that's a discussion for another time. I'm unsure if Helion does or not...

rfomin:
> But I think a feature like this is worth restoring as a majority of source ports already do, including even GZDoom. The only source port I can think of that doesn't currently support this is the KEX engine, which I personally disagree with, but that's a discussion for another time. I'm unsure if Helion does or not...

The KEX port supports the "screen size" feature, although it doesn't render the border. Helion doesn't support it at all. Personally, I do not think this feature is worth restoring. Not many users even know about it these days.

Another small issue - background flat tiling changes when the zoom level feature is activated.
Before:
image
After:
image

andrikpowell:
> Another small issue - background flat tiling changes when the zoom level feature is activated.
> Before:
> image
> After:
> image

This should be fixed now.

Pedro-Beirao:
> Personally, I do not think this feature is worth restoring. Not many users even know about it these days.

Eh I agree. If it hadnt been removed, it wouldnt hurt to keep it, but at the same time, no point in bringing it back.

If for some reason there is demand for this in the future, we can merge this, but for now I think its fine to keep it a nyan only thing. Thank you for the PR anyway!

Pedro-Beirao:
And important to note that this was used to effectively reduce the resolution so that slower machines could play the game (3DO port is a good example), we dont need it now that we can choose the resolution.

So even though Im also not a fan of removing vanilla features, I see this as an improvement/replacement of the feature, not simply removing it.

Pedro-Beirao:
Closing as explained above

PR #617, opened 2025-02-27
Invert Statusbar "Use Red Numbers" logic

andrikpowell:
I think this is self-explanatory.


Pedro-Beirao:
I dont like that this will force everyone that uses colored numbers to toggle the setting again. But I do agree its a much better name for the option.

andrikpowell:
> I dont like that this will force everyone that uses colored numbers to toggle the setting again. But I do agree its a much better name for the option.

I'll mention this was partly inspired by Woof doing this change a while back.

I even used the same `sts_colored_numbers` config name:
https://github.com/fabiangreffrath/woof/commit/d3db71a12e14eccb255ffba9d7e3bfffbd13648c

Pedro-Beirao (approved):

PR #618, opened 2025-02-27
Automap - "Secret Exit" and "Tag Finder" Line Colours

andrikpowell:
This adds two additional line colours to the automap:
- Alt Secret Exit Line
- Tag Finder Line

For the secret exit line, if blank, it will default to using the normal "exit line" color. That's why I added "alt" to the name.

For the "tag finder" line, I'm unsure if we want to include that in the OPTIONS lump or not.

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -1883,7 +1883,7 @@ static void AM_DrawConnections(void)
       AM_SetMPointFloatValue(&l.b);
     }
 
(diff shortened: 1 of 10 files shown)


Pedro-Beirao:
~~No need for the "alt" name imo. It doesnt really describe it as "whenever its unset use the other one".~~
Changed my mind, I like it.

Pedro-Beirao (approved):

PR #620, opened 2025-02-27
Cheat Additions / Tweaks

andrikpowell:
This PR includes multiple parts:
- Added a "SKILL" cheat to change the skill in-game
- Added TNTCOMP, SKILL, and IDCLEV "status" cheats
 (from Woof)
- Disabled TNTCOMP for Raven games

### DIFF
--- a/docs/guides/console.md
+++ b/docs/guides/console.md
@@ -211,6 +211,7 @@ See the [build mode guide](./build_mode.md) for more info.
 - `idbeholdr`
 - `idbeholda`
 - `idbeholdl`
+- `skill`
 - `tntcomp`
 - `tntem`
 - `tnthom`
--- a/prboom2/src/dsda/console.c
+++ b/prboom2/src/dsda/console.c
@@ -2464,6 +2464,7 @@ static console_command_entry_t console_commands[] = {
   { "idbeholda", console_BasicCheat, CF_DEMO },
   { "idbeholdl", console_BasicCheat, CF_DEMO },
 
+  { "skill", console_BasicCheat, CF_DEMO },
   { "tntcomp", console_BasicCheat, CF_DEMO },
   { "tntem", console_BasicCheat, CF_DEMO },
   { "tnthom", console_BasicCheat, CF_DEMO },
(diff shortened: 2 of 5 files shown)


Pedro-Beirao (changes requested):
Need to add the new cheat to the console as well in `console_commands[]` of `console.c`.

And can we name it `TNTSKILL`? Even though its not a real Boom cheat, it fits right next to `TNTCOMP`.
Alternatively we could start new `DSDA` cheats (`DSDASKILL`), but I would like to avoid that, unless we introduce many new cheats.

andrikpowell:
> And can we name it `TNTSKILL`? Even though its not a real Boom cheat, it fits right next to `TNTCOMP`.

It's named SKILL because that's what Woof uses. Imo it's easier to type SKILL over TNTSKILL.

andrikpowell:
> Need to add the new cheat to the console as well in `console_commands[]` of `console.c`.
> 
> And can we name it `TNTSKILL`? Even though its not a real Boom cheat, it fits right next to `TNTCOMP`. Alternatively we could start new `DSDA` cheats (`DSDASKILL`), but I would like to avoid that, unless we introduce many new cheats.

Just coming back to this. I added `SKILL` as a console cheat.

I know what I said above, but I did think of another reason I'm not a fan of doing `TNTSKILL`, besides the length.

I don't think we should add `TNT`- prefix cheats, since those are typically associated with classic Boom cheats. I know that `TNTCOMP` is sorta similar in functionality, but that was an actual Boom cheat.

Of course, I also wanna keep some cheats like this consistent with Woof, and currently the cheat `SKILL` is used in that port.

I think the downside of having `DSDA` cheats would be that `DSDA` is a little long and hard to type as prefix. (feel free to disagree with me here). But also the minor problem with doing cheats this way, is that other ports that wanted to add those cheats would be stuck with the `DSDA` prefix as well.

Pedro-Beirao:
Showing the `next` map in `idclev0` may spoil the player that its the last map or that there is a secret map accessable from the current.
- current map30, next map01
- current map15, next map31

I think its best to just present the `current` map

fabiangreffrath:
Well, at this point the player is cheating anyway...

Pedro-Beirao (approved):

Pedro-Beirao:
Yes its cheating, but maybe `idclev` is being used to play again a previous level or something. I wouldnt want to be spoiled by it.

PR #622, opened 2025-03-07
Interpolate bobbing for animated weapons (Chainsaw)

andrikpowell:
While most weapon's bob looks nice and smooth at high framerates, animated weapons with multiple WeaponReady frames look jittery and awful... This PR fixes that.

I've coded this in a particular way where it applies to any animated idle weapon... as opposed to only the Chainsaw. Let me know if there's a better approach.

I tested this with the [Corruption 2 Demo](doomworld.com/forum/topic/150350), which has animated Chainsaw and BFG weapons.

### DIFF
--- a/prboom2/src/r_things.c
+++ b/prboom2/src/r_things.c
@@ -1105,7 +1105,7 @@ static void R_DrawPSprite (pspdef_t *psp)
           R_ApplyWeaponBob(NULL, false, &psp_sy, weapon_attack_alignment == CENTERWEAPON_BOB);
       }
     }
-    else if (psp->state->action == A_WeaponReady && psp->state->tics > 1)
+    else if (psp->state->action == A_WeaponReady && psp->state->tics > 1 && movement_smooth)
     {
       // Interpolate bobbing for animated weapons (Chainsaw)
       R_ApplyWeaponBob(&psp_sx, true, &psp_sy, true);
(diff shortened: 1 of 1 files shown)


Pedro-Beirao:
Right, seems like a good solution. But can we keep this behind a config option? I like my chainsaw jittery and awful.
`Display -> Options -> Smooth Chainsaw` maybe

andrikpowell:
> Right, seems like a good solution. But can we keep this behind a config option? I like my chainsaw jittery and awful. `Display -> Options -> Smooth Chainsaw` maybe

Hmm tbh this is basically a bug with weapon interpolation... Meaning that it is not the intended behaviour for multi frame idle weapons to not be interpolated with all other weapons being interpolated.

I do think that it is weird that weapon interpolation is never disabled when "uncapped framerate" is disabled, especially when other interpolation tweaks are disabled when that option is turned off.

I see this as an error with the implementation of weapon interpolation, so it feels weird to have an option to re-enable the "bug". I think we could have a config-only option, but I'm not crazy about including it in the menu, and think that the fix should be enabled by default.

I do think it's weird to retain the bug behaviour when Woof has the smooth behaviour without any option to disable it.

Pedro-Beirao:
Yea, on Woof its smooth when using Uncapped FPS.
Sure Im ok with doing the same. No point to have a config option, unless we get multiple people asking for it.

PR #623, opened 2025-03-07
Add colored statusbar background option

andrikpowell:
This is my attempt to add this feature: https://github.com/fabiangreffrath/woof/pull/655

I've set this as a draft because it still needs work.

We can't use the method that Woof uses because it's software only, and so I've tried a slightly different method that also works for OpenGL.

it's based off these 2 functions: [dsda_CalculateFontBounds](https://github.com/kraflab/dsda-doom/blob/51d4226b7e7d4db8d51512a906f2e5de124dc6d2/prboom2/src/dsda/cr_table.c#L76-L120) and [dsda_PaletteEntryLightness](https://github.com/kraflab/dsda-doom/blob/51d4226b7e7d4db8d51512a906f2e5de124dc6d2/prboom2/src/dsda/palette.c#L102-L141) which are used to determine brightness (and colours) for fonts in DSDA (hence this works in both software and opengl).

Main issue with this at the moment is that it needs a way to average the values... which I guess I can't seem to figure out.

### DIFF
--- a/prboom2/src/dsda/configuration.c
+++ b/prboom2/src/dsda/configuration.c
@@ -27,6 +27,7 @@
 #include "r_main.h"
 #include "r_segs.h"
 #include "s_sound.h"
+#include "st_stuff.h"
 #include "smooth.h"
 #include "v_video.h"
 #include "z_zone.h"
@@ -311,6 +312,10 @@ dsda_config_t dsda_config[dsda_config_count] = {
   [dsda_config_sts_traditional_keys] = {
     "sts_traditional_keys", dsda_config_sts_traditional_keys,
     CONF_BOOL(0), &sts_traditional_keys
+  },
+  [dsda_config_sts_solid_bg_color] = {
+    "sts_solid_bg_color", dsda_config_sts_solid_bg_color,
+    CONF_BOOL(0), NULL, NOT_STRICT, ST_SetResolution
   },
   [dsda_config_strict_mode] = {
     "dsda_strict_mode", dsda_config_strict_mode,
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -3330,6 +3330,7 @@ setup_menu_t display_statbar_settings[] =  // Demos Settings screen
 {
   { "Hide Status Bar Horns", S_YESNO, m_conf, DM_X, dsda_config_hide_horns },
   { "Single Key Display", S_YESNO, m_conf, DM_X, dsda_config_sts_traditional_keys },
+  { "Solid Color Background", S_YESNO, m_conf, DM_X, dsda_config_sts_solid_bg_color },
   EMPTY_LINE,
   TITLE("Coloring", DM_X),
   { "Gray %",S_YESNO, m_conf, DM_X, dsda_config_sts_pct_always_gray },
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -493,6 +493,73 @@ void R_InitBuffer(int width, int height)
 }
 
 //
+// R_FillBackColor
+// Fills the statusbar widescreen area
+// with a color
+//
+
+void R_FillBackColor (void)
+{
+  extern patchnum_t stbarbg;
+  byte col, col_top;
+  int i, j, pixel_cnt;
+  int r, g, b;
+  int stbar_top = SCREENHEIGHT - ST_SCALED_HEIGHT;
+  int ST_SCALED_BORDER = brdr_b.height * patches_scaley/2;
+  const unsigned char *playpal = V_GetPlaypal();
+
+  const byte* lump;
+  const byte* p;
+  short width;
+  byte length;
+  byte entry;
+
+  pixel_cnt = 0;
+  r = g = b = 0;
+
+  lump = W_LumpByNum(stbarbg.lumpnum);
+  width = *((const int16_t *) lump);
+  width = LittleShort(width);
+
+  for (i = 0; i < width; ++i) {
+    // Skip irrelevant data in the doom patch header
+    int32_t offset;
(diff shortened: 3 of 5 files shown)


andrikpowell:
Ok, so I finally got this working, so it's now ready for review :)

prboom2/src/r_draw.c
```diff
@@ -492,6 +492,73 @@ void R_InitBuffer(int width, int height)
...
+
+  for (i = 0; i < width; ++i) {
+    // Skip irrelevant data in the doom patch header
+    int32_t offset;
+    p = lump + 8 + 4 * i;
+    offset = *((const int32_t *) p);
+    p = lump + LittleLong(offset);
+
+    while (*p != 0xff) {
+      p++;
```
fabiangreffrath, line 532: You'll need to comment that part of the code as well, please.

prboom2/src/dsda/configuration.c
```diff
@@ -312,6 +313,10 @@ dsda_config_t dsda_config[dsda_config_count] = {
     "sts_traditional_keys", dsda_config_sts_traditional_keys,
     CONF_BOOL(0), &sts_traditional_keys
   },
+  [dsda_config_stbar_bg_color] = {
+    "config_stbar_bg_color", dsda_config_stbar_bg_color,
+    CONF_BOOL(0), NULL, NOT_STRICT, ST_SetResolution
+  },
```
Pedro-Beirao, line 319: What about `sts_solid_bg_color`? So that it fits with the similar options.

prboom2/src/m_menu.c
```diff
@@ -3330,6 +3330,7 @@ setup_menu_t display_statbar_settings[] =  // Demos Settings screen
 {
   { "Hide Status Bar Horns", S_YESNO, m_conf, DM_X, dsda_config_hide_horns },
   { "Single Key Display", S_YESNO, m_conf, DM_X, dsda_config_sts_traditional_keys },
+  { "Colored Background", S_YESNO, m_conf, DM_X, dsda_config_stbar_bg_color },
```
Pedro-Beirao: `Solid Color Background` is more descriptive

prboom2/src/r_draw.c
```diff
@@ -541,6 +609,9 @@ void R_FillBackScreen (void)
         V_FillPatch(brdr_b.lumpnum, 1, 0, stbar_top, ST_SCALED_OFFSETX, brdr_b.height, VPT_NONE);
         V_FillPatch(brdr_b.lumpnum, 1, SCREENWIDTH - ST_SCALED_OFFSETX, stbar_top, ST_SCALED_OFFSETX, brdr_b.height, VPT_NONE);
       }
+
+      if (stbar_color)
+        R_FillBackColor();
```
Pedro-Beirao: This can be moved to right before `if (V_IsOpenGLMode())` with an added `return;`. 
And maybe name the variable `stbar_solid_bg`?

prboom2/src/r_draw.c
Pedro-Beirao: So this is looping through the posts instead of the height. Yeah would help to have comments here.

Other than that, its easy to understand and everything looks correct 😄

andrikpowell: Unfortunately, I don't personally know what's happening here.

This was based on previous DSDA-Doom code, and Kraflab didn't explain what he was doing there:
 [dsda_CalculateFontBounds](https://github.com/kraflab/dsda-doom/blob/51d4226b7e7d4db8d51512a906f2e5de124dc6d2/prboom2/src/dsda/cr_table.c#L76-L120)

According to my programming friend, this section has something to do with bitmasks? So I think the `0xff` dictates the end of buffer.

Unfortunately that's above my programming knowledge, so I can't really write a comment on something I don't fully understand myself.

prboom2/src/dsda/configuration.c
andrikpowell: I agree, updated!

prboom2/src/m_menu.c
andrikpowell: Done👍

prboom2/src/r_draw.c
andrikpowell: Yeah I tested moving this up top.
I assume you wanted to do this to avoid having to print GRNROCK at all.
Seems to work fine. Also tweaked the variable name.

I had to add a `V_EndUIDraw()` before the `return` for OpenGL ofc.

prboom2/src/r_draw.c
rfomin: `0xff` is the end of "post" in Doom graphics patch format. See https://doomwiki.org/wiki/Picture_format#Posts

fabiangreffrath: I'm asking for clarification of the `p++; length = p++; p++;` sequence.

Pedro-Beirao: Just put a comment above the `while` loop like
```
// Loop vertically through the pixels using the column of posts, which ends with 0xff
```

rfomin: > I'm asking for clarification of the `p++; length = p++; p++;` sequence.

It skips the padding bytes.

In general I don't think we should comment much here, it's Doom's patch format, we have a lot of code like this without comments. To simplify it, we can draw the patch to the buffer using `V_DrawPatch` like in Woof, although it might be more complicated in DSDA-Doom due to OpenGL and different video buffer management.

andrikpowell: Originally I attempted using a method closer to what Woof does, with drawing an actual patch... But that would end up crashing OpenGL.

That's why I went with this method, as it's what's used to grab the lightness of the font patches, and obviously that works for both software and OpenGL.

Pedro-Beirao (approved):

PR #627, opened 2025-03-14
Disable strict options in the menu during strict mode

Pedro-Beirao:
Ive been wanting this for so long, sometimes its hard to figure out what is allowed in demos and whats not, this will help.
Just setting this as a draft for now until I finalize marking options as strict. Otherwise it should be all good.

Screenshot 2025-03-14 at 15 40 09

TODO for another time
Allow options to be disabled based on the state of some other options


Pedro-Beirao:
New commit, the strict mode option is now only temporary to prevent accidentally recording tas runs.
This was suggested on the discord

PR #629, opened 2025-03-17
Reset best kills, items, secrets on new highest skill completion

thecnoNSMB:
I noticed while reading the source code that these are not reset when they probably should be. I reset them to 0 since that appears to be the value they initialize with. Creating a PR as requested on the Discord. (I'm not a C developer, please have patience, thanks!)


thecnoNSMB:
~~Converting this to a draft since I would like the changes to be reviewed by an actual C developer. (I'm not sure if that's what drafts are for, feel free to change it back if you see fit.)~~ (yeah i'm pretty sure that's not what drafts are for, nevermind!)

Pedro-Beirao:
Thank you!

For clarification, this fixes a case where you first complete a level in a lower skill with many kills/secrets/items, but when you complete it again in an higher skill but less kill/secrets/items it would preserve the higher stats from the lower skill

PR #631, opened 2025-03-28
Dont allow saving the game on strict mode

Pedro-Beirao:
Requested on discord. Saving the game can be used to open the save in a new dsda instance to check where remaining monsters are.

### DIFF
--- a/prboom2/src/dsda/save.c
+++ b/prboom2/src/dsda/save.c
@@ -33,6 +33,7 @@
 #include "dsda/mapinfo.h"
 #include "dsda/music.h"
 #include "dsda/options.h"
+#include "dsda/settings.h"
 
 #include "save.h"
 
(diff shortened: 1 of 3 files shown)


fabiangreffrath:
He, isn't this perverting strict mode somehow? My understanding of strict mode is that it should make it impossible to use source port advantages that go beyond the possibilities that Vanilla already provided.

Pedro-Beirao:
Strict mode blocks all tas features. Yes in vanilla you can save and open it in another instance to cheat, but I think we shouldnt allow it here anymore

Pedro-Beirao:
Strict mode also prevents the use of cheats (IDDT for example), while vanilla doesnt.

fabiangreffrath:
Yes right, but preventing to save is really mean. Maybe restrict it to demo recording?

Pedro-Beirao:
`dsda_StrictMode()` checks for both `demorecording` and if strict mode is on.
Am I missing something?

fabiangreffrath:
Alright, thanks!

Issue #636, opened 2025-04-02
Tall sky textures are positioned differently in Software vs Hardware modes


bangstk:
I assume this was caused by the Heretic fix in [#374](https://github.com/kraflab/dsda-doom/pull/374#issuecomment-1550588868)

In Woof, Eternity, and DSDA's own Software renderer, sky textures with 200px height, 240px height, and 256px height are all drawn with the top of the sky texture starting at the top of the screen (when status bar is off)

![Image](https://github.com/user-attachments/assets/713ab48a-7de8-448a-89c0-3a07ef56a13b)

However, in DSDA's Hardware renderer (and using r_skymode 2 - the 2D sky renderer) the vertical positioning diverges with 200px, 240px, and 256px height skies; starting with the bottom of the texture at the middle of the screen at 200px and shifting lower and lower the taller the sky is.

![Image](https://github.com/user-attachments/assets/d13fb9ea-f716-4628-be2e-1297552cffc9)

I assume this is a hack intended for Heretic sky behavior; maybe it needs a check to only do it in Heretic?

Attached is a test wad containing 128px, 200px, 240px, and 256px skies for comparison with Software vs hardware and with other software rendering ports.

MAP01 uses MBF sky transfers to set the sky differently in all 4 quadrants of the map;
MAP02-MAP05 uses UMAPINFO to directly set the sky texture to a 128px (MAP02), 200px (MAP03), 240px (MAP04), and 256px (MAP05) and uses no MBF sky transfers. MAP03-MAP05 are the subject of this issue.
[skyalign.zip](https://github.com/user-attachments/files/19561964/skyalign.zip)

All tests are with mouselook off & default FOV.

Note that this affects skies NOT being set by MBF sky transfer line special.
The positioning when using the MBF sky transfer appears consistent with other ports (Woof, Eternity) whether using software or hardware mode so I don't think anything needs to be touched there.

r_skymode 3 has inconsistencies as well, but that probably is for another issue.

bangstk:
DSDA's hardware 2d sky positioning logic appears to be based on this comment block in software renderer's [r_sky.c](https://github.com/kraflab/dsda-doom/blob/3fadd3c52b1d16a4c9e5f3fff7575a6cc4e6a2ba/prboom2/src/r_sky.c#L78):

```    // There are various combinations for sky rendering depending on how tall the sky is:
    //        h <  128: Unstretched and tiled, centered on horizon
[...]

PR #637, opened 2025-04-03
Fix GL mode's vertical position of Doom 2D tall skies to match Software mode

bangstk:
Addresses #636 - demonstration WAD for comparison is in there

For DSDA v0.26 the GL 2D sky type's code was altered to fix positioning of tall sky textures in Heretic/Hexen, but it also affects Doom.
This caused the vertical alignment of tall sky textures (h > 128) to no longer match up with the Software renderer's sky position in Doom.

Fixed by making the new cases only affect Raven games & added new math for Doom mode to match SW.
Top of sky texture aligns with top of screen with mouselook off & default FOV.
Also adjusted a comment in r_sky (software offset) to reflect that the logic only applies to Raven games.

Heretic/hexen is not affected.
MBF sky transfers is not affected.
The GL 3D dome sky type is not affected (however it also does misalign with software sky in tall sky situations, but perhaps to be left for another issue)

Now, the sky alignment of tall skies matches Software mode, as well as other MBF21 ports such as Woof and Eternity.

Admittedly, this alignment makes tall skies not help at all with mouselook, but I think that can be considered for a future change, or left to SKYDEFS, and until then sw/gl sky can be more consistent with each other & other ports.

### DIFF
--- a/prboom2/src/r_sky.c
+++ b/prboom2/src/r_sky.c
@@ -75,7 +75,7 @@ void R_InitSkyMap(void)
     if (!textureheight)
       return;
 
-    // There are various combinations for sky rendering depending on how tall the sky is:
+    // There are various combinations for Raven sky rendering depending on how tall the sky is:
     //        h <  128: Unstretched and tiled, centered on horizon
     // 128 <= h <  200: Can possibly be stretched. When unstretched, the baseline is
     //                  28 rows below the horizon so that the top of the texture
--- a/prboom2/src/r_sky.c
+++ b/prboom2/src/r_sky.c
@@ -75,7 +75,7 @@ void R_InitSkyMap(void)
     if (!textureheight)
       return;
 
-    // There are various combinations for Raven sky rendering depending on how tall the sky is:
+    // There are various combinations for sky rendering depending on how tall the sky is:
     //        h <  128: Unstretched and tiled, centered on horizon
     // 128 <= h <  200: Can possibly be stretched. When unstretched, the baseline is
     //                  28 rows below the horizon so that the top of the texture
--- a/prboom2/src/gl_sky.c
+++ b/prboom2/src/gl_sky.c
@@ -186,7 +186,9 @@ void gld_AddSkyTexture(GLWall *wall, int sky1, int sky2, int skytype)
       wall->skyyaw = skyXShift;
       wall->skypitch = skyYShift;
       // Choose offset based on logic from r_sky.c
-      if (h >= 128 && h < 200)
+      if (!raven)
+        wall->skyoffset = (100 - h) / (float)h;
+      else if (h >= 128 && h < 200)
         wall->skyoffset = -28.0f / 128.0f;
       else if (h > 200)
         wall->skyoffset = (200 - h) / 128.0f;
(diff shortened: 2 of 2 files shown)


prboom2/src/r_sky.c
```diff
@@ -75,7 +75,7 @@ void R_InitSkyMap(void)
     if (!textureheight)
       return;
 
-    // There are various combinations for sky rendering depending on how tall the sky is:
+    // There are various combinations for Raven sky rendering depending on how tall the sky is:
```
Pedro-Beirao: Im having a hard time figuring this out. This comment looks to be about Raven skies, but there's a `if (raven || !dsda_MouseLook())` condition that makes this code not run under Raven...

So while your fix seems to make sense (looking at `skytexturemid = (raven ? 200 : 100) * FRACUNIT;`, this whole sky code feels wrong.

bangstk: I think it is because of "skytexturemid" being set according to "skystretch" in the second half of that block, totally overriding "skytexturemid" in the case of height >= 200.

In short, the comment block does not seem to have ever accurately described the sky positioning behavior under software mode.

It does not seem intended, but I have checked back through DSDA/PrBoomPlus history and it has behaved this way as far back as mouselook has been a part of PrBoomPlus (and even before then), and changing the software positioning behavior now would alter any wads in the last 20 years which have used tall skies with non MBF sky transfers.

Otherwise I might have to check out 129 thru 199 height case which the test WAD does not cover, that might be the only one that actually shifts the sky texture upwards in software mode.

Pedro-Beirao: What I mean is that in `r_sky.c`, the code that handles `>= 128`, `<200`, `>200`, etc, is only called when **NOT** playing Raven and with Mouse Look On.
While on `gl_sky.c`, with your fix, the code that handles `>= 128`, `<200`, `>200`, etc, is only called when playing Raven.

I think this is more correct, or atleast points in the correct direction
```c
if (raven || !dsda_MouseLook())
  wall->skyoffset = ...
else if (h >= 128 && h < 200)
  wall->skyoffset = -28.0f / 128.0f;
else if (h > 200)
  wall->skyoffset = (200 - h) / 128.0f;
else
  wall->skyoffset = 0.0f;
```

But this whole sky rendering is confusing, Ive only tried to look at the conditions themselves and havent tested this.
I want to better look into sky rendering in the future and properly understand and fix all this. For now lets just fix this condition.

bangstk: That's doesn't work out because the r_sky.c "skytexturemid" and the gl_sky.c "skyoffset" are different calculations:

"skytexturemid" selects which row of the texture is placed at the center of the screen.
"skyoffset" selects between 0.0 to 1.0 (or -1.0) what pecentage of the texture's height should be placed at the center of the screen.

Based on this, I think I can rewrite this block to calculate "skyoffset" directly based on "skytexturemid" as the r_sky init code is ran even in GL mode.

bangstk (commented):
@Pedro-Beirao - what do you think now?

I was able to totally eliminate that 2d vertical sky position logic in gl_sky.c by making it use the texture mid row variable already calculated in r_sky,c - this way the logic is literally used for both SW & HW 2D rather than replicating it separately in gl_sky.c. And so the behavior for both can be changed at once by just changing the logic in r_sky.c 

Though the unmodified r_sky.c logic is still somewhat strange code that doesn't match up with that comment block this should make it easier to modify both SW + HW offsets in the future.

The result matches up for all of the sky height cases (including 129 < h < 199), and in Heretic/hexen still by my testing.

Pedro-Beirao (approved):
Looks better now. Thanks!

Very happy to finally have correct OpenGL tall skies

PR #640, opened 2025-04-05
Enable the Boom ANIMATED feature for the default sky texture

### LINKED ISSUE #639: Animated skies won't animate when defined via UMAPINFO
If UMAPINFO is used to define an animated sky texture, it won't animate -- seems to be a UMAPINFO-specific thing; they work if using a sky transfer line.

Here's a quick test map -- requires launching with [muumipack.wad](https://www.doomworld.com/forum/topic/130774-muumipack-v2-muumis-resource-of-animated-textures-big-update-28-animated-skies/) since that's where all the cool animated skies live.

[muumipack_umapinfotest.zip](https://github.com/user-attachments/files/19611280/muumipack_umapinfotest.zip)

rfomin:
Fix #639

### DIFF
--- a/prboom2/src/gl_sky.c
+++ b/prboom2/src/gl_sky.c
@@ -139,7 +139,7 @@ void gld_AddSkyTexture(GLWall *wall, int sky1, int sky2, int skytype)
 {
   side_t *s = NULL;
   line_t *l = NULL;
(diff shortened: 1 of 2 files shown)


Pedro-Beirao (approved):
Thanks!

PR #642, opened 2025-04-07
Fix tiled flat (grnrock) offsets on statusbar

andrikpowell:
So while the my "Zoom-Levels" PR #614 was discarded, this should probably be added in.

This issue was also pointed out in that PR, but also mentioned here on the Nyan Doom thread asking to be added to DSDA as well: https://www.doomworld.com/forum/post/2910001

Basically the way that the background flat (grnrock in doom) is tiling is actually not correct. The reason for this is because for some reason it's only being printed on the right and left sides of the statusbar...

While this make some sense, it also means that the starting point of the tiling ends up being incorrect. And so this PR fixes this issue.

### DIFF
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -520,19 +520,10 @@ void R_FillBackScreen (void)
     {
       int stbar_top = SCREENHEIGHT - ST_SCALED_HEIGHT;
 
-      if (V_IsOpenGLMode())
-        V_FillFlat(grnrock.lumpnum, 1, 0, stbar_top, SCREENWIDTH, ST_SCALED_HEIGHT, VPT_STRETCH);
+      if (V_IsOpenGLMode()) // OpenGL has no way to adjust y-offset independent from height
+        V_FillFlat(grnrock.lumpnum, 1, 0, 0, SCREENWIDTH, SCREENHEIGHT, VPT_STRETCH);
       else
-      {
-        V_FillFlat(grnrock.lumpnum, 1,
-          0, stbar_top, ST_SCALED_OFFSETX, ST_SCALED_HEIGHT, VPT_STRETCH);
-        V_FillFlat(grnrock.lumpnum, 1,
-          SCREENWIDTH - ST_SCALED_OFFSETX, stbar_top, ST_SCALED_OFFSETX, ST_SCALED_HEIGHT, VPT_STRETCH);
-
-        // For custom huds, need to put the backfill inside the bar area (in the copy buffer)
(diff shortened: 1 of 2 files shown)


prboom2/src/r_draw.c
```diff
@@ -520,19 +520,7 @@ void R_FillBackScreen (void)
...
-        V_FillFlat(grnrock.lumpnum, 1,
-          0, stbar_top, ST_SCALED_OFFSETX, ST_SCALED_HEIGHT, VPT_STRETCH);
-        V_FillFlat(grnrock.lumpnum, 1,
-          SCREENWIDTH - ST_SCALED_OFFSETX, stbar_top, ST_SCALED_OFFSETX, ST_SCALED_HEIGHT, VPT_STRETCH);
-
-        // For custom huds, need to put the backfill inside the bar area (in the copy buffer)
-        V_FillFlat(grnrock.lumpnum, 0,
-          ST_SCALED_OFFSETX, stbar_top, SCREENWIDTH - 2 * ST_SCALED_OFFSETX, ST_SCALED_HEIGHT, VPT_STRETCH);
-      }
+      V_FillFlat(grnrock.lumpnum, 1, 0, 0, SCREENWIDTH, SCREENHEIGHT, VPT_STRETCH);
```
Pedro-Beirao: Shouldnt it be `V_FillFlat(grnrock.lumpnum, 1, 0, stbar_top, SCREENWIDTH, ST_SCALED_HEIGHT, VPT_STRETCH);`?
So that it only draws it behind the stat bar and not the whole screen.

I understand that nyan needs to draw in the whole screen since you have the zoom levels, but we dont.

bangstk:
Any thoughts on scaling up the border of it in higher resolutions as well?

andrikpowell:
> Any thoughts on scaling up the border of it in higher resolutions as well?

This is surprisingly more difficult than it may seem (due to how DSDA deals with hi-res scaling).
Nyan Scaled Border

andrikpowell:
Updated this PR with current Nyan Doom code.

Now prints GRNROCK in the STBAR area.
- Software is now correct with the y offset always starting at 0 height regardless of area that's printed
- OpenGL is trickier. There is no way I could find to adjust the y offset independent from box rendering. So it prints fullscreen, although this doesn't really affect OpenGL performance,

prboom2/src/r_draw.c
andrikpowell: This code is now outdated, so I'm going to resolve this.

Pedro-Beirao (approved):

PR #644, opened 2025-04-08
A Few Minor fixes

andrikpowell:
Really small fixes that I think would be good to add in before release.

- Helpscreen fixes
  - Remove scrollbar
  - Add dark overlay
- Allow episode 4 if PWAD "E4M1" is found
  - I feel this is going to cause problems if we always remove the 4th episode pre complevel 3, since all other source ports don't remove the fourth episode... so if a mapper were to use UMAPINFO to add a new episode (to say replace the missing E4), DSDA would had 4 episodes while all other ports would have 5 including the duplicate 4th UMAPINFO episode
- Fixed a bunch of GCC warnings
  - I'm having some trouble fixing the strncat warning tied to the demo-footer, so I haven't fixed that yet.

### DIFF
--- a/prboom2/src/dsda/episode.c
+++ b/prboom2/src/dsda/episode.c
@@ -17,6 +17,7 @@
 
 #include "doomstat.h"
 #include "lprintf.h"
+#include "w_wad.h"
 #include "z_zone.h"
 
 #include "dsda/mapinfo.h"
@@ -55,7 +56,7 @@ void dsda_AddOriginalEpisodes(void) {
     dsda_AddEpisode("e2m1", NULL, "M_EPI2", 't', true);
     dsda_AddEpisode("e3m1", NULL, "M_EPI3", 'i', true);
 
-    if (gamemode == retail && compatibility_level >= ultdoom_compatibility)
+    if (gamemode == retail && (compatibility_level >= ultdoom_compatibility || W_PWADLumpNameExists("E4M1")))
       dsda_AddEpisode("e4m1", NULL, "M_EPI4", 't', true);
   }
 }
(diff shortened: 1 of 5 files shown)


fabiangreffrath:
>   * I'm having some trouble fixing the strncat warning tied to the demo-footer, so I haven't fixed that yet.

I believe you can just `strncat(..., ..., 3)`, because the third byte will be `'\0'` anyway?

fabiangreffrath:
And then you could just `strcat(...), right?

Pedro-Beirao:
The scrollbar is showing in the help screen because there are some keys out of the page.
Instead of making this exception, we should reorganise the keys (I have some changes there I want to make, so Ill do it).

I dont understand what you are saying about E4. How would other ports have 5 Eps (including a duplicate 4th)?

andrikpowell:
> I dont understand what you are saying about E4. How would other ports have 5 Eps (including a duplicate 4th)?

Ok, let me explain. UMAPINFO episode definitions are additive... Meaning they don't clear previous episodes unless "episode = clear" is called.

This can cause a problem because currently only DSDA-Doom is removing Episode 4 when running below complevel 3. This means that if a mapper used UMAPINFO to add an episode for a Doom 1 under complevel 2, they would only have the 3 main episodes with the new episode after.

However, use that same UMAPINFO in other ports like GZDoom, and they would get 4 main episodes with the new episode after.

See the issue? Now we have a discrepancy of listed episodes based on the port. So my solution is if a PWAD "E4M1" is detected, to allow the fourth episode to show, thus avoiding discrepancies between ports if "episode = clear" isn't used.

Another thing to consider is that some mappers like me prefer not to use UMAPINFO for Vanilla Doom 1 WADs, as it messes with the previous/next map key functionality... Instead of cycling through all of Doom 1 maps (i.e. pressing next on E1M8 goes to E2M1), UMAPINFO forces that logic to only that episode. This functionality is even worse if you have just 1 UMAPINFO entry to fix something, as then the previous/next map keys only cycle between that map only.

andrikpowell:
I've made some tweaks to this PR

- Due to us removing the Dynamic Help screen, I've removed the fade for that.
- I have kept the logic to remove the scrollbar from such screens though, as the scrollbar should never show there
- Added a GCC fix for the demofooter (@Pedro-Beirao please check and see if this is working correctly)

Pedro-Beirao (approved):

Pedro-Beirao (approved):

PR #646, opened 2025-04-10
Better Skill Definitions

andrikpowell:
So when adding the new bonus UV Plus skill for Nyan (and when working on a new skill definition functionality), I noticed some flaws with some of the logic pertaining to skills.

Much of the logic for stuff like the level table was specific to 4 being UV and 5 being NM skill. The problem with this is when a WAD uses MAPINFO to add skill definitions... as then NM skill may not be skill 5.

And so this PR tweaks some of these definitions to consider the last skill to be the "nightmare" skill, with the one under being the "best normal skill".

I also added Skill Cheat string support for Raven games, as well as custom MAPINFO skills... so it now uses the actual difficulty names of the skills when using the SKILL cheat.

Either way, this is worth implementing as whenever we get to the point of having a universal skill customisation lump, the skill logic will have already bene set up for that.

_(PS I have replaces SK5 with NM in the level table, however that may not be the best given that the final difficulty may not be nightmare - aka raven skills. Although we should rename SK5 to something else, as there's no guarantee the final skill will be skill 5)_

### DIFF
--- a/prboom2/src/dsda/skill_info.h
+++ b/prboom2/src/dsda/skill_info.h
@@ -52,6 +52,7 @@ typedef struct {
 } skill_info_t;
 
 extern skill_info_t skill_info;
+extern skill_info_t *skill_infos;
 
 extern int num_skills;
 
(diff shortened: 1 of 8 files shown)


Pedro-Beirao (approved):
Looks correct

PR #648, opened 2025-04-10
Heretic - Fix Inaccessible Menu Bug

andrikpowell:
Ok, so this is going to require a bit of background to explain what this issue is, and why it was an issue.

So there was some special logic added for the Heretic End Screens pertaining to E2's graphic. The reason for this is because Heretic E2's graphic is unique in that it uses a completely different palette. A sort of fix was implemented to avoid the player bringing up the menu during this screen... Since if they did, the menu would look really messed up as it would try and use the E2 end screen palette.

For E2, this worked fine... However it caused an issue with all the other end screens... Essentially what would happen is that if the player ended an episode other than E2, waited til the graphic to show after the text crawl, and then chose another episode, they would be unable to bring up the menu and press many of the keys when starting the new episode... Aka ESC and the F# keys would do nothing.

The culprit here was the line in `M_Responder()` function, as removing that fixed the issue. I had to rearrange a few things in the Heretic `f_finale.c` to get it to work the same... But now when starting a new episode after an episode end screen, it no longer blocks the player from accessing the menu and a bunch of other inputs.

The second commit is probably unneeded for DSDA-Doom, but it is a fix I had to add to Nyan Doom, as Nyan has an "Vanilla" option to allow for text screens and demos to play under the menu... And so to avoid weird palette issues, I had to force the menu to exit if it was up.

### DIFF
--- a/prboom2/src/heretic/f_finale.c
+++ b/prboom2/src/heretic/f_finale.c
@@ -83,24 +83,25 @@ void Heretic_F_StartFinale(void)
   S_ChangeMusic(heretic_mus_cptd, true);
 }
 
-dboolean F_BlockingInput(void)
+static dboolean F_BlockingInput(void)   // Avoid bringing up menu when loading Heretic's custom E2 palette
 {
   return finalestage == 1 && gameepisode == 2;
 }
 
 dboolean Heretic_F_Responder(event_t * event)
 {
-  if (event->type != ev_keydown)
-  {
-    return false;
-  }
-
   if (F_BlockingInput())
   {                           // we're showing the water pic, make any key kick to demo mode
     V_SetPlayPal(playpal_default);
     finalestage++;
     return true;
   }
+
+  if (event->type != ev_keydown)
+  {
+    return false;
+  }
+
   return false;
 }
 
--- a/prboom2/src/heretic/f_finale.c
+++ b/prboom2/src/heretic/f_finale.c
@@ -18,6 +18,7 @@
 #include "doomstat.h"
 #include "w_wad.h"
 #include "v_video.h"
+#include "m_menu.h"
 #include "s_sound.h"
 #include "sounds.h"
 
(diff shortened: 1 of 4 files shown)


fabiangreffrath:
Hm, I think a messed up menu is better than no menu at all. How does Vanilla Heretic itself handle that case?

andrikpowell:
> Hm, I think a messed up menu is better than no menu at all. How does Vanilla Heretic itself handle that case?

Well my point here was that Kraflab specifically avoided the menu coming up on that screen... That fact and logic hasn't changed... I'll go check how Vanilla Heretic deals with this to double check... Tbh it may work the same as it does here..

The main thing I fixed here was the inability to access the menu keys after the end screen of any other episode and entering another episode... which is a pretty big bug... Basically it made it impossible to open the menu or even save the game... Like you couldn't even quit the game without alt-f4.

andrikpowell:
> Hm, I think a messed up menu is better than no menu at all. How does Vanilla Heretic itself handle that case?

oh also to clarify, if the player hits the ESC key (or any key), or waits for the special E2 screen duration to continue, it draws the TITLE and resets the palette, allowing for the menu to be accessed again.

andrikpowell:
@fabiangreffrath oh actually very interesting...

Actually my way of kicking the player out of the menu is exactly what Vanilla Heretic does. Like I kept the menu up as it transitioned to the E2 screen, and it disabled the menu.

It actually can't draw both the menu and the E2 specific-palette screen.

When the player gets to that screen, if they hit escape, it sends them forward to the next "finalestate" which is TITLE, and resets the palette allowing menu access again.

So the way this PR works is exactly how Vanilla Heretic works.

prboom2/src/heretic/f_finale.c
```diff
@@ -83,24 +84,25 @@ void Heretic_F_StartFinale(void)
   S_ChangeMusic(heretic_mus_cptd, true);
 }
 
-dboolean F_BlockingInput(void)
+dboolean F_BlockingInput(void)   // Avoid bringing up menu when loading Heretic's custom E2 palette
```
fabiangreffrath: static?

fabiangreffrath:
Cool, thanks for checking!

prboom2/src/heretic/f_finale.c
andrikpowell: Fixed 👍

Pedro-Beirao (approved):

PR #650, opened 2025-04-12
properly scale yoffset value for out-of-bounds patches

### LINKED ISSUE #651: Some M_DOOM images are distorted in software rendering
https://www.doomworld.com/forum/topic/150766-dsda-doom-v0290-2025-04-08/?page=5&tab=comments#comment-2916282

This is an old bug existing since at least PrBoom+. Some M_DOOM logos are distorted in software rendering, like the columns are shifted up/down from where they should be. Some examples of this are Hellbound:

Hellbound, software rendering

and Skulltiverse:

Skulltiverse, software rendering

This only occurs with software rendering. In OpenGL (and in other ports), the M_DOOM renders fine:

Hellbound, OpenGL rendering

Skulltiverse, OpenGL rendering

I've also noticed that the amount of distortion seems to depend on the vertical resolution, with higher resolutions being more distorted than lower resolutions.

fabiangreffrath:
Fixes the Doom menu title for Hellbnd.wad

Fixes #651

### DIFF
--- a/prboom2/src/v_video.c
+++ b/prboom2/src/v_video.c
@@ -588,12 +588,12 @@ static void V_DrawMemPatch(int x, int y, int scrn, const rpatch_t *patch,
         }
 
         if (dcvars.yl < 0) {
-          yoffset = 0-dcvars.yl;
+          yoffset = (0-dcvars.yl) * 200/params->video->height;
           dcvars.yl = 0;
           dcvars.edgeslope &= ~RDRAW_EDGESLOPE_TOP_MASK;
         }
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
It's still completely broken at 320x200 resolution. We should consolidate stretched and unstretched patch drawing into one code path at some point, but that's for a separate PR.

Pedro-Beirao (approved):
I agree we should consolidate stretched and unstretched code, its so ugly and hard to work with currently.
Guess its ok to leave 320x200 broken for now. Thanks!

PR #654, opened 2025-04-16
Help screen improvements

Pedro-Beirao:
Allow customizing the key for showing the help screen (frees up F1).
Remove dynamic help screen.

~~Could be in another PR, but while Im reorganising the help screen Ive removed the Larger/Smaller View keys since they do the same job as the HUD key.~~

### DIFF
--- a/prboom2/src/dsda/input.h
+++ b/prboom2/src/dsda/input.h
@@ -129,7 +129,6 @@ typedef enum {
   dsda_input_menu_escape,
   dsda_input_menu_clear,
   dsda_input_help,
-  dsda_input_escape,
   dsda_input_cycle_profile,
   dsda_input_iddqd,
   dsda_input_idkfa,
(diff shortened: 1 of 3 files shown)


rfomin:
Alternatively, I suggest removing the "dynamic" help screen entirely. We did it in Woof and no one seems to notice.

Pedro-Beirao:
Yea I was also debating if we should just get rid of the dynamic help screen. Done in [1f6e954](https://github.com/kraflab/dsda-doom/pull/654/commits/1f6e954131f470d15fe767bbebe6094a0177d249)

A bit of a downgrade showing old static keybinds instead of what the player uses, but no one uses the help screen anyway

rfomin (approved):
LGTM

Pedro-Beirao:
Force push to no longer remove the Larger/Smaller View keys, since people are used to them

PR #655, opened 2025-04-17
Skip reading SMMU swirling flats

andrikpowell:
This fixes WADs with ANIMATED lumps that use SMMU swirling flats.

Fixes "Quoth the Raven":
https://www.doomworld.com/idgames/levels/heretic/Ports/puss28_qtr

Note that this doesn't add support for SMMU swirling flats. That would require more work, as well as finding some kind of solution for OpenGL.

All this PR does is skip over swirling flat definitions in ANIMATED, avoiding a hard crash.


Pedro-Beirao (approved):
Yea I think we should eventually add support to swirly flats since wads may expect it

PR #663, opened 2025-05-02
Port level brightness adjustment slider from Woof!

mincardona:
Woof! has a feature to raise the brightness of the lighting in a map (different from gamma adjustment) (https://github.com/fabiangreffrath/woof/pull/778). I find this useful for combating mild monitor glare, and so I've attempted to port it to DSDA-Doom.

The setting is not available in Woof! when strict mode is enabled, and I've kept that restriction here (though I personally don't see any harm in allowing it). The tweaks to the S_THERMO call are to scale the widget's central bar position according to the max setting value.

I know this port can have a skeptical view of certain kinds of "enhancements", so if this feature is not wanted, I will keep it in my own forked repo instead.

### DIFF
--- a/prboom2/src/dsda/features.h
+++ b/prboom2/src/dsda/features.h
@@ -25,7 +25,7 @@ typedef enum {
   uf_quickstartcache,
   uf_100k,
   uf_console,
-  uf_levelbrightness,
+  // 7
   // 8
   // 9
   // 10
@@ -87,7 +87,8 @@ typedef enum {
   // uf_fuzz = 64
   uf_vanillatrans = 65,
   uf_ghosttrans,
-  // 67
+  uf_levelbrightness,
+  // 68
 
   // 127
 } dsda_feature_flag_t;
--- a/prboom2/src/dsda/features.h
+++ b/prboom2/src/dsda/features.h
@@ -25,7 +25,7 @@ typedef enum {
   uf_quickstartcache,
   uf_100k,
   uf_console,
-  // 7
+  uf_levelbrightness,
   // 8
   // 9
   // 10
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -2040,7 +2040,7 @@ static void M_DrawSetting(const setup_menu_t* s, int y)
   }
 
   if (flags & S_THERMO) {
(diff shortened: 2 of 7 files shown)


Pedro-Beirao:
@rfomin do people like this in Woof?

> (though I personally don't see any harm in allowing it)

Obviously a TAS feature as its like playing with fullbright on.

And you need to track TAS features using `dsda_TrackFeature()` (setting it as strict is not enough)

rfomin:
> @rfomin do people like this in Woof?

It was originally in the Unity port (I think the KEX port has it too). We don't have much feedback on it. There are definitely no complaints.

mincardona:
> Obviously a TAS feature as its like playing with fullbright on.
> 
> And you need to track TAS features using `dsda_TrackFeature()` (setting it as strict is not enough)

That makes sense. I've added feature tracking for the new setting.

prboom2/src/dsda/features.h
```diff
@@ -25,7 +25,7 @@ typedef enum {
   uf_quickstartcache,
   uf_100k,
   uf_console,
-  // 7
+  uf_levelbrightness,
```
Pedro-Beirao: This needs to be at the end of the enum, really cannot be here.
Another PR has just been merged that adds features to the end as well, so this needs to be at 67

prboom2/src/m_menu.c
```diff
@@ -3159,6 +3159,7 @@ setup_menu_t gen_video_settings[] = {
   EMPTY_LINE,
   { "Fake Contrast", S_CHOICE, m_conf, G_X, dsda_config_fake_contrast_mode, 0, fake_contrast_list },
   { "OpenGL Light Fade", S_CHOICE, m_conf, G_X, dsda_config_gl_fade_mode, 0, gl_fade_mode_list },
+  { "Extra Lighting", S_THERMO, m_conf, G_X, dsda_config_extra_level_brightness },
```
Pedro-Beirao: Would make more sense to have both Gamma and Extra Lighting sliders, to not encourage people to use a TAS feature.
So lets not put this option in the menu yet, just keep it as a config.

Ill organize the page better to make space for both sliders sometime in the future

Pedro-Beirao (approved):
Looks good. Thank you!

PR #667, opened 2025-05-05
Dont allow autoload in strict demos

Pedro-Beirao:
Discussed in the speedrunning discord

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1906,6 +1906,15 @@ static void D_DoomMainSetup(void)
     I_SafeExit(0);
   }
 
+  // Recording strict demos will disable autoload
+  arg = dsda_Arg(dsda_arg_record);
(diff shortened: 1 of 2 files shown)


Pedro-Beirao:
As @andrikpowell pointed out, this would make using custom DSDAHUD in strict mode inconvenient. So we need to rethink this, and not simply disable autoload

rfomin:
We can add WADs from autoload to the demo footer.

Pedro-Beirao:
Imo the footer should only have the necessary wads for a certain run. Having different wads there (with potentially weird names) will only confuse people.

4shock also requested for a checksum of the wads loaded to be present in the footer (to make sure the wads are not modified), so we can only add to the footer known wads from idgames/dsdarchive

Pedro-Beirao:
Will need to rethink how this should be done.

PR #669, opened 2025-05-06
Fix scroll wheel not being allowed as bind for weapons

### LINKED ISSUE #645: Individual weapons bound to Scroll Wheel do not equip in game.
Scrolling Next and Previous weapons works but not individual weapons like binding Chaingun to Wheel Down. It will have no effect when scrolling down. This however works in some other ports but DSDA-Doom is my preferred. Affects current version v0.29.0 and many (all?) previous versions.

Just adding this here for visibility/tracking.  Thanks :)

Pedro-Beirao:
Fixes https://github.com/kraflab/dsda-doom/issues/645

Is this correct and safe?

### DIFF
--- a/prboom2/src/g_game.c
+++ b/prboom2/src/g_game.c
@@ -895,15 +895,15 @@ void G_BuildTiccmd(ticcmd_t* cmd)
       {
         // HERETIC_TODO: fix this
         newweapon =
-          dsda_InputActive(dsda_input_weapon1) ? wp_fist :    // killough 5/2/98: reformatted
-          dsda_InputActive(dsda_input_weapon2) ? wp_pistol :
-          dsda_InputActive(dsda_input_weapon3) ? wp_shotgun :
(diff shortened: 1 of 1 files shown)


rfomin:
When the SDL_MOUSEWHEEL event is detected, we send both `ev_keydown` and `ev_keyup` events on the same tick: https://github.com/kraflab/dsda-doom/blob/fda2c331513e5ac554269a61cddf854b9c7b2698/prboom2/src/SDL/i_video.c#L377-L386

In Woof we delay the "key up" event by one tick. Not sure which approach is better.

Pedro-Beirao:
If we delayed by a tic, we would still need a way to restrict the mouse wheel from being bound to movement keys.
While with this a approach it is a given (even though its pretty confusing).

So if there's no other downsides to simply using `dsda_InputTickActivated()`, I think we should go with it.

PR #674, opened 2025-05-13
Clickable progess bar

Pedro-Beirao:
Click progress bar to skip to point. Mouse gets visible and progress bar bigger when the mouse is moving.

In the future we should add clickable buttons like in the image (people in the speedrunning discord seemed to like this idea)

### DIFF
--- a/prboom2/src/SDL/i_sshot.c
+++ b/prboom2/src/SDL/i_sshot.c
@@ -59,15 +59,8 @@ int renderH;
 
 void I_UpdateRenderSize(void)
 {
-  if (V_IsOpenGLMode())
-  {
-    renderW = gl_window_width;
-    renderH = gl_window_height;
-  }
-  else
-  {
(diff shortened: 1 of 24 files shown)


fabiangreffrath:
Wow, this is really cool!

Pedro-Beirao:
Its cool and definitely very helpful in many situations, but for big maps it will still be pretty slow.
For going backwards, currently it will skip from the beginning (unless a keyframe is close to the desired time). Hopefully implementing skipping backwards isnt too hard

Pedro-Beirao:
Light blue lines are new playback_key_frames. They are spaced out in the demo to allow quick skipping. Aka they are my solution for "skipping backwards".

Green lines are rewind auto_key_frames.

https://github.com/user-attachments/assets/ad0ddfde-a162-4177-aab1-6730c2898635

Pedro-Beirao:
Finally should be all done.
Added window_rect and viewport_rect as opposed to the opengl-only variables we had before. Slowly improving how windows are handled.

Issue #487, opened 2024-04-07
dumb project is archived


chenrui333:
👋 just a heads up that dump project has been unmaintained for quite some time and now it is archived.

relates to https://github.com/Homebrew/homebrew-core/pull/168144

chenrui333:
https://github.com/Homebrew/homebrew-core/pull/228916

PR #678, opened 2025-05-25
Add a tracker music player based on libopenmpt

qookei:
Intended as an alternative to the DUMB player since DUMB is unmaintained (see bug #487).

I don't have a Windows machine handy so I didn't test building with msvc and vcpkg (but I did make sure a vcpkg package for libopenmpt exists).

`Findlibopenmpt.cmake` was based on `Findlibzip.cmake` (basically just `s/zip/openmpt/g`) and might not be 100% correct, I'd appreciate someone else taking a look over it, as I'm not well-versed in CMake.


Pedro-Beirao:
@ceski-1 can I ask you (or anyone that understands audio) to review this?

ceski-1:
I'll defer to @rfomin since he handled the equivalent for Woof (libxmp).

prboom2/cmake/Findlibopenmpt.cmake
```diff
@@ -0,0 +1,118 @@
...
+
+find_path(
+  libopenmpt_INCLUDE_DIR
+  NAMES libopenmpt/libopenmpt.h
+  HINTS "${PC_LIBOPENMPT_INCLUDEDIR}"
+)
+
+find_file(
+  libopenmpt_DLL
+  NAMES libopenmpt.dll libopenmpt.dll
```
rfomin: There is duplication of strings. It should be `NAMES openmpt.dll libopenmpt.dll`

prboom2/src/MUSIC/libopenmptplayer.c
```diff
@@ -0,0 +1,184 @@
...
+static void mpt_render (void *dest, unsigned nsamp)
+{
+  short *sdest = (short *)dest;
+
+  if (mpt_cur_module && !mpt_paused)
+  {
+    size_t nout, i;
+    float *buf = malloc(sizeof(float) * nsamp * 2);
+
+    nout = openmpt_module_read_interleaved_float_stereo(mpt_cur_module,
```
rfomin: Why use `openmpt_module_read_interleaved_float_stereo` instead of just `openmpt_module_read_stereo`?

prboom2/src/MUSIC/libopenmptplayer.c
```diff
@@ -0,0 +1,184 @@
...
+                    mpt_samplerate, nsamp, buf);
+
+    for (i = 0; i < nout; i++)
+    {
+      sdest[i * 2 + 0] = (short) (buf[i * 2 + 0] * mpt_multiplier);
+      sdest[i * 2 + 1] = (short) (buf[i * 2 + 1] * mpt_multiplier);
```
rfomin: How about setting gain with [openmpt_module_set_render_param](https://lib.openmpt.org/doc/group__libopenmpt__c.html#gaecd0bb67bb3b720340bb12f5df86db2a) instead of using `mpt_multiplier`?

rfomin (commented):
Thank you for your contribution. Should we replace libdumb?

fabiangreffrath:
> Thank you for your contribution. Should we replace libdumb?

Yes, please.

prboom2/src/MUSIC/libopenmptplayer.c
qookei: Because I kinda missed the `OPENMPT_MODULE_RENDER_MASTERGAIN_MILLIBEL` render param when looking at the docs :^) I'll change that in a moment.

prboom2/src/MUSIC/libopenmptplayer.c
qookei: I figured since the code was doing the volume math on it's own, doing it on floats would yield better precision, but as noted below we can let libopenmpt do the volume math for us and just use `openmpt_module_read_stereo` directly into the buffer instead.

fabiangreffrath:
> I'll defer to @rfomin since he handled the equivalent for Woof (libxmp).

Btw, SDL2_Mixer nowadays prefers libxmp over libopenmtp, if both are found. Any reason why you prefer the latter?

qookei:
> Btw, SDL2_Mixer nowadays prefers libxmp over libopenmtp, if both are found. Any reason why you prefer the latter?

I'm not too experienced when it comes to audio programming, and wasn't aware of libxmp. libopenmpt is explicitly mentioned in the description of https://github.com/kode54/dumb as an alternative. I can also rework the PR to make it use libxmp instead if that's desired.

fabiangreffrath:
> I'm not too experienced when it comes to audio programming, and wasn't aware of libxmp. libopenmpt is explicitly mentioned in the description of https://github.com/kode54/dumb as an alternative. I can also rework the PR to make it use libxmp instead if that's desired.

No problem for me. I just wanted to know if there was a specific reason to choose the one library over the other.

qookei:
I've resolved the comments. 

> > Thank you for your contribution. Should we replace libdumb?
> 
> Yes, please.

I can add a commit removing DUMB support to this PR.

prboom2/src/MUSIC/libopenmptplayer.c
```diff
@@ -0,0 +1,195 @@
...
+  int ret;
+
+  #pragma GCC diagnostic push
+  #pragma GCC diagnostic ignored "-Wcast-qual"
+  mpt_cur_module = (openmpt_module *) handle;
+  #pragma GCC diagnostic pop
+  ret = openmpt_module_set_repeat_count(mpt_cur_module, looping ? -1 : 0);
+
+  if (!ret)
+    mpt_cur_module = NULL;
```
rfomin: We should return here if `openmpt_module_set_repeat_count` fails. Perhaps we shouldn't check for this error since it's unlikely to fail.

Pedro-Beirao:
> I can add a commit removing DUMB support to this PR.

Yes please

prboom2/src/MUSIC/libopenmptplayer.c
```diff
@@ -0,0 +1,195 @@
...
+    lprintf(LO_WARN, "mpt_setvolume: failed to set current module master gain\n");
+}
+
+static const void* mpt_registersong (const void *data, unsigned len)
+{
+  openmpt_module *mod = NULL;
+
+  mod = openmpt_module_create_from_memory2(data, len, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
+  if (!mod)
+    lprintf(LO_WARN, "mpt_registersong: failed to open song\n");
```
rfomin: An effort was made to clean up terminal messages in DSDA-Doom compared to PrBoom+. We should set the `LO_DEBUG` level for this `lprintf`; otherwise, it will always print an error when a regular MIDI file is registered.

Another issue is that `openmpt_module_create_from_memory2` always prints "openmpt: openmpt_module_create_from_memory2: ERROR: error loading file" if non-tracker music is registered. This is not a big problem, but perhaps we could try defining [`openmpt_error_func`](https://lib.openmpt.org/doc/group__libopenmpt__c.html#gaa68155b220861de6218bc6b599071f21) or [`openmpt_log_func`](https://lib.openmpt.org/doc/group__libopenmpt__c.html#gaeec3c004981950af696bfd2688e250b2).

rfomin (commented):
Thanks, works good for me.

rfomin:
> I'll defer to @rfomin since he handled the equivalent for Woof (libxmp).

I chose libxmp because the vcpkg build of libopenmpt was broken at the time. libxmp seems cleaner and simpler to me, but I don't know much about tracker music.

fabiangreffrath:
> Btw, SDL2_Mixer nowadays prefers libxmp over libopenmtp, if both are found. Any reason why you prefer the latter?

Apparently, I have mixed up openmpt and modplug. Support for the former is currently considered for SDL_mixer, but libxmp is still the default. 

https://github.com/libsdl-org/SDL_mixer/issues/590

prboom2/src/MUSIC/libopenmptplayer.c
qookei: After a quick glance at the libopenmpt sources, I think the only way for this to fail is to feed it a null pointer as the module, so yeah I think we can just get rid of the check.

prboom2/src/MUSIC/libopenmptplayer.c
qookei: Will change the log level. For libopenmpt messages just a custom `openmpt_log_func` is enough, the default error func just logs via whatever the log func is (and tries to store the message but we don't give it anywhere to do so). I've made the custom log func just print the message with the debug log level as well.

qookei:
I've addressed the comments and added a commit that removes DUMB support and updates the various documentation files to mention libopenmpt instead.

rfomin (approved):
LGTM

Pedro-Beirao:
Thank you for the PR! And thank you for the review!

PR #683, opened 2025-05-29
OpenGL fixes for both dark overlay and statusbar background color

andrikpowell:
Here's the lowdown (I'll go more in-depth below):
- Create new indexed reference for menu elements (currently used for menu overlay and stbar color)
- `V_IsMenuLightmodeIndexed()` forces the main colormap (like `V_IsAutomapLightmodeIndexed()`)
- This fixes OpenGL inverting of the overlay and the statusbar background colour during Invuln or really any colormap change
- Use `playpal_lightest` and `playpal_darkest` for invuln `gld_DrawShaded()` check.

Ok, so here's the problem I ran into. In some WADs the dark overlay would sometimes invert to white when the Invulnerability colormap was active. I already knew about this, but didn't exactly know of a a good fix. However I ran into the exact same issue with the new "statusbar background color" option I added, as I found that would also invert during Invulnerability. It was then that I was able to confirm my suspicions when I checked a WAD that included underwater colormaps, that it also changed the colour of the statusbar background.

So the solution was to add a new Indexed lightmode item to track. This is already done with the automap, and it is used to force the automap to use ONLY the main colormap. And so I did the same thing but for menu elements. The two things that it's currently being used for is the statusbar color background and the menu overlay. This method seems to fix the issues that I was running into before, as we do want palette changes to affect these UI elements (berserk, item pickup, rad suit), but we don't want them to change based on colormap changes.

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -526,10 +526,10 @@ void D_Display (fixed_t frac)
     D_DrawPause();
   }
 
-  V_BeginUIDraw();
+  V_BeginMenuDraw();
   if (M_MenuIsShaded())
     M_ShadedScreen(0);
-  V_EndUIDraw();
+  V_EndMenuDraw();
 
   // menus go directly to the screen
   M_Drawer();          // menu is drawn even on top of everything
--- a/prboom2/src/gl_light.c
+++ b/prboom2/src/gl_light.c
@@ -51,6 +51,7 @@
 
 dboolean gl_ui_lightmode_indexed = false;
 dboolean gl_automap_lightmode_indexed = false;
+dboolean gl_menu_lightmode_indexed = false;
 
 static float lighttable[5][256];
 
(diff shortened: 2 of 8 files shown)


andrikpowell:
Alright, just forced pushed!

Basically I found that I could dramatically simplify the `DrawShaded` color logic now that `V_IsMenuLightmodeIndexed()` is being used.

Since I'm no longer using `V_IsUILightmodeIndexed()` for either the overlay and stbar color background, other colormaps are never applied. This means there's no longer a need to have to "undo" the invuln invert.

andrikpowell:
Alright another force push. I found a more simplified way to deal with the `V_BeginMenuDraw()` > `V_EndMenuDraw()` block.

In Nyan Doom, I have moved where this code is due to it needing to be refreshed for the stbar colors able to be changed via animation... But for now I don't think it's needed here yet.

Although usually I do like to keep both Nyan/DSDA code consistent...

I also have a better method for grabbing average patch colours, but I'll leave that for a separate PR.

Pedro-Beirao:
Thank you!

Issue #680, opened 2025-05-28
Blue colored text in dsdahud using status bar font is too bright


thekovic:
Whereas the green colored text matches the vanilla status bar, the blue colored text displays too bright with a glowing outline.

Vanilla status bar:

![Image](https://github.com/user-attachments/assets/6aaef2df-3384-49e1-ab77-5cc53731e1aa)

Extended status bar:

![Image](https://github.com/user-attachments/assets/207ee722-fc43-4b88-9eb2-386f277efb3b)

Pedro-Beirao:
A future update to dsdahud will address this, but its not a priority right now.

Calinou:
I actually find the bright blue more readable here. The previous blue color always looked too dark to me compared to all other colors (beige, red, yellow, green). It was often difficult to read in practice and didn't convey the "supercharged" feeling as well.

This is what it'd look like with bright blue in the status bar (mockup, so the `%` sign is unchanged):

![Image](https://github.com/user-attachments/assets/467f8dcf-9c11-4bc4-b9a4-b61a8b8a93bd)

It may be worth its own option, but I'm not sure. If you use a `COLORMAP` mod that replaces the shades of blue with less saturated variants, it might look good enough at the cost of affecting general game rendering as well.

Pedro-Beirao:
> I actually find the bright blue more readable here. The previous blue color always looked too dark to me compared to all other colors (beige, red, yellow, green). It was often difficult to read in practice and didn't convey the "supercharged" feeling as well.

You can use DSDATC to set the stbar armor to use lightblue instead of blue.
`stbar_armor_two 11`

Issue #679, opened 2025-05-28
Allow non-colorcoded "big_" elements in dsdahud


thekovic:
`dsdahud.lmp` components that use the status bar font (`big_ammo`, `big_armor`, `big_health`) are implicitly color-coded, similar to the Display->Status Bar->Coloring->Colored Numbers option for vanilla status bar. I would appreciate the option to have these elements use the vanilla coloring (for example, red when playing DOOM2.WAD).

Pedro-Beirao:
A future update to dsdahud will address this, but its not a priority right now.

PR #684, opened 2025-05-29
Various Colored Number Fixes

andrikpowell:
PR fixes both #679 and #680 as well as allowing "Gray %" regardless of whether "Colored Numbers" is enabled.

Usually I wouldn't bother, seeing that the dsdahud needs an update in the future anyway, but it was easy and quick enough :)

### DIFF
--- a/prboom2/src/dsda/hud_components/ready_ammo_text.c
+++ b/prboom2/src/dsda/hud_components/ready_ammo_text.c
@@ -25,6 +25,21 @@ typedef struct {
 
 static local_component_t* local;
 
+int dsda_AmmoColor(player_t* player) {
+  int ammo_percent;
+
+  ammo_percent = P_AmmoPercent(player, player->readyweapon);
+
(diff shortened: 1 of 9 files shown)


prboom2/src/dsda/hud_components/ready_ammo_text.c
```diff
@@ -25,6 +25,21 @@ typedef struct {
 
 static local_component_t* local;
 
+int dsda_AmmoColor(player_t* player) {
+  int ammo_percent;
+
+  ammo_percent = P_AmmoPercent(player, player->readyweapon);
```
Pedro-Beirao, line 31: This function isnt used. Can be deleted right?

Pedro-Beirao: Nvm Im blind. It is used in this file.

Pedro-Beirao (approved):

PR #685, opened 2025-05-31
Add Customisable View and Weapon Bob

andrikpowell:
So on my discord (Nyan Doom Development), I got asked to port over the configurable view- and weapon- bobbing from Woof...

And so I did. Instead of a simple toggle on/off, each option has settings that increment by 25% (off, 25%, 50%, 75%, 100%). This is the same as in Woof, however Woof uses a thermo instead, whereas I just used choices instead. _(Note I renamed the option in the config to avoid new configs starting with 25% if set to on)_

Any thoughts on adding this? or any thoughts against it?

### DIFF
--- a/prboom2/src/dsda/configuration.c
+++ b/prboom2/src/dsda/configuration.c
@@ -1022,12 +1022,12 @@ dsda_config_t dsda_config[dsda_config_count] = {
     CONF_BOOL(1), NULL, STRICT_INT(1)
   },
   [dsda_config_viewbob] = {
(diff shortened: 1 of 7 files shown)


Pedro-Beirao:
I like this. No reason not to make make it a percentage like this.

Pedro-Beirao (approved):

Pedro-Beirao:
We'll use a THERMO in a future update to the menu.
Thanks for the PR!

PR #687, opened 2025-06-10
Fix pixel brightness for certain crosshair styles

thekovic:
The assets for crosshair styles Cross, Angle, Dot, and Small (CROSS1 to CROSS4) had their pixel brightness set to 191 as opposed to 255 of the other crosshair styles. This manifested as these crosshair styles rendering in a darker color than the one set in the options menu and could be exaggerated when running a PWAD that makes changes to the palette. Here's a comparison between Dot and Big style in DOOM2.WAD and AALIENS.WAD.

image

This PR sets pixel brightness to 255 across all crosshair styles to make them consistent and equally bright.


Pedro-Beirao:
Thank you!

PR #688, opened 2025-06-14
option to display things as hitboxes on automap

vadosnaprimer:
options - automap - things appearance

- triangle tells the facing but not how things will collide (would it be useful to still indicate facing?)
- hitbox display is always scaled because it needs to appear accurate to in-game
- new enum value is before max to not mess with existing configs
- player hitbox appears too even tho it somewhat duplicates the trail, but maybe it's useful because it shows the final position on top without interfering with the trail itself

e7b96fcf786fb550

3144171650592b9f


Pedro-Beirao:
Thank you!

PR #689, opened 2025-06-16
CMake: Enable cross-compilation in one build

FtZPetruska:
This PR is an attempt at streamlining cross-compilation, especially around building `dsda-doom.wad`.

The previous system, as far as I could tell, relied on having already built `rdatawad` on the host system (exporting a CMake target through `export(TARGETS ...)` in the process), and then loading that target during cross-compilation by setting `IMPORT_EXECUTABLES` to the export file. This works, however, it requires having all the dependencies for both the target system but also the host system since the project is fully configured twice.

The approach used in this PR is to make the content of `data` a standalone CMake project and to rely on `ExternalProject` when cross-compilation is detected. What this module does is handling the configuration and building of the sub-project.
Conveniently, the generation of `dsda-doom.wad` is completely self-contained: `rdatawad` can be built on its own without needing to know anything about the rest of the build, and the output of the process is a single binary file that is only needed at runtime.

There were a few tweaks necessary to make `rdatawad` standalone, namely:
- not including `config.h`, which was not used in the first place anyway.
- adding `cmake_minimum_required` and `project` calls
- adding the custom target to the `ALL` target, so the default build step of the external project always builds the pwad
- installing the pwad in the top-level CMakeLists, technically it could be kept in the subproject but I find better to put all the `install` commands around the same place. See also the caveat with `INSTALL_BYPRODUCTS` in the next paragraph.

Here's a breakdown of the options given to ExternalProject:
- `BUILD_ALWAYS` is set so the build will be re-run every time. This may seem wasteful, but this is necessary if changes are made to `rdatawad`'s sources, or the lumps. And if no changes were made, nothing will be rebuilt.
- `BUILD_BYPRODUCTS` tell CMake that the build step is what is going to generate the pwad. There is also an `INSTALL_BYPRODUCTS` option, but that was only added in CMake 3.26, so I avoided direct installs from the sub-project.
- `INSTALL_COMMAND ""` disables the install step, since there is nothing to install
- `CMAKE_CACHE_ARGS` is used to set the cache for the sub-project.

For flexibility, a toolchain can be provided with `DSDA_HOST_TOOLCHAIN`, so if the default host compiler is for some reason unable to compile `rdatawad`, the user can provide a toolchain file to forward to the ExternalProject.

The main caveats with that approach are that compiler warnings and defines from the top-level CMakeLists may not be available, so I removed them altogether and added a `TODO`. I have a few ideas to address that, but did not want to clutter this PR with more changes than necessary.
Another issue is that CMake sets `CMAKE_CROSSCOMPILING` to true very eagerly. Simply setting `CMAKE_SYSTEM_NAME` manually, even if you set it to the same value as `CMAKE_HOST_SYSTEM_NAME`, will set `CMAKE_CROSSCOMPILING` to true. However, considering the size of the subproject, I only noticed a 0.3s slowdown when it incorrectly detects cross-compilation.
Lastly, `install(TARGETS ... RUNTIME_DEPENDENCIES ...)` is not supported when cross-compiling.

---

Testing that code path is as simple as setting `CMAKE_SYSTEM_NAME`, so if you're building on Linux that would be something like:
```
cmake \
  -S prboom2 \
  -B build \
  -D CMAKE_SYSTEM_NAME=Linux
```

In a real use scenario, this would look like:
```
cmake \
  -S prboom2 \
  -B build \
  -D CMAKE_TOOLCHAIN_FILE=toolchain-for-other-platform.cmake
```

And if you need a specific host toolchain, you would do it as such:
```
cmake \
  -S prboom2 \
  -B build \
  -D CMAKE_TOOLCHAIN_FILE=toolchain-for-other-platform.cmake \
  -D DSDA_HOST_TOOLCHAIN=host-toolchain.cmake
```

Where `host-toolchain.cmake` would contain:
```cmake
# CMAKE_SYSTEM_NAME should NOT be set
# At least a C compiler is needed for rdatawad
set(CMAKE_C_COMPILER my_special_compiler)

# If additional flags are required
set(CMAKE_C_FLAGS_INIT "extra-compiler-flag-here")
set(CMAKE_EXE_LINKER_FLAGS_INIT "extra-linker-flag-here")
...
```


rfomin (approved):
LGTM, thank you!

Pedro-Beirao:
Great! Thank you!

PR #690, opened 2025-06-16
option not to quit on demo end (quit by default)

vadosnaprimer:
Initially faced by us over at tasvideos because we always include game ending into encodes, but might be useful for all demos because you won't have to sit on the final screen just to make the stats clearly visible. You can end your demo as soon as possible to make that screen appear and then have shorter overall time, while also being able to show the final results as long as you want in the encode.

Old behavior is retained by default.


Pedro-Beirao:
After thinking more about this, Id prefer just having a setting for customizable intermission length, which would also apply to the end stats.
Guess this feature should just stay in your branch?

vadosnaprimer:
We're currently at my fork of prboom+. I was hoping we could stop depending on custom forks and switch to a regularly updated upstream once and for all.

It's hard to justify this change for TASers, and even regular players because vanilla doesn't have it. But isn't longtics in demos a similar case? I felt not quitting on demo end would be a nice little option that won't interfere with anyone's workflow while also giving DSDA-Doom more users (and potentially more improvements through networking).

Pedro-Beirao:
"little option that won't interfere with anyone's workflow"
I always try to avoid new options

What do the others think about this?

Pedro-Beirao:
Oh well, sure lets add this option. Thanks!

vadosnaprimer:
Thank you!!!

PR #691, opened 2025-06-18
Improved Stbar Color Background Algorithm

andrikpowell:
This is my improved patch color grabbing function.

It's also been split off in case it wants to be used in other functions. _(Nyan Doom uses it for both the stbar color background and the colored borderbox, but I'm guessing DSDA isn't interested in the later feature)_

### DIFF
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -501,65 +501,37 @@ void R_InitBuffer(int width, int height)
 void R_FillBackColor (void)
 {
   extern patchnum_t stbarbg;
-  byte col, col_top;
-  int i, j, pixel_cnt;
-  int r, g, b;
-  int stbar_top = SCREENHEIGHT - ST_SCALED_HEIGHT;
-  int ST_SCALED_BORDER = brdr_b.height * patches_scaley/2;
-  const unsigned char *playpal = V_GetPlaypal();
+  static byte col;
+  static byte col_top;
+  static int prevlump = -1;
+  const int stbar_top = SCREENHEIGHT - ST_SCALED_HEIGHT;
+  const int ST_SCALED_BORDER = brdr_b.height * patches_scaley/2;
+  int lump = stbarbg.lumpnum;
 
-  const byte* lump;
-  const byte* p;
-  short width;
-  byte length;
-  byte entry;
+  if (prevlump != lump)
+  {
+    const unsigned char *playpal = V_GetPlaypal();
+    SDL_Color stbar_color = V_GetPatchColor(lump);
+    int r = stbar_color.r;
+    int g = stbar_color.g;
+    int b = stbar_color.b;
 
-  pixel_cnt = 0;
-  r = g = b = 0;
+    // Convert to palette and tune down saturation
+    col = V_BestColor(playpal, r/3, g/3, b/3);
+    col_top = V_BestColor(playpal, r/2, g/2, b/2);
 
-  lump = W_LumpByNum(stbarbg.lumpnum);
-  width = *((const int16_t *) lump);
-  width = LittleShort(width);
+    // If colors are the same, brighten top
+    if (col_top == col)
+      col_top = V_BestColor(playpal, r, g, b);
 
-  for (i = 0; i < width; ++i) {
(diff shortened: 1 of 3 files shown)


prboom2/src/v_video.c
```diff
@@ -1201,6 +1201,130 @@ int V_GetPlaypalCount(void)
...
+          pixel_cnt++;
+        }
+
+        p++;
+      }
+    }
+  }
+  else
+  {
+    // Get color from all pixels
```
rfomin: Do you calculate the average color of the entire patch? Why is that better for DSDA-Doom? To be honest, I'm not sure if this feature belongs to DSDA-Doom.

andrikpowell: Currently this isn't getting used as it only averages the entire patch if the width is less than 16 pixels... But there's no reason for this to be there at this time... Although in the future, there may be a feature or two that rely on this condition. But that may only be in Nyan Doom, so I don't mind cutting this part here for now.

prboom2/src/r_draw.c
```diff
@@ -501,63 +501,36 @@ void R_InitBuffer(int width, int height)
...
-        b += playpal[3 * entry + 2];
-        pixel_cnt++;
-      }
+  static byte col;
+  static byte col_top;
+  static int r = 0, g = 0, b = 0;
+  static int prevlump = -1;
+  const int stbar_top = SCREENHEIGHT - ST_SCALED_HEIGHT;
+  const int ST_SCALED_BORDER = brdr_b.height * patches_scaley/2;
+  const int lump = stbarbg.lumpnum;
```
Pedro-Beirao: No need for rgb to be static.

prboom2/src/v_video.h
```diff
@@ -299,6 +299,13 @@ void V_FreePlaypal(void);
 // [XA] get number of palettes in the current playpal
 int V_GetPlaypalCount(void);
 
+typedef struct ColorEntry_s
+{
+  int r, g, b;
+} ColorEntry_t;
+
+ColorEntry_t V_GetPatchColor (int lumpnum);
+
```
Pedro-Beirao, line 305: PalEntry_s and SDL_Color already exist for exactly this.
Id prefer if SDL_Color was used everywhere.

prboom2/src/v_video.c
```diff
@@ -1201,6 +1201,99 @@ int V_GetPlaypalCount(void)
...
+          col.r += playpal[3 * entry + 0];
+          col.g += playpal[3 * entry + 1];
+          col.b += playpal[3 * entry + 2];
+          pixel_cnt++;
+        }
+
+        p++;
+      }
+    }
+  }
```
Pedro-Beirao, line 1254: Instead of duplicating the for loop, you can use a bit of logic

```c
for (i = 0; i < width; ++i) {
  byte length;
  byte entry;
  const byte* p;
  int32_t offset;

  // Only calculate for the leftmost and rightmost 16 columns
  if (width > 32 && i > 16 && i < width - 16)
    continue;

  // Skip irrelevant data in the doom patch header
  p = lump + 8 + 4 * i;
  offset = *((const int32_t *) p);
  p = lump + LittleLong(offset);

  while (*p != 0xff) {
    p++;
    length = *p++;
    p++;

    // Get RGB values per pixel
    for (j = 0; j < length; ++j) {
      entry = *p++;
      col.r += playpal[3 * entry + 0];
      col.g += playpal[3 * entry + 1];
      col.b += playpal[3 * entry + 2];
      pixel_cnt++;
    }

    p++;
  }
}
```
    
This way it also works for < 16 width patches
No need for the `if (width >= 16)`

andrikpowell: Much better, used this logic! 👍

prboom2/src/v_video.h
andrikpowell: Used `SDL_Color` instead 👍

prboom2/src/r_draw.c
andrikpowell: Sorry went too hard on the static. Fixed.

Pedro-Beirao (approved):
Thanks!

PR #692, opened 2025-06-18
Rework Centering of Large Patches

andrikpowell:
- Only center patches larger than 320 for specific cases (backgrounds, stbar, levelname)
- Fixes widescreen assets that rely on offsets (Example Doom E3 intermission wide assets)

Also took the time to restructure the Doom 1 finale sequence, so it doesn't call `V_DrawNamePatch` a bunch of times.

Note that Woof takes this approach having a separate "FullScreen" patch drawing function.


Pedro-Beirao (approved):
Looks good!

PR #695, opened 2025-06-18
Automap Mark Tweaks

andrikpowell:
- Skip "Cleared spot 0" for clearing marks (since if you've cleared mark 0, all marks are cleared)
- Allow raven games to add marks


Pedro-Beirao (approved):
`namelen` would be slightly more appropriate, but doesnt matter

andrikpowell:
> `namelen` would be slightly more appropriate, but doesnt matter

Extremely easy tweak, so I just did it :)

PR #696, opened 2025-06-18
Mark Death Exits on the Automap

andrikpowell:
Decided to use both Exit and Secret Exit map colours for death exits.
Disabled in Raven (there are no death exits).


Pedro-Beirao (approved):
Looks right

PR #698, opened 2025-06-19
CMake: Fix Visual Studio ARM64 cross-compilation

FtZPetruska:
This is a follow-up to #689, as it turns out the previous PR was not enough to support cross-compiling with the Visual Studio generator.

If you want to target Windows on ARM64 from x64 with the Visual Studio generator, you typically only pass to CMake `-A ARM64` instead of a toolchain file, which leads to these two problems:
- `CMAKE_CROSSCOMPILING` is not set, so it would try to compile for the target and fail to run `rdatawad`.
- Even when forcefully set, `ExternalProject` re-uses the parent's project generator platform (`-A ARM64`), and `rdatawad` gets built for the target platform once again.

This PR aims to solve that issue by:
- Always using `ExternalProject` when Visual Studio is the generator.
- Telling `ExternalProject` to use the host's generator platform (this is thankfully provided by CMake through `CMAKE_VS_PLATFORM_NAME_DEFAULT`).

Apologies for the inconvenience.


rfomin (approved):
Thank you!

PR #700, opened 2025-06-20
CMake: Apply flags and defines through functions 

FtZPetruska:
This PR reworks how warnings, fast math flags, and defines were checked and set to better match how CMake target works. And as a follow-up to #689, this enables warnings back and silences deprecation on `rdatawad`.

The main idea is that it moves from the error-prone:
```cmake
get_my_flags(MY_COMPILE_FLAGS)
...
target_compile_options(my_target PRIVATE ${MY_COMPILE_FLAGS}) 
```

To:
```cmake
target_set_my_flags(my_target)
```

---

Here's the long breakdown of what it does:

- To apply warnings, `dsda_target_set_warnings(tgt)` will add the supported warnings to `tgt` compile options. If the default set of warnings is not supported on the compiler, `DSDA_ENABLED_WARNINGS` can be manually set to a list of flags the user wants.
- To silence deprecations, `dsda_target_silence_deprecation(tgt)` will add the needed defines to `tgt`. On Apple platforms, `GL_SILENCE_DEPRECATION` is added to silence the many OpenGL deprecation warnings.
- To enable fast math, `dsda_target_enable_fast_math(tgt)` will add the needed flags to `tgt` compile options. If none of the default flags are supported by the compiler, `DSDA_FAST_MATH_FLAG` can be manually set to a list of flags to provide the feature.


rfomin (approved):
LGTM, thanks.

PR #703, opened 2025-06-27
CMake: Move config.h generation to its own module.

FtZPetruska:
This moves all config.h-related checks and target settings to its own module.

In addition, a basic check is added to verify all the needed CMake variables are defined.


Pedro-Beirao (approved):
Looks good. Thanks!

PR #709, opened 2025-07-04
CMake: Group and update options

FtZPetruska:
This PR moves all CMake options into two modules:
- `DsdaDepsSetup`, for all dependencies and vcpkg options, as they should be set *before* `project` is called.
- `DsdaOptions`, for all the other options.

Additionally, a few tweaks are made:
- vcpkg options are set in cache with a default value instead of being forcefully enabled, and are only set if the vcpkg toolchain is detected.
- Instead of forcing the install directory of the license file and the binary on Windows, the cache settings `DSDA_INSTALL_COPYRIGHT_DIR` and `DSDA_INSTALL_BINDIR` are added and default to the same values used on each platforms before this patch.
- The `ENABLE_LTO` option **is removed**. Setting `CMAKE_INTERPROCEDURAL_OPTIMIZATION` offers the exact same functionality, but also let's the user selectively enable it for specific build config by setting `CMAKE_INTERPROCEDURAL_OPTIMIZATION_<build config>`. A warning is printed if `ENABLE_LTO` is set, encouraging users to use CMake's built-in variable.

### DIFF
--- a/prboom2/CMakeLists.txt
+++ b/prboom2/CMakeLists.txt
@@ -1,38 +1,9 @@
 cmake_minimum_required(VERSION 3.9)
 
-option(WITH_XMP "Use libxmp if available" ON)
-if(WITH_XMP)
-    list(APPEND VCPKG_MANIFEST_FEATURES "libxmp")
-endif()
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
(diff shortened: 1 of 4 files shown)


Pedro-Beirao (approved):
Much cleaner. Thanks!

FtZPetruska:
Moving it back to draft.

I'm waiting for #721 (or its alternative) to get merged to avoid conflicts.

rfomin:
Thank you!

PR #710, opened 2025-07-07
Catch crash if animation has speed 0

Pedro-Beirao:
Division by zero when animation speed is 0. Somehow it doesnt crash on mac, but does on windows and linux.
Woof also crashes, it would be good to catch it there as well.

[animcrash.wad.zip](https://github.com/user-attachments/files/21093562/animcrash.wad.zip)

### DIFF
--- a/prboom2/src/p_spec.c
+++ b/prboom2/src/p_spec.c
@@ -246,16 +246,22 @@ void P_InitPicAnims (void)
 
     lastanim->istexture = animdefs[i].istexture;
     lastanim->numpics = lastanim->picnum - lastanim->basepic + 1;
+    lastanim->speed = LittleLong(animdefs[i].speed);
 
     // [crispy] skip reading SMMU swirling flats
     if (lastanim->speed < 65536 && lastanim->numpics != 1)
-    { 
+    {
       if (lastanim->numpics < 2)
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
I think you can catch this much earlier during parsing the lump:
https://github.com/kraflab/dsda-doom/blob/21d4f43501261188256d0b524a1dc51980389460/prboom2/src/p_spec.c#L258

BTW, why is `lastanim->speed < 65536` checked before it is set?

Pedro-Beirao:
> I think you can catch this much earlier during parsing the lump:

If some existing wad acidentally set an unused animation to 0, then it would exit here (even though no maps uses it and it wouldnt crash further along). So I changed it to when it is used.
Do you think its fine to do it when parsing?

> BTW, why is lastanim->speed < 65536 checked before it is set?

Yea thats wrong... the condition was introduced to prevent crash with swirling flats. I guess in the exact wad this was tested with it worked? I can fix that later today.

fabiangreffrath:
> Do you think its fine to do it when parsing?

Yes, this sounds reasonable! 

> I can fix that later today.

Thank you!

fabiangreffrath (approved):
Looks good, thank you!

PR #712, opened 2025-07-10
Dont delete save using Enter

Pedro-Beirao:
> A usability issue I've been having:
This input prompt accepts ENTER as Y, which is a problem for me because ENTER is very close to DEL, so I occasionally fat-finger both of those keys and boom, I permanently lose a save-game while just trying to load. It would be nice if this prompt only literally accepted Y as a confirmation input for safety.

### DIFF
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -5546,7 +5546,7 @@ typedef enum {
 
 static confirmation_t M_EventToConfirmation(int ch, int action, event_t* ev)
 {
-  if (ch == 'y' || action == MENU_ENTER)
+  if (ch == 'y' || (!delete_verify && action == MENU_ENTER))
     return confirmation_yes;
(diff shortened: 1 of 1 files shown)


ceski-1:
With this change, there's no way to delete a save using a controller.

Pedro-Beirao:
> With this change, there's no way to delete a save using a controller.

True... dsda's controller support isnt great, so I dont usually take that in mind. And this is a type of request that I dont put much thought into...
Still I think its a valid request to not want Enter to delete saves.

Any suggestion?

ceski-1:
> Any suggestion?

Maybe ignore any confirmation for n tics? Or just accept the occasional accidental press.

fabiangreffrath:
> Maybe ignore any confirmation for n tics? Or just accept the occasional accidental press.

I agree. That's probably nothing a port is supposed to fix.

PR #713, opened 2025-07-10
Make ST_HealthColor values inclusive

Pedro-Beirao:
Just to make it consistent with `health <= hud_health_green`

### DIFF
--- a/prboom2/src/st_stuff.c
+++ b/prboom2/src/st_stuff.c
@@ -828,9 +828,9 @@ void M_ChangeApplyPalette(void)
 
 int ST_HealthColor(int health)
 {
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
Typical case of "it has always been like this":

https://github.com/Doom-Utils/historic-ports/blob/3ae2d25740b9d4ea896f0fde0c73f6b112135ce5/st_stuff.c#L708-L716

Pedro-Beirao:
> Typical case of "it has always been like this":

Doesnt hurt to change right? Coincident asked in #dsda-doom, and I can see it be annoying for people that customize this setting.

PR #715, opened 2025-07-12
rdatawad: Cleanup header inclusion

FtZPetruska:
This PR updates the headers and how they're included in rdatawad's source.

Previously they sort of work through "happy accidents", there were no header guard at all, and some headers relied on being included after `size_t` was defined instead of including `stddef.h`.

This PR attempts to clean that mess a little, by:
- Adding header guards, so things don't explode if a header is included twice.
- Making all header self-contained, so they can be included on their own.
- Updating the inclusions done in .c files (drop unused headers, group them by type, and sort them).

### DIFF
--- a/prboom2/data/rd_graphic.c
+++ b/prboom2/data/rd_graphic.c
@@ -3,14 +3,14 @@
 
 // Convert portable pixmap to Doom patch format
 
-#include <stdlib.h>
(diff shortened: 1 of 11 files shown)


fabiangreffrath (approved):
Very good, thank you!

PR #720, opened 2025-07-14
Fix module looping

### LINKED ISSUE #719: Tracker music doesn't loop correctly
I'll make a PR for this.

Report 1:

> Just found a bug with IT module playback
> In 0.29.2, when the IT module music in this map is finished playing (about 9 minutes in), it does not loop back to the start and the playback just stops right there. In 0.29.0 the music loops back to the start normally after reaching the end.
> Haven't tested this with other IT files or other tracker module formats
> https://www.doomworld.com/idgames/levels/doom2/Ports/p-r/portglacia

Report 2:

> dsda has some trouble playing the music on NoSp3 MAP19
> fails to loop
> https://www.doomworld.com/idgames/levels/doom2/Ports/megawads/nosp3

Xaser's comment:

> the module itself is being naughty, unfortunately -- it uses a global volume fade at the end of the track, but doesn't set it back to max volume at the beginning, so after fading it out stays at zero volume after the loop

Xaser's examples:

> the music suite for dead-wire works as a sub-song test-case -- basically check if the first ambient subtrack still loops [mad-labs.zip](https://github.com/user-attachments/files/21208228/mad-labs.zip)
> 
> here's an example of one that seemingly doesn't loop, because it fades out the global volume and doesn't reset it (i.e. the common oopsie) [fusionfaktor.zip](https://github.com/user-attachments/files/21208230/fusionfaktor.zip)

All of the above in one wad: [tracker_songs.zip](https://github.com/user-attachments/files/21208236/tracker_songs.zip)
```
Name       Time        Source
D_RUNNIN   0:09:12.94  PORTGLACIA.wad MAP01 D_RUNNIN
D_STALKS   0:05:53.98  mad-labs.it
D_COUNTD   0:04:43.41  fusionfaktor.it
D_BETWEEE  0:03:04.30  NoSp3.wad MAP19 D_SHAWN2
```

ceski-1:
Fixes https://github.com/kraflab/dsda-doom/issues/719

### DIFF
--- a/prboom2/src/MUSIC/libopenmptplayer.c
+++ b/prboom2/src/MUSIC/libopenmptplayer.c
@@ -68,6 +68,7 @@ const music_player_t mpt_player =
 static int mpt_samplerate;
 static int mpt_paused;
 static int mpt_gain;
+static int mpt_looping;
 static openmpt_module *mpt_cur_module;
 
 static const char *mpt_name (void)
@@ -140,7 +141,7 @@ static void mpt_play (const void *handle, int looping)
   #pragma GCC diagnostic ignored "-Wcast-qual"
   mpt_cur_module = (openmpt_module *) handle;
   #pragma GCC diagnostic pop
-  openmpt_module_set_repeat_count(mpt_cur_module, looping ? -1 : 0);
+  mpt_looping = looping;
 
   ret = openmpt_module_set_render_param(mpt_cur_module,
                   OPENMPT_MODULE_RENDER_MASTERGAIN_MILLIBEL,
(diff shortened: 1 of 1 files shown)


rfomin:
It seems GZDoom uses the same approach: https://github.com/ZDoom/ZMusic/blob/a70ec38e4b83b15d6f2ae8a01521bf80ae51a7d1/source/streamsources/music_libxmp.cpp#L144-L149

I vote for using this looping method and switching to libxmp for consistency with the other ports, such as GZDoom, Helion (which also uses ZMusic), Woof, and SDL_Mixer.

PR #721, opened 2025-07-15
Switch to libxmp for tracker music

### LINKED ISSUE #719: Tracker music doesn't loop correctly

ceski-1:
Fixes https://github.com/kraflab/dsda-doom/issues/719
Alternative to https://github.com/kraflab/dsda-doom/pull/720


prboom2/src/MUSIC/xmpplayer.c
```diff
@@ -0,0 +1,188 @@
...
+{
+  if (xmp_playing && !xmp_paused)
+  {
+    if (xmp_play_buffer(context, dest, nsamp * 4, xmp_looping ? 0 : 1) == 0)
+    {
+      short *sdest = (short *)dest;
+
+      for (int i = 0; i < nsamp * 2; i++)
+      {
+        sdest[i] = (short)(sdest[i] * xmp_volume);
```
rfomin, line 150: We can use the  [xmp_set_player](https://xmp.sourceforge.net/libxmp.html#int-xmp-set-player-xmp-context-c-int-param-int-val) function with the `XMP_PLAYER_VOLUME` master volume parameter.

rfomin (commented):
Thank you!

prboom2/src/MUSIC/xmpplayer.c
ceski-1: That's the first method I tried and it didn't work correctly. It seems like the song can adjust that value whenever it wants.

rfomin: Okay, I understand.

rfomin (approved):

prboom2/vcpkg.json
FtZPetruska, line 1: Just a minor nitpick: `vcpkg format-manifest prboom2/vcpkg.json` should be run on the manifest.

ceski-1: Fixed, thanks.

ceski-1:
Okay, this is ready to go (use squash merge).

axunes:
I would like to ask if the removal of the library is necessary. libxmp is great but it's missing some of the module formats' features (fm support for scream tracker 3 modules, amiga resampler) as well as loads of quality-of-life features (known as MPT hacks) from libopenmpt (accurate configurable tempo, dsp effects, song- and instrument-defined interpolation modes, and a bunch of extra commands to make up for limitations within the original formats)

a lot of that may sound like fluff, but along with configurable interpolation, the accurate tempo mode in particular is sorely missed, it allows for a configurable tick rate for smoother pitch slides and such, where normally you are limited to historical frame-based timings for computers with wavetable mixing hardware, which gives even less precision than doom's original MUS format. with software mixing that is not a concern, so it would be nice to have the option of libopenmpt which is not only designed to accurately play modules of the time but also add features necessary to accommodate the modules being created today

elf-alchemist:
The main concern with using libxmp is cross-port consistency. If you look at many modern, actively developed ports, Crispy (via SDL_Mixer), Woof, DSDA, U/GZDoom (via ZMusic), Helion (also via ZMusic) are all using using libxmp under the hood, this is by design, there are many fundamental glaring differences between tracker implementations due to the lack of proper format standardization, which is incredibly lacking compared to MIDI, for example. This is highly problematic as every port can and will sound different from each other, and we will receive independent but fundamentally related bug reports for said inconsistent behavior everywhere. As such it is far easier for us to use the same library across the board _and_ send issues to said single library, so we will all receive the same updates from upstream, it would be far more costly, in both time and effort to submit these kinds of separate reports to multiple upstream libraries separately and independently, or even worse trying to make patches within the source ports using multiple workarounds. In short it is much simpler and less burdenous for port devs and composers alike to share the same support, it just makes our lives easier.

Related, but there have already been reports made to libxmp to add support for openmpt extensions, as per one U/GZDoom contributor noticing that Graf Zahl updated ZMusic to play tracker with OMPT extensions with libDUMB instead of sending a fix/report upstream. Similarly, a libopenmpt compat issue had also been reported as well:

https://github.com/libxmp/libxmp/issues/885

https://github.com/libxmp/libxmp/issues/886

If you can, I would also ask that you contribute your feature requests to their GitHub issues as well, unreported issues are forever untackled, after all.

PR #722, opened 2025-07-17
CMake: Bump minimum version to 3.17

FtZPetruska:
This PR bumps the minimum version of CMake to 3.17.

The main reasons for this change are:
- No more deprecation warning every time CMake is configured! And since the version is set as a range, there shouldn't be any for a decade or two.
- The project was actually using features from versions >3.9, most notably CPack's External generator is a 3.13 feature.
- When using vcpkg in completely static builds on macOS/Linux, it may fail the link step due to how link flags are handled.
- vcpkg's `X_VCPKG_APPLOCAL_DEPS_INSTALL` (which copies DLL to the install directory) requires CMake 3.14.

If bumping the minimum version sounds scary, here's the platform support for reference:
- macOS/Windows typically get the latest (or close to) CMake version through brew, chocolatey, or even Visual Studio.
- the current Debian oldstable ships with CMake 3.18.
- Installing a newer CMake version on Linux can be done rather trivially, either downloading the prebuilt binaries from cmake.org, or by running `pip install cmake`.

---

Don't hesitate to let me know if that is too high of a requirement for the project! At a bare minimum, 3.13 should be used to reflect what is actually used in the project, but I picked 3.17 as it ensures all the policies we want are available while having good platform support.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -29,7 +29,6 @@ jobs:
               -A x64
               -DCMAKE_TOOLCHAIN_FILE="${env:VCPKG_INSTALLATION_ROOT}\scripts\buildsystems\vcpkg.cmake"
               -DVCPKG_TARGET_TRIPLET=x64-windows-static-release
-              -DCMAKE_POLICY_DEFAULT_CMP0091=NEW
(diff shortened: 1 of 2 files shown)


rfomin (approved):
> If bumping the minimum version sounds scary

No, not at all. I just never got around to reading the manual on how to do it properly since CMake 4.0 broke a lot of things. Thank you!

PR #724, opened 2025-07-18
Rework Game Modifiers to be temporary

andrikpowell:
Alright, @Pedro-Beirao asked me to port over the Custom Skill functionality from Nyan Doom... But in order to do that, we are going to need to push this through.

This is because certain elements such as "pistol start" or "no monsters" are not part of the skill definition, but are instead separate flags (game modifiers). In order to allow these to be changed via the custom skill menu, we need to rework the game modifiers.

This PR does what Nyan Doom currently does and changes game modifiers to be temporarily stored during runtime and in savegames, instead of them being part of the config file.

The reasoning behind this change, is that some game modifiers like "coop spawns" and "no monsters" shouldn't be a permanent setting and should only be toggled via args (-coop_spawns) or turned on via the config in a temporary manner. I've made an exception with pistol start via a "always pistol start" option.

With how this functionality is laid out, it means that users can disable these settings in-game (previously not possible when using flags) if not during demos. This will also allow me to do a fix with the TNTFAST cheat, since I can just toggle the config option.


andrikpowell:
@Pedro-Beirao Alright, I've updated this PR based on our discussions in Discord.

Changes:
- "Pistol Start" is now a variable, and is treated similar to `respawnparm`, `fastparm`, etc.
- Renamed a few functions from "flags" to "modifiers"
- In save games, the parameters saved are now directly tied to `respawnparm`, `fastparm`, etc.
- Added additional comments for further clarity

Pedro-Beirao (approved):
This looks ok. There's some things Id prefer to be different but Ill do it after the skill builder PR itself.

andrikpowell:
I don't wanna be that guy, but I have a Skill Builder PR set up... Just waiting for this PR to go through to move forward.

PR #725, opened 2025-07-18
CMake: FluidSynth and libmad find module updates

FtZPetruska:
Just a few fixes for find modules:

- On Windows, fluidsynth adds a `lib` prefix and a `-3` suffix, so check for them too.
- Libmad's [fork from tenacityteam](https://codeberg.org/tenacityteam/libmad) (used by vcpkg) ships with a CMake package that provides the `mad::mad` target, so align with that.


rfomin:
Thank you!

PR #726, opened 2025-07-18
CMake: Make a helper for setting build config.

FtZPetruska:
This PR moves setting the default build configuration to its own helper.

Additionally, it better handles multi-config generators, setting the default there too when possible (currently only for `Ninja Multi-Config`), and not setting `CMAKE_BUILD_TYPE` when unnecessary.


Pedro-Beirao (approved):
Looks good. Thank you!

PR #727, opened 2025-07-19
Fix building with libxmp disabled

FtZPetruska:
xmpplayer.c fails building when `WITH_XMP` is disabled from NULL not being defined.

Other players deal with this by including `string.h` when the backend is disabled, this switches to `stddef.h` so it *hopefully* has clearer intent.


rfomin:
Thank you!

PR #728, opened 2025-07-19
CI: Fix unsupported warnings on macOS workflows

FtZPetruska:
Looking at the recent CI runs, macOS builds were not building with warnings due to missing `-Wno-format-truncation`.

This warning was only added on the Clang side in LLVM 18, and Apple Clang 17 (from Xcode 16.3). So only add it in those cases.


Pedro-Beirao (approved):
Version numbers look correct. Thank you!

PR #729, opened 2025-07-21
Fix warnings on mingw

FtZPetruska:
This PR brings a few fixes to have a warning-free build on an x86_64 mingw toolchain with GCC 15.

The warnings were:
- In `i_sound.c`, the cast to `(void*)` would trigger `-Wcast-qual`, it was possible to make the function just accept a const pointer.
- In `e6y.c` and `SDL/i_video.c`, building with LTO highlights that the global `mouse_hide_timer` had mismatched type declaration. It *seems* that the correct type is `int`, but correct me if I'm wrong.
- In `f_finale.c`, the call to `sprintf` would trigger `-Wformat-overflow`. Probably a false positive with the analyser not noticing the early returns with `finalecount`. Since the size of `name` is known, moving to `snprintf` is trivial and silences the warning, alternatively, clamping `stage` between 0 and 6 would also fix this.

### DIFF
--- a/prboom2/src/SDL/i_sndfile.c
+++ b/prboom2/src/SDL/i_sndfile.c
@@ -51,7 +51,7 @@ static sf_count_t sfvio_tell(void *user_data)
   return mem_ftell((MEMFILE *)user_data);
 } 
 
-void *Load_SNDFile(void *data, SDL_AudioSpec *sample, void **sampledata,
+void *Load_SNDFile(const void *data, SDL_AudioSpec *sample, void **sampledata,
                    Uint32 *samplelen)
 {
   SNDFILE *sndfile;
--- a/prboom2/src/SDL/i_sndfile.h
+++ b/prboom2/src/SDL/i_sndfile.h
@@ -18,7 +18,7 @@
 
 #include "SDL_audio.h"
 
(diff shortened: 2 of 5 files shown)


Pedro-Beirao (approved):
Thank you! Warning fixes are always appreciated.

On my machine running appleclang, there are so many warnings that I am not able to notice new ones, so for example that `mouse_hide_timer` one completely slipped. I need to go through and fix them somtime.

FtZPetruska:
> On my machine running appleclang, there are so many warnings that I am not able to notice new ones, so for example that `mouse_hide_timer` one completely slipped. I need to go through and fix them somtime.

Same here with AppleClang. A big chunk of the macOS warnings are from mismatched function prototypes and prototypes without arguments.
There also are *fun* type issues due to some platforms having `long` and `int` being different sizes, and Windows having 32-bits `long`. So attempting to fix a warning on not-Windows may lead to a warning on Windows, and vice-versa.

rfomin:
> There also are _fun_ type issues due to some platforms having `long` and `int` being different sizes, and Windows having 32-bits `long`. So attempting to fix a warning on not-Windows may lead to a warning on Windows, and vice-versa.

We should switch to the `stdint.h` type definitions (and C99 standard) at some point.

rfomin (approved):

PR #730, opened 2025-07-21
CMake: Handle dependencies through an interface.

FtZPetruska:
This moves all the dependency finding, linking, (and workarounds), to a single module.

To achieve this, an INTERFACE library is used to collect all the library dependencies, which is then linked to the executable.

---

As to the gritty details of the changes, the main update has to do with how SDL is handled. I went through all release versions from 2.0.7 to 2.32.8, building with CMake and autotools, as many linux distros often favour the latter to build. Here were my findings:
- While 2.0.7 was the minimum version required, if SDL was built with CMake, it could not be used with the current setup. CMake builds did not define `SDL2_INCLUDE_DIRS` and `SDL2_LIBRARIES` until 2.0.16, and the way include directories was setup did not allow for `#include "SDL.h"` until 2.0.9 (expecting `SDL2/SDL.h` instead).
- Before 2.0.12, the CMake package did not contain version information when built with autotools.
- `SDL2::*` CMake targets are guaranteed to exist starting 2.0.12, regardless of how SDL was built.
- Until 2.24.0, CMake packages from autotools builds were defining all targets, regardless of whether the actual underlying library exists on disk or not. (i.e. a static-only build would still create an `SDL2::SDL2` target, pointing to an inexistent file).

Knowing all this, this PR changes how finding SDL2 is handled:
- Require a minimum of 2.0.12, regardless of how the library was built, the CMake package provides version information and targets. This also allows dropping the find module altogether.
- Work around the issue with autotools build before 2.24.0 by checking whether the library files actually exist before linking, and wrap it all under the `dsda::SDL2` interface library.

In addition, a few little changes were made:
- Avoid creating target aliases with upstream names, preferring generator expressions or if/else.
- Prefer linking to `OpenGL::GL` and `OpenGL::GLU` targets instead of using their respective variables.
- On macOS, XQuartz ships with OpenGL 1.4 libraries, which may get picked instead of the system's `OpenGL.framework`. This would link successfully but fail at runtime. So make sure only the framework is searched for.


Pedro-Beirao (approved):
I like this, much cleaner. Thank you!

The XQuartz issue is a bit ugly, but I understand why it may be needed. Even though Ive never run into this problem having XQuartz installed.

FtZPetruska:
> The XQuartz issue is a bit ugly, but I understand why it may be needed. Even though Ive never run into this problem having XQuartz installed.

I've only hit it recently when building with vcpkg on my mac. Since I find it the most convenient to test having all static dependencies.

For some reasons, under these circumstances XQuartz's paths are searched before the macOS SDK. So their OpenGL gets picked up, it builds fine, but creating the window fails when starting up.

rfomin (approved):
LGTM, thank you!

PR #732, opened 2025-07-26
CMake: Add sanitizers options

rfomin:
Patch by @FtZPetruska 

UBSAN reports many `"member access within misaligned address"` errors in the save/restore code. I think we should fix them.


Pedro-Beirao:
I dont know anything about sanitzers setup, you can merge. Thank you!

PR #733, opened 2025-07-26
Fix errors reported by UBSan

rfomin:
There are only `"left shift of negative value"` errors left, which may not be worth fixing.

### DIFF
--- a/prboom2/src/m_swap.h
+++ b/prboom2/src/m_swap.h
@@ -59,32 +59,32 @@
 #ifdef HAVE_ASM_BYTEORDER_H
 #include <asm/byteorder.h>
 #ifdef __arch__swab16
-#define doom_swap_s  (signed short)__arch__swab16
+#define doom_swap_s  (int16_t)__arch__swab16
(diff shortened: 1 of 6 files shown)


Pedro-Beirao (approved):
Looks good thanks!

Its weird that it complains about P_ReplaceMobjWithIndex, but whatever.

rfomin:
> Its weird that it complains about P_ReplaceMobjWithIndex, but whatever.

The actual problem is the `P_SAVE_TYPE_REF` macro. One way to fix the alignment issue is to use an additional `memcpy`, but then the saving code will be slower.

Kraflab removed the padding in the old Doom code. See: <https://github.com/coelckers/prboom-plus/blob/969515162c5aebea4ad7a125ee178dcad3d576ad/prboom2/src/p_saveg.c#L53>
It seems that we can resolve the issue by restoring and updating it.

How important is misalignment? I remember at least one crash related to this issue in Woof on Linux. There are a lot of changes in this PR, so more testing is needed.

PR #734, opened 2025-07-26
CMake: Add option for strict dependency search

FtZPetruska:
This adds the CMake option `DSDA_STRICT`, which, when turned on, makes all the optional dependencies a requirement.

This option is mainly useful for CI and package manager. For CI, it ensures that dependencies do not silently stop being found, and for new dependencies, that they are installed and used on CI as well.

For package managers, turning it on allows raising a build error when the dependencies change when going from one version to another. A recent example of this is with Homebrew's package. When going from 0.29.0 to 0.29.1, the package was updated as-is, despite `dumb` being replaced with `libopenmpt`, and a follow-up PR had to be made to update the dependencies. With this new option enabled, the build would be failing until the dependencies are updated.


Pedro-Beirao (approved):
Yea thats a good reason to add this. Thanks!

We dont NEED to make everything have `dsda` prefix. I would even prefer to remove most `dsda` prefixes from the code.
Specially when we take into account forks like `nyan-doom` or whatever exists in the future.

`STRICT_FIND` maybe?

FtZPetruska:
The recommended practice for options (and everything that would be visible from having dsda being a subproject) is to prefix them so they do not clash with names other projects could use. The common justification is that "you should always assume that the project may end up as a child project of another one".

Now for DSDA, I don't see many ways this could happen, other than making a super-build arrangement that uses FetchContent to build all the dependencies along with DSDA itself...

Just to be clear, even though I'm in favour of having a prefix on options, I understand and would not mind removing it if you prefer to have no prefix. What do others think ?

rfomin:
> Just to be clear, even though I'm in favour of having a prefix on options, I understand and would not mind removing it if you prefer to have no prefix. What do others think ?

I also prefer to remove the prefix from the options.

rfomin (approved):
Thank you!

PR #735, opened 2025-07-27
Add "TITLEPIC/CREDIT cycle" for WADs with blank internal demos

andrikpowell:
PR does what the title says.

Due to the cycle being based on other code, I brought over Nyan functionality separating the types of CREDITS (graphic and "dynamic"), as well as allowing the HELP screen to show CREDIT (or Dynamic Credits) based on what lumps are found in the PWAD.

Order goes: PWAD HELP > PWAD CREDIT > DYNAMIC CREDIT

The reasoning for this is that for the longest time, PrBoom used CREDIT for the HELP screen erroneously... which resulted in many wads (i.e Scythe 2) including a CREDIT lump, but not a HELP lump.

Edit: I think it's worth having a discussion around automatically disabling the IWAD internal demos for PWADs (w/ maps)... As Nyan Doom has this functionality tied to a config option. I don't see any reason why anyone would want to see a desyncing "Circle of Death" demo for wads that don't include custom internal demos. We could even just have it enabled by default without a config option. Thoughts?

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -683,7 +683,7 @@ static void D_PageDrawer(void)
     V_DrawNamePatchFS(0, 0, 0, pagename, CR_DEFAULT, VPT_STRETCH);
   }
   else
-    M_DrawCredits();
+    M_DrawCreditsDynamic();
 }
 
 //
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -654,6 +654,18 @@ void D_PageTicker(void)
     D_AdvanceDemo();
 }
 
+// Check whether to skip IWAD Demos
+static int dsda_SkipIwadDemos(void)
+{
+  int pwaddemo = W_PWADLumpNameExists("DEMO1");
+  int pwadmaps = W_PWADLumpNameExists("THINGS");
+
+  if ((pwadmaps && !pwaddemo) || lumpinfo[W_CheckNumForName("DEMO1")].size == 0)
+    return true;
+  
+  return false;
+}
+
 //
 // D_PageDrawer
 //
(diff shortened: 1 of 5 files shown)


rfomin:
> Edit: I think it's worth having a discussion around automatically disabling the IWAD internal demos for PWADs (w/ maps)... 

I like the idea.

andrikpowell:
> > Edit: I think it's worth having a discussion around automatically disabling the IWAD internal demos for PWADs (w/ maps)...
> 
> I like the idea.

I've added Nyan's implementation in the latest commit: [7dadf59](https://github.com/kraflab/dsda-doom/pull/735/commits/7dadf59277d3083b10868ab5942cc77d74fd0079)
_(minus the config option - although I'm not even sure we need a config option for this)_

andrikpowell:
Simplified / changed my approach to integrating this. It's somewhat based on an implementation proposed by @Pedro-Beirao and removes the dummy demostates I had before. I guess my thought was if ID24 demoloop functionality was ever integrated in the future, it would make sense to use another demosequence... But I will admit this approach uses less code.

Originally I had the idea of just skipping the demo states, however pushing the demostates forward caused an crash regarding the optional DEMO4.

In my opinion it's better to only check for DEMO1, since if that's been replaced it's almost guaranteed the other demos will be as well. The logic is also much simpler to follow. In addition, I'm not a fan of just skipping DEMOS, as I'd like to match ZDoom behaviour, which just limits the cycle to TITLEPIC / CREDIT.

prboom2/src/d_main.c
```diff
@@ -654,6 +654,18 @@ void D_PageTicker(void)
     D_AdvanceDemo();
 }
 
+// Check whether to skip IWAD Demos
+static int dsda_SkipIwadDemos(void)
+{
+  int pwaddemo = W_PWADLumpNameExists("DEMO1");
+  int pwadmaps = W_PWADLumpNameExists("THINGS");
```
Pedro-Beirao: Maybe add a check for "TEXTMAP" for UDMF?

andrikpowell: Added this via a new function in latest commit.

PR #736, opened 2025-07-27
Intermission "Now Entering" Extended Pause

andrikpowell:
This is a pretty simple PR. Many people complain (including me) that for Doom 2 you can't read the mapname on "Now Entering" screen, like in some other ports like GZDoom.

This PR adds an extended pause (not in demos) that is also skippable via a keypress.

I also threw in another fix that's tangentially related, which allows the "Now Entering" to show when entering MAP31. Currently it shows nothing when entering MAP31 (you can see this in Eviternity II). This is because this code was supposed to avoid the "Now Entering" for after MAP30 (i.e. next map would be MAP31), but ended up also not showing "Now Entering" for any map entering MAP31.

### DIFF
--- a/prboom2/src/wi_stuff.c
+++ b/prboom2/src/wi_stuff.c
@@ -1169,8 +1169,7 @@ void WI_drawShowNextLoc(void)
     return; // MAP08 end game
 
   // draws which level you are entering..
-  if ( (gamemode != commercial)
-     || wbs->next != 30)  // check for MAP30 end game
+  if ( (gamemode != commercial) || (gamemap != 30) )  // allows for MAP31 entering screen, if not MAP30
   WI_drawEL();
 }
 
--- a/prboom2/src/wi_stuff.c
+++ b/prboom2/src/wi_stuff.c
@@ -35,6 +35,7 @@
 #include "doomstat.h"
 #include "m_random.h"
 #include "w_wad.h"
+#include "f_finale.h"
 #include "g_game.h"
 #include "r_main.h"
 #include "v_video.h"
@@ -968,6 +969,15 @@ void WI_End(void)
     WI_endStats();
 }
 
+#define WI_LONGER_NOW_ENTERING (allow_incompatibility && gamemode == commercial && !netgame)
+
+void WI_wait(void)
+{
+  if (WI_LONGER_NOW_ENTERING && !F_ShowCast())
+      cnt = TICRATE + 10;
+  else
+      cnt = 10;
+}
 
 // ====================================================================
 // WI_initNoState
@@ -979,7 +989,7 @@ void WI_initNoState(void)
 {
   state = NoState;
   acceleratestage = 0;
-  cnt = 10;
+  WI_wait();
 }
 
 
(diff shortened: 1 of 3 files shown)


rfomin:
~~This will break the compatibility for multi-level demos. We need to add checks for `demorecording` etc.~~

However, I don't think this change is appropriate in DSDA-Doom.

Edit: I didn't notice `allow_incompatibility`

prboom2/src/wi_stuff.c
```diff
@@ -968,6 +969,17 @@ void WI_End(void)
     WI_endStats();
 }
 
+#define WI_INTERMISSION_PAUSE (!dsda_StrictMode() && allow_incompatibility && gamemode == commercial && !netgame)
```
Pedro-Beirao: Why `!dsda_StrictMode()`? It should be like this in any demo, so `allow_incompatibility` is enough.
And maybe `WI_LONGER_NOW_ENTERING` is better?

prboom2/src/wi_stuff.c
```diff
@@ -968,6 +969,17 @@ void WI_End(void)
...
+
+void WI_wait(void)
+{
+  int final_map = (gamemap == 30 || (gamemission == pack_nerve && allow_incompatibility && gamemap == 8) || dsda_FinaleShortcut());
+
+  if (WI_INTERMISSION_PAUSE && !final_map)
```
Pedro-Beirao: `final_map` is the same as `F_ShowCast()`

```
extern dboolean F_ShowCast();
if (WI_INTERMISSION_PAUSE && !F_ShowCast())
```

I think this is easier to understand.

Pedro-Beirao:
I think this is a good change for dsda-doom. Ive definetely seen requests to make the `Now Entering` screen longer, and this is not a hard to mantain piece of code.

elf-alchemist:
I like this feature, would love to port this over to Woof once it's done :slightly_smiling_face:

rfomin:
> I like this feature, would love to port this over to Woof once it's done 🙂

We already have it for the ID24 intermissions.

prboom2/src/wi_stuff.c
andrikpowell: Ah you are correct, I'll use that then.

prboom2/src/wi_stuff.c
andrikpowell: Guess I was being extra careful... There's absolutely no way for `dsda_StrictMode()` to be true unless it's either demo recording / playback, right?

While I'm not super crazy on the wording of `WI_LONGER_NOW_ENTERING`, I'll admit it does do a better job of explaining it's purpose.

andrikpowell:
> > I like this feature, would love to port this over to Woof once it's done 🙂
> 
> We already have it for the ID24 intermissions.

Can you point me to this code? I'm curious on it's implementation.

rfomin:
> > > I like this feature, would love to port this over to Woof once it's done 🙂
> > 
> > We already have it for the ID24 intermissions.
> 
> Can you point me to this code? I'm curious on it's implementation.

See the calls of `NextLocAnimation()` function: https://github.com/fabiangreffrath/woof/blob/c8c723ea81b0623100fb24ee33ccf6a0acee4f1d/src/wi_stuff.c#L647
It's about switching into "Doom 1 mode" due to the nature of ID24 intermissions.

Pedro-Beirao (approved):

PR #744, opened 2025-08-01
CMake: actually apply -Wno-format-truncation.

FtZPetruska:
The check I added in 376fecc366676f4acf42c274ac64b141a90f304a incorrectly added the warning to the list of flags to check, and not to the list of warnings to be used.

Thanks to @andrikpowell for letting me know about the issue!


Pedro-Beirao (approved):
Looks good, thank you!

PR #745, opened 2025-08-01
CMake: Move packaging into its own subdirectory.

FtZPetruska:
This moves all the CPack-related scripts and logic into their own subdirectory.

There were also a few tweaks made to the handling of `External` scripts:
- Stop relying on `$<TARGET_FILE:...>`, configuring would fail when multi-config generators were used from the `file(GENERATE ...)` call. Instead of pulling in the executable from the build directory, pick the one from the staged install directory, since it is known to be at `<staging prefix>/${DSDA_INSTALL_BINDIR}`. This also lets us move away from generating the script entirely, instead just loading the correct script for the given platform.
- On Linux, support downloadinglinuxdeploy on `i386`, `armhf`, and `aarch64` hosts.
- On macOS, fix packaging when any of `DSDAPWADDIR`, `DSDA_INSTALL_BINDIR` or `DSDA_INSTALL_COPYRIGHT_DIR` has been changed. Using their value instead of assuming the default.


Pedro-Beirao (approved):
Looks good, thank you!
Very well organized

rfomin (approved):
Looks good, thank you!

PR #746, opened 2025-08-02
highlight non-countable pickup items on automap

vadosnaprimer:
regular sprite color switched to white because those sprites are not useful in gameplay. green that they were using is now the pickup color, because those are valuable in gameplay.

изображение

изображение

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -2488,15 +2488,22 @@ static void AM_drawThings(void)
         angle = 0x40000000;
       }
 
+      int color = mapcolor_p->sprt;
+
+      if (t->flags & MF_FRIEND && !t->player)
+        color = mapcolor_p->frnd;
+      /* cph 2006/07/30 - Show count-as-kills in red. */
+      else if ((t->flags & (MF_COUNTKILL | MF_CORPSE)) == MF_COUNTKILL)
+        color = mapcolor_p->enemy;
+      /* bbm 2/28/03 Show countable items in yellow. */
+      else if (t->flags & MF_COUNTITEM)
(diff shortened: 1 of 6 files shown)


prboom2/src/am_map.c
```diff
@@ -2488,15 +2488,22 @@ static void AM_drawThings(void)
         angle = 0x40000000;
       }
 
+      int color = mapcolor_p->sprt;
```
Pedro-Beirao: Variable definition should be at the start of the scope.

```
/home/runner/work/dsda-doom/dsda-doom/prboom2/src/am_map.c: In function ‘AM_drawThings’:
/home/runner/work/dsda-doom/dsda-doom/prboom2/src/am_map.c:2491:7: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
 2491 |       int color = mapcolor_p->sprt;
      |       ^~~
      ```

**Comment by Pedro-Beirao** [comment]
Maybe slightly grey instead of white? Since they are not useful in gameplay, they dont need to be this noticeable.

File: prboom2/src/am_map.c (hunk 1 above)
Comment (vadosnaprimer) [reply prboom2/src/am_map.c]: `color` is used for keys, what variable name would you suggest? Or maybe change the former?

Comment (Pedro-Beirao) [reply prboom2/src/am_map.c]: Its not the variable name, name it whatever. In C all variables need to be declared at the start of the scope. So, put it up there with
```
      mpoint_t p;
      angle_t angle;
      fixed_t scale;
```

vadosnaprimer:
Is this okay?

изображение

Pedro-Beirao (approved):
Thank you!

PR #748, opened 2025-08-02
Fix WeaponBob Demo Desync

### LINKED ISSUE #747: Changing weapon bobbing alters game behavior slightly, resulting in occasional demo desyncs
If you record a demo with weapon bobbing 100% enabled and then play it back with any other percentage, the demo may desync at certain spots. Not all demos desync, don't know the exact cause.
Conversely, if you record a demo with a weapon bobbing percentage other than 100%, playing back that demo with any other weapon bobbing percentage than what you recorded it with in the first place may cause the playback to desync.

Through testing, the issue first started came up in `v0.29.1`. This was also the first version to introduce the weapon bobbing percentage.

andrikpowell:
Fixes #747

Ugh was kinda hoping for the previous PR to be significantly reviewed last time, cuz I was iffy about the code... but regardless here is a fix.

In order to enable various WeaponBob increments, we need to always apply `R_ApplyWeaponBob` (I'll admit that DSDA's way of dealing with weaponbob confused me when I first made the PR).

So funny enough my chainsaw jittery fix is required to obey WeaponBob increments.

### DIFF
--- a/prboom2/src/p_pspr.c
+++ b/prboom2/src/p_pspr.c
@@ -52,7 +52,6 @@
 #include "dsda.h"
 #include "dsda/aim.h"
 #include "dsda/excmd.h"
-#include "dsda/settings.h"
 
 #define LOWERSPEED   (FRACUNIT*6)
 #define RAISESPEED   (FRACUNIT*6)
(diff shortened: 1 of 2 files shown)


andrikpowell:
Updated PR with a slightly less greedy approach... This means that it re-enables the jittery chainsaw when interpolation is disabled AND is using the default weaponbob.

There's no way to keep the jittery chainsaw with lower weaponbob %s.

Pedro-Beirao (approved):
Looks good. Thank you!
Obviously changing `psp` was desync prone, dont know how I missed it...

PR #749, opened 2025-08-04
CMake: Use PkgConfig::* targets in Find modules

FtZPetruska:
This updates how dependencies are searched using pkg-config.

Previously, the find modules were using the result variables of `pkg_check_modules` to search manually the libraries, include directories, and setup the target's properties. This would get fragile especially when transitive dependencies are involved, because `<prefix>_LIBRARIES` is only a list of library names.

This moves to using the `PkgConfig::<prefix>` imported target created when `pkg_check_modules` is given `IMPORTED_TARGET`. This target has all the necessary details already setup, so we can just make an alias of it and return early.

---

This change fixes a few things with the `SndFile` find module. It fixes static linking and provides the version info (used to define `HAVE_SNDFILE_MPEG`).


Pedro-Beirao (approved):
Looks fine, but tbh I dont know how to review this 😅 
Thank you!

rfomin (approved):
Thanks for the fixing `SndFile` module!
LGTM.

PR #750, opened 2025-08-05
Custom Skill Builder

andrikpowell:
@Pedro-Beirao asked me to port over Nyan's Custom Skill Builder functionality. It is based on Nugget Doom's implementation, but uses mapinfo skill properties (DSDA's original skills also use this).

In order to get the Custom Skill working correctly, a few changes to the menu system and a NOMONSTERS skill flag were added.

I'm sure there will be quite a bit of questions and thoughts on this, since it's a pretty big PR.


vadosnaprimer:
~~I want to test it but can't figure out how to access the custom skill menus (or the feature itself through the menus).~~ Oh it's at the start of the game!

vadosnaprimer (approved):
Works great, I love it!!!

doom00

I'm using co-op spawns to maximize fun, but with ITYTD HP and ammo, and BTSX finally feels welcoming!

vadosnaprimer:
Found a bug.

- Start playing with a custom skill and make a save. Getting to other levels within that session will make the skill persist.
- Exit the game, launch it again and load your save. The level it's saved in will have the correct skill. But getting to other levels will reset your custom skill.

I experienced this when beating levels normally across several sessions, but it behaves the same if you simply hit the Next Level hotkey (Page Down) before and after restart.

andrikpowell:
> Found a bug.
> 
> * Start playing with a custom skill and make a save. Getting to other levels within that session will make the skill persist.
> * Exit the game, launch it again and load your save. The level it's saved in will have the correct skill. But getting to other levels will reset your custom skill.
> 
> I experienced this when beating levels normally across several sessions, but it behaves the same if you simply hit the Next Level hotkey (Page Down) before and after restart.

Hmm yeah, seems that when loading a game, I need to update the "custom skill" parameters.

You can easily test this by typing the SKILL cheat on loading a custom skill game (during new game session), and you'll see "skill 6 - (null)".

I'll have to think on the best way to update that.

andrikpowell:
Alright so I think I've found a solution.
I now store the custom skill into the saved game.

However I've added a condition (which I think will be fine), that only saves it in the saved game if it's a custom skill saved game. This way saves/loads made under the normal skills still work... Plus imo we don't need to store custom skill info for normal skills anyway.

vadosnaprimer:
Thanks, it works now.

PR #752, opened 2025-08-05
Add support for non-power-of-2 textures

elf-alchemist:
Test WAD: [clamptest.zip](https://github.com/user-attachments/files/21604338/clamptest.zip)
(missing floor texture is an unrelated feature)

Software, on `master`:
doom29

Software, this branch:
doom30

OpenGL, either branch:
doom31

### DIFF
--- a/prboom2/src/r_data.c
+++ b/prboom2/src/r_data.c
@@ -106,8 +106,8 @@ int       *texturetranslation;
 //
 
 const byte *R_GetTextureColumn(const rpatch_t *texpatch, int col) {
-  const uint width = texpatch->width;
-  const uint mask = texpatch->widthmask;
+  const int width = texpatch->width;
+  const unsigned int mask = texpatch->widthmask;
 
   while (col < 0)
     col += width;
--- a/prboom2/src/r_data.c
+++ b/prboom2/src/r_data.c
@@ -106,9 +106,16 @@ int       *texturetranslation;
 //
 
 const byte *R_GetTextureColumn(const rpatch_t *texpatch, int col) {
+  const uint width = texpatch->width;
+  const uint mask = texpatch->widthmask;
+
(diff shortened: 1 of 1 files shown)


andrikpowell:
I'm not crazy about allowing this in Vanilla complevels... Mostly because what will end up happening is that someone will make a Vanilla wad, expect textures to act a certain way, and then will get blindsided when someone complains that the textures look wrong when playing back in Vanilla.

It's probably fine for Boom and up (even though original Boom and MBF probably had this restiction)

elf-alchemist:
I must emphasize an important point, that DSDA (and basically any port that isn't Chocolate, CRL, etc) already fix the vast majority of vanilla renderring bugs and static limits. And if someone is testing a WAD that is meant to target the DOS EXE solely on DSDA, then a few textures being clamped down to mext lowest power-of-2 will be the least of the issues they'll encounter 😛

andrikpowell:
I sorta retract my statement above... DSDA-Doom is currently not following the Vanilla vertical tiling behaviour under vanilla complevels... So I _guess_ it doesn't matter if we make a change to the horizontal tiling.

I will say that if I was mapping for Vanilla and was using a "256 tall by 368px wide" texture (with this PR fix integrated) in my wad, I would get pretty annoyed if I had to learn later on after testing my map in DSDA-Doom that in Choco/Vanilla it did not tile the textures of my map in the same way.

Of course, this is less of an issue for mappers that understand how Vanilla textures work... But still... NinjaDelphox ran into this issue slightly when working on 150 Line Massacre, as he didn't realise that 256 tall textures will only vertically tile by the first 128 pixels in Vanilla, but DSDA-Doom tiled it via 256 pixels.

Pedro-Beirao (approved):
Thank you!

PR #755, opened 2025-08-05
Use demo_tics instead of playback_tics

### LINKED ISSUE #753: Total time in HUD is not counted in demos where player restarts the level
In games where the player retries after dying, the brown time counter appears on the HUD (screenblocks=11), timing the total time spent on that level, including all deaths, with the green one representing the time spent on the current attempt.

A demo of said game will have the brown counter appear after the first death, but it will never go above 0:00.

Pedro-Beirao:
Fixes https://github.com/kraflab/dsda-doom/issues/753

### DIFF
--- a/prboom2/src/dsda/demo.c
+++ b/prboom2/src/dsda/demo.c
@@ -46,13 +46,14 @@
 
 #define INITIAL_DEMO_BUFFER_SIZE 0x20000
 
+int demo_tics;
+
 static char* dsda_demo_name_base;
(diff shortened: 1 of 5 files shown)


0n-s:
This resolves the HUD issue, but the -export_text_file issue is still there.

Although, the demos I'm using here only contain failed attempts. If this is actually intentional behavior, then please disregard this.

Pedro-Beirao:
-export_text_file will return the time of the last level exit. If you dont exit any level then it will be 0

PR #756, opened 2025-08-05
Warp mouse to the middle of the screen

### LINKED ISSUE #754: Fedora / GNOME issue
Under Fedora with default GNOME desktop, the top left hot corner will cause any mouse buttons to not be registered.

To duplicate: while playing move the mouse to the upper left and try shooting 

Workaround: a different desktop environment might fix it, but also you can run it in windowed mode. This is not ideal but it works.

Pedro-Beirao:
Fixes https://github.com/kraflab/dsda-doom/issues/754

> SDL_WarpMouseInWindow()
This function generates a mouse motion event if relative mode is not enabled. If relative mode is enabled, you can force mouse events for the warp by setting the [SDL_HINT_MOUSE_RELATIVE_WARP_MOTION](https://wiki.libsdl.org/SDL2/SDL_HINT_MOUSE_RELATIVE_WARP_MOTION) hint.

Relative mode is enabled when the mouse is grabbed, so Im guessing this is fine.

### DIFF
--- a/prboom2/src/SDL/i_sshot.c
+++ b/prboom2/src/SDL/i_sshot.c
@@ -59,8 +59,8 @@ int renderH;
 
 void I_UpdateRenderSize(void)
 {
-  renderW = window_rect.w;
-  renderH = window_rect.h;
+  renderW = renderer_rect.w;
+  renderH = renderer_rect.h;
 }
 
 //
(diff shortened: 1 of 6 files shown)


Pedro-Beirao:
Latest commit fixes a bug reported on discord https://discord.com/channels/218106532406099969/776233266411339786/1401869827790082170

PR #757, opened 2025-08-06
Fix wrong path in Linux cpack

Pedro-Beirao:
Issue introduced in https://github.com/kraflab/dsda-doom/pull/745 but I missed it since the CI didnt output any error

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -191,6 +191,7 @@ jobs:
         with:
           name: ${{ matrix.config.package_name }}
           path: ${{ matrix.config.artifact-path }}
+          if-no-files-found: error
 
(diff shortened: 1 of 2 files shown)


Pedro-Beirao:
No problem on setting the retention days to the max right? I dont have access to the settings of the repo, but it feels like it is set to like 5 days.

rfomin:
> No problem on setting the retention days to the max right? I dont have access to the settings of the repo, but it feels like it is set to like 5 days.

Yes, it's okay.

Pedro-Beirao:
> Warning: Retention days cannot be greater than the maximum allowed retention set within the repository. Using 1 instead.

Alright... nvm.
Ill remove that commit tmrw, and ask kraf to change it instead

PR #761, opened 2025-08-11
Specify Homebrew dependencies with a Brewfile

FtZPetruska:
This PR adds a `Brewfile` at the top of the repository to handle Homebrew dependencies more conveniently.

This has the advantage of:
- Having all the macOS dependencies listed in a single location, instead of having it duplicated in the build guide and CI.
- Installing everything in a single command without listing every packages (`brew bundle`).
- For CI, this fixes warnings about trying to install the dependencies already pre-installed on the runner.

---

I have also updated the macOS build guide, so it recommends `brew bundle` to install dependencies. Also took the opportunity to update the option name since `ENABLE_LTO` was removed, and to recommend using `cpack` to create a package instead of manually baking a release.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -121,24 +121,7 @@ jobs:
 
       - name: Install Dependencies (macOS)
         if: runner.os == 'macOS'
-        run: |
-          brew update
-          env HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1 \
-            brew install --overwrite \
-              python \
-              ninja \
-              pkg-config \
-              libxmp \
-              fluid-synth \
-              libvorbis \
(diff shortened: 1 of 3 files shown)


Pedro-Beirao (approved):
All very good thank you!
I didnt even know `brew bundle` was a thing. Never seen any other project using it.

FtZPetruska:
> I didnt even know `brew bundle` was a thing. Never seen any other project using it.

I only stumbled upon it recently. I saw projects such as [vorbis](https://gitlab.xiph.org/xiph/vorbis/-/blob/master/Brewfile?ref_type=heads) or [opusfile](https://github.com/xiph/opusfile/blob/main/Brewfile) use it, read about it and thought it would be a nice fit to handle macOS dependencies.

PR #763, opened 2025-08-12
CMake: Use custom package name

FtZPetruska:
When refactoring packaging with CPack in #745, I wrongly assumed that the default names would be enough. With the default name, the most glaring issue is that both x64 and arm64 packages for macOS have the same name, meaning when a release is made, whichever finishes last overwrites the first.

This PR manually sets the `CPACK_PACKAGE_FILE_NAME` to `dsda-doom-<version>-<platform>-<arch>`.

The arch part has a few caveats that need to be worked around:
- In most cases, `CMAKE_SYSTEM_PROCESSOR` is enough.
- In Visual Studio, when targeting another architecture (e.g. ARM64), `CMAKE_SYSTEM_PROCESSOR` holds the architecture of the host. Instead `CMAKE_VS_PLATFORM_NAME` holds the target architecture.
- On macOS, the target architecture(s) may be set in `CMAKE_OSX_ARCHITECTURES`, similar to Visual Studio, `CMAKE_SYSTEM_PROCESSOR` still refers to the host.

The platform name is also manually defined. The default `CPACK_PACKAGE_FILE_NAME` uses `CPACK_SYSTEM_NAME`, but this has the following caveats:
- On macOS, it evaluates to `Darwin` which may not be obvious to some user that this is meant for mac.
- On Windows, it evaluates to `win32` when targeting x86 and `win64` when targeting any other architecture, including arm.

After this change, here is what release artifacts would look like: https://github.com/FtZPetruska/dsda-doom/releases/tag/custom-package-name-test-release


Pedro-Beirao (approved):
Look good, thank you!
But now I will have to update my [windows script](https://github.com/Pedro-Beirao/dsda-launcher/blob/master/updater/dsda-updater-windows.bat) since it expects a different name😄
(ik it looks terrible with the powershell calls, but I dont usually write scripts for windows, so whatever)

PR #764, opened 2025-08-12
CMake: Always copy pwad next to the executable

FtZPetruska:
Copying the pwad next to the executable is not only needed on MSVC. More generally, all multi-config generators (Visual Studio, Xcode, Ninja Multi-Config) need this `POST_BUILD` command for the executable to work from the build directory without manually specifying the pwad.

On single-config platform the command is a no-op, as such it can be run unconditionally.

### DIFF
--- a/prboom2/src/CMakeLists.txt
+++ b/prboom2/src/CMakeLists.txt
@@ -553,6 +553,11 @@ function(AddGameExecutable TARGET SOURCES)
     )
 
     add_dependencies(${TARGET} dsda-doom-wad)
+    add_custom_command(TARGET ${TARGET}
+        POST_BUILD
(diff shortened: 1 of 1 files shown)


Pedro-Beirao (approved):
I dont have a way to test in multi-config generators, but this cant hurt, so sure. Thanks!

FtZPetruska:
> I dont have a way to test in multi-config generators, but this cant hurt, so sure. Thanks!

Ninja provides a multi-config generator on all platforms, you only need to configure with `"-GNinja Multi-Config"`!

Pedro-Beirao:
Thanks for letting me know, this is pretty cool! And now I see the issue this fixes, thanks!

PR #766, opened 2025-08-16
CMake: Avoid CMAKE_FIND_REQUIRED

FtZPetruska:
When using vcpkg, the toolchain wraps `find_package`. Some packages (notably `zlib` in this case) use this to insert an additional script before calling CMake's `find_package`. Enabling `CMAKE_FIND_REQUIRED` may lead to an error under vcpkg's `*-release` triplets (see https://github.com/microsoft/vcpkg/issues/46924).

So, let's just always use the keyword approach.


Pedro-Beirao (approved):
Thank you!

PR #774, opened 2025-09-24
Fix "Pistol Start" Temporary Modifier to Actually Update

andrikpowell:
A fix to #724

Pistol start wouldn't update when changing the menu option, because of some incorrect logic (on my part).
- Previous logic only updated pistolstart when there was a conflict between "Pistol Start" and "Always Pistol Start"

To be honest, pretty annoyed as this faulty behaviour has been in Nyan for 1-2 months now.


Pedro-Beirao (approved):
Thank you!

PR #780, opened 2025-10-12
Minor tweaks to the codebase

AbitTheGray:
There are many changes that I've bundled into a single MR. Please say which changes you want to keep and I'll cherrypick them into dedicated MR(s).

- First commit (eb5b67b) are many changes (see text of the commit), but those are the main changes I suggest merging.
- Second commit (96d93d9) just replaces `35` in all the places with `TICRATE`. I did not try changing value of `TICRATE` before and after the change as I don't see it as a constant that is meant to be tweaked but more like naming the "magic number" to make the code more readable
- Third commit (01f6cfc) unifies line indentation style in several files that had multiple styles in them. I recommend merging this one, especially for UDMF where 1st entry used one indent and the rest used a different one.

There are two more changes I've considered:
- `const char * comp_lev_str[MAX_COMPATIBILITY_LEVEL]` is checked as `if (sizeof(comp_lev_str)/sizeof(comp_lev_str[0]) != MAX_COMPATIBILITY_LEVEL)` (both in `g_game.c`) which should never happen and would result in `I_Error("G_ReadDemoHeader: compatibility level strings incomplete");`
- `cmd->angleturn = ((unsigned char)(at = *(*data_p)++))<<8;` in `g_game.c` where `unsigned char` is shifted left by 8, which is the same size as `unsigned char` - as I understand it, it is auto-promoted to `unsigned int` before the shift but I think it should be cast to it instead of casting it to type it already is (`at` is already `unsigned char`).

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -736,7 +736,7 @@ void D_SetPage(const char* name, int tics, int music)
 
 static void D_DrawTitle1(const char *name)
 {
-  D_SetPage(name, TICRATE * 170 / 35, mus_intro);
+  D_SetPage(name, TICRATE * 170 / 35, mus_intro); // 35 * 170 / 35, is that correct math?
 }
 
 static void D_DrawTitle2(const char *name)
--- a/prboom2/src/dsda/endoom.c
+++ b/prboom2/src/dsda/endoom.c
@@ -193,7 +193,7 @@ static const char* cp437_to_utf8[256] = {
   "\xc3\xb9",
   "\xc3\xbf",
   "\xc3\x96",
-  "\xc3\x9c	",
+  "\xc3\x9c\t",
   "\xc2\xa2",
   "\xc2\xa3",
   "\xc2\xa5",
--- a/prboom2/src/MUSIC/flplayer.c
+++ b/prboom2/src/MUSIC/flplayer.c
@@ -438,8 +438,8 @@ static void fl_writesamples_ex (short *dest, int nsamp)
   if (nsamp * 2 > fbuff_siz)
   {
     float *newfbuff = (float*)realloc (fbuff, nsamp * 2 * sizeof (float));
-	if (!newfbuff) return;
-	fbuff = newfbuff;
+    if (!newfbuff) return;
+    fbuff = newfbuff;
     fbuff_siz = nsamp * 2;
   }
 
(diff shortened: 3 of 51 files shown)


prboom2/src/dsda/endoom.c
```diff
@@ -193,7 +193,7 @@ static const char* cp437_to_utf8[256] = {
   "\xc3\xb9",
   "\xc3\xbf",
   "\xc3\x96",
-  "\xc3\x9c	",
```
AbitTheGray, line 196: This surprised me. It is a tab but without visible whitespace characters, it looks like a space. Found it only while looking for `\t` indentation. Did not check whenever or not it makes sense there, but there are other strings with 3 characters in the array, just no tab.

thecnoNSMB: This looks like it's in a conversion table between two binary text formats, so it's probably supposed to be there, and definitely shouldn't be changed without understanding the data thoroughly.

thecnoNSMB:
~~I'm not a code reviewer, but I've noticed several places where your whitespace changes break alignment of similar elements in the code. Did you "find-replace" tabs with spaces in the codebase without reviewing the changes yourself?~~ Nevermind, I'd read the diff backwards lol

prboom2/src/dsda/endoom.c
AbitTheGray: Changing it to explicit `\t` does not change any behavior, just makes it easier to recognize for the character it is.

thecnoNSMB: Ah, that's totally reasonable, then. I'd apparently misunderstood the diff.

AbitTheGray:
Thanks to @thecnoNSMB , I've double-checked the indentations and there was one find&replace error that I did not notice in `gamemodes` of `DSDADoom_misc.cfg`, I've fixed it. Even while I prefer tabs for indentation, I don't think they should appear anywhere else than at the start of the line as they usually break the formatting when you change the width when they are in the text (like there were, by my mistake).

prboom2/src/d_main.c
```diff
@@ -736,7 +736,7 @@ void D_SetPage(const char* name, int tics, int music)
 
 static void D_DrawTitle1(const char *name)
 {
-  D_SetPage(name, TICRATE * 170 / 35, mus_intro);
+  D_SetPage(name, TICRATE * 170 / 35, mus_intro); // 35 * 170 / 35, is that correct math?
```
Pedro-Beirao, line 739: You can make this just be 170 ig

PR #789, opened 2025-10-25
add angle indication for hitbox thing appearance

vadosnaprimer:
Forgot this was possible when [replacing triangles with squares](https://github.com/kraflab/dsda-doom/pull/688). Got reminded about it by [besus](https://github.com/complevel9/susdoom).

> //  also it looks like doom builder

изображение

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -292,6 +292,12 @@ mline_t thingbox_guy[] =
 #undef R
 #define NUMTHINGBOXGUYLINES (sizeof(thingbox_guy)/sizeof(mline_t))
 
+#define R (FRACUNIT)
+mline_t arrowhead_guy[] = {{{(fixed_t)0, (fixed_t)(-.5 * R)}, {(fixed_t)(.75 * R), (fixed_t)0}},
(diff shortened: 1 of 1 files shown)


Pedro-Beirao:
Dont know if I like how this looks. Any other opinions or suggestions?

andrikpowell:
> Dont know if I like how this looks. Any other opinions or suggestions?

I suggest drawing them like how Nugget Doom (and now Nyan Doom) draws them... Meaning that the bounding box is drawn independently from the actual thing...

nyan-hitboxes-1

This means it's possible to have either scaled / classic thing sizes, while keeping the bounding box the actual size of the thing. Also it allows customisation of the bounding box colour itself (I'm not a fan of the bounding box being the same colour as the symbol).

nyan-hitboxes-2

Pedro-Beirao:
> I suggest drawing them like how Nugget Doom (and now Nyan Doom) draws them... Meaning that the bounding box is drawn independently from the actual thing...

Yea thats better and looks really cool

vadosnaprimer:
I'll make it a new PR then.

PR #791, opened 2025-10-29
Add clangd cache folder to gitignore

elf-alchemist:
Clangd generates cached indexing files under this sub directory, they very easily get in the way of any Git operation, cluttering the list of untracked files.


Pedro-Beirao (approved):
I have this in my local gitignore, so alright for me

rfomin (approved):
Yes, I use it too.

PR #792, opened 2025-10-30
Fix default line id on UDMF maps

elf-alchemist:
An oversight on my end. Corrected as per the official spec. Also updated official docs.


Pedro-Beirao (approved):
Thank you!

PR #794, opened 2025-11-04
Add ZDoom "Line Horizon" special

elf-alchemist:
Commented out ZDoom line specials that aren't supported yet (or never will), this makes it dead easy to track what is actually missing.

[line_horizon_test.zip](https://github.com/user-attachments/files/23326581/line_horizon_test.zip)

Tested on Odamex, UZDoom -- code from both were used as reference.

(There's a thin line HOM in the distant horizon, but that's just a consequence of not having a sky flat.)
doom107


Pedro-Beirao:
Thank you! 🫡

PR #799, opened 2025-11-07
Minimap Coordinates Fix

### LINKED ISSUE #574: Minimap Does Not Refresh Coordinates when Changing Resolution
How to test:
1. Make sure "Enable Minimap" is turned on under "Automap Settings"
2. Change the resolution to anything else.
3. You will notice that the minimap does not seem to adjust coordinates (to be in the same-ish position) based on the resolution change.

In the code, I'd guess `dsda_CopyMinimapCoordinates` is part of addressing this, but obviously something is missing... I wasn't able to figure out why the minimap would change place on the screen.

Software always has the minimap visible on the screen, but will jump to different parts of the screen (instead of the default top-right). However in OpenGL, the minimap seems to disappear. OpenGL must deal with the coordinates differently than Software.

I've also found that if you change the resoultion, quit, and then relaunch DSDA Doom, the minimap seems to pull the correct minimap coordinates for that resolution.

That's why my main hunch is that when changing resolutions, the minimap coordinates aren't getting updated.

andrikpowell:
Fixes #574 
See that issue for more detail.

Basically the minimap coordinates set in DSDAHUD would only be used once, and then be overwritten. This caused a problem whenever you changed your resolution, as it didn't have access to the original values.

This actually could cause a crash if you had the minimap enabled and went from 2560x1440 to a very small resolution like 640x400, as it would result in the minimap coordinates going out-of-bounds.

This PR fixes that.

Note there is an issue when changing resolution where the minimap doesn't center around the player... This was an issue before, and should eventually be fixed... although I haven't found a solution quite yet. but you can fix this pretty easily by just entering and exiting the main automap.

Also I added the second commit, because imo it doesn't make sense for the minimap to not follow the player.

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -2866,7 +2866,7 @@ void AM_Drawer (dboolean minimap)
 
   V_BeginAutomapDraw();
 
-  if (automap_follow)
+  if (automap_follow || minimap)
     AM_doFollowPlayer();
 
   // Change the zoom if necessary
--- a/prboom2/src/dsda/hud_components/minimap.c
+++ b/prboom2/src/dsda/hud_components/minimap.c
@@ -25,15 +25,21 @@ typedef struct {
 
 static local_component_t* local;
 
-void dsda_InitMinimapHC(int x_offset, int y_offset, int vpt, int* args, int arg_count, void** data) {
-  *data = Z_Calloc(1, sizeof(local_component_t));
-  local = *data;
+// backups of original values
+// (so we can access them later)
+static int xx, yy, ww, hh;
+static int yy_offset, flags;
 
-  local->x = x_offset;
-  local->y = dsda_HudComponentY(y_offset, vpt, 0);
-  local->width = args[0];
-  local->height = args[1];
-  local->scale = args[2];
+static void dsda_UpdateMinimapCoordinates(void) {
+
+  // Must recalculate y each update
+  yy = dsda_HudComponentY(yy_offset, flags, 0);
+
+  // Varables
+  local->x      = xx;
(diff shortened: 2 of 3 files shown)


Pedro-Beirao:
Besides not centering around the player, opengl also breaks, especially with textured automap on.

Screenshot 2026-02-17 at 00 03 30

andrikpowell:
> Besides not centering around the player, opengl also breaks, especially with textured automap on.
> 
> Screenshot 2026-02-17 at 00 03 30

Can you go thru all the steps that you did?

Note that I am aware of an issue of the player not centering in the automap when changing the screensize / renderer... That issue is unrelated to this PR.

The main goal of this PR is to fix the minimap boundaries... Previously there was an issue that could actually cause a crash when switching from a high to low resolution, in that the minimap boundary coordinates did not update themselves when the resolution changed... basically if you set the minimap for 640x480, and then went up to 1920x1080, the minimap would be in the centre of the screen, since the coordinates were mutated, and the original coordinates from DSDAHUD would be lost.

It's kinda hard to tell what's happening in that screenshot... I haven't noticed an issue with the textured automap on my end... that's why it'd be very useful to know exactly what steps you did, so I can try and recreate it...

Also on my end, if I hit the automap button (tab), and then come back to the main screen, it usually fixes the centring.

Pedro-Beirao:
Oh seems like the problem is with rotating automap, and it is also fixed by opening/closing the automap

andrikpowell:
> Oh seems like the problem is with rotating automap, and it is also fixed by opening/closing the automap

The issue should be fixed with the latest commit :)

Pedro-Beirao:
Thank you!

PR #801, opened 2025-11-09
Allow Doom-in-Hexen format maps to use Lightning

andrikpowell:
So this is a very very early draft...

Need y'all to look into this. Technically this does work, but we may wanna structure this better.

Requires Doom-in-Hexen map format and MAPINFO with the line "Lightning".
Here's a very basic demo map: [lightning-doom-test-map.wad](https://drive.google.com/file/d/1FW54JznWt_x0utBEcWLV8izXva55gBGp/view?usp=sharing)

Currently the lightning seems to work, but it's currently using `pr_hexen` for randomness (not sure if this is good or not).
Also note that I've disabled the sound currently for Doom.. Not sure how we are gonna deal with that, since currently DSDA only allows sounds in the `sfxenum_t` list.

There is a MAPINFO field called "LightningSound" but it probably uses names from a lump like SNDINFO. ZDoom tends to use "World/Thunder" as the default value. I just don't think DSDA is currently setup for this functionality quite yet.

### DIFF
--- a/prboom2/src/dsda/mapinfo/doom.c
+++ b/prboom2/src/dsda/mapinfo/doom.c
@@ -702,7 +702,12 @@ int dsda_DoomPrepareFinished(void) {
 }
 
 int dsda_DoomMapLightning(int* lightning) {
-  return false;
+  if (raven || !current_map || !current_map->lightning)
+    return false;
+
+  *lightning = current_map->lightning;
+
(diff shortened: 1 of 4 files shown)


andrikpowell:
Doom MapInfo is gone.... No point in this pr anymore.

PR #802, opened 2025-11-11
Remove forced ExCmdDemo recording on DSDA-UDMF/MapInfo wads

elf-alchemist:
Somewhat experimental change that has consequences for demo recording.

This _should_ completely fix demo recording for Eviternity II, without needing the mapinfo patch -- but also has the major fix of enabling normal demos (non-ExCMdDemo) for DSDA-UDMF/ZDoom maps. Thus allowing valid recording for DSDA-UDMF map sets, like NoSp4, as ExCmdDemos always fail analysis testing, being classified as "Other".

### DIFF
--- a/prboom2/src/dsda/demo.c
+++ b/prboom2/src/dsda/demo.c
@@ -794,28 +794,9 @@ void dsda_WriteDSDADemoHeader(byte** p) {
 }
 
 void dsda_ApplyDSDADemoFormat(byte** demo_p) {
-  dboolean use_dsda_format = false;
-
-  if (map_format.zdoom)
-  {
-    if (!mbf21)
-      I_Error("You must use complevel 21 when recording in advanced formats.");
-
-    use_dsda_format = true;
-  }
-  else if (dsda_UseMapinfo())
-  {
-    use_dsda_format = true;
-  }
-
   if (dsda_Flag(dsda_arg_dsdademo))
   {
-    use_dsda_format = true;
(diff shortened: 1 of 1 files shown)


rfomin:
These new demos will not be compatible with other ports or old versions of DSDA-Doom. We will need a new complevel for UDMF/MAPINFO demos.
If we add a new complevel, there is no reason to remove ExCmdDemo; analysis should be updated to support the new complevel.

elf-alchemist:
I would not use the term conplevel, here, but if it just means putting these advanced, ZDoom-originated, features behind a user-provided CLI parameter or the like, I wouldn't object to it. But the current state of demo recording for these kinds of maps is extremely sub-bar for end users, it needs some important revision, and most solutions I can think of are tricky or include some major compromise (I may elaborate further, later). I will update this PR in a bit, I've noticed an issue with how it's currently laid out.

rfomin:
> I will update this PR in a bit, I've noticed an issue with how it's currently laid out.

The current approach won't work — these new demos won't be compatible. This defeats the purpose of a demo-compatible port.
We will need to add a new complevel (`demo_version`), after which we can make any changes or improvements we want.

elf-alchemist:
Will rethink and rework this, the thought of players being able to demo DSDA-UDMF has been haunting me for a while, but I'd like to look at maybe porting the ZDoom-y/Hexen-y mapping features to other ports first, and have a few maps tested to see if there's any major cross-port discrepancies first.

PR #808, opened 2025-11-19
Try opening the music lump with the XMP player first

### LINKED ISSUE #807: [Bug] IT tracker incompatibility
From discord user "valhen_saipiam":

> hiiii i recently came across a weird issue regarding dsda and its mod compatibility, so i've got this one song ("=Deeper Into The Sun=") that i wanted to use in a wad, yet for some reason it doesnt seem to play at all. what's interesting too is that i tried on woof 15.3.0 which apparently has the same library for mods as dsda does and yet it seems to work there. here's the song + wad btw

From grapes:

> I changed it to an XM and it works, if that helps diagnose it

[IncompatTracker.zip](https://github.com/user-attachments/files/23616368/IncompatTracker.zip)

rfomin:
There is no magic number or simple identification method for MP3.

Fix #807

### DIFF
--- a/prboom2/src/SDL/i_sound.c
+++ b/prboom2/src/SDL/i_sound.c
@@ -1058,9 +1058,9 @@ static int music_player_was_init[NUM_MUS_PLAYERS];
 // order in which players are to be tried
 char music_player_order[NUM_MUS_PLAYERS][200] =
 {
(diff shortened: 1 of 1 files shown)


Pedro-Beirao:
Makes sense, thank you!

PR #812, opened 2025-11-22
Fix "music.restart" console command for tracks defined in UMAPINFO

### LINKED ISSUE #811: Restarting UMAPINFO-defined music breaks it
Example WAD (map 33): https://www.doomworld.com/idgames/?id=21015

If this map's music is restarted with `music.restart` console command, it goes silent until you reload the level.

rfomin:
Fix #811 

I'm not sure about this, it needs testing.

### DIFF
--- a/prboom2/src/s_sound.c
+++ b/prboom2/src/s_sound.c
@@ -674,9 +674,6 @@ void S_ChangeMusic(int musicnum, int looping)
 
   music = &S_music[musicnum];
 
-  if (mus_playing == music)
-    return;
-
   // shutdown old music
   S_StopMusic();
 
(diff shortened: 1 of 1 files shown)


prboom2/src/s_sound.c
```diff
@@ -674,9 +674,6 @@ void S_ChangeMusic(int musicnum, int looping)
 
   music = &S_music[musicnum];
 
-  if (mus_playing == music)
-    return;
-
```
Pedro-Beirao, line 679: Think we shouldnt remove this one. Might not affect anything, but it doesnt hurt.

In `S_ChangeMusInfoMusic` we already check if `mus_playing` has the same `lumpnum`, so it is safe to remove the return there.

fabiangreffrath: > Think we shouldnt remove this one. 
> [...]
> so it is safe to remove the return there.

He? 😉

rfomin: Hmm, so it's ok?

PR #814, opened 2025-11-30
Remove pointeless carraige return

elf-alchemist:


rfomin:
It's pointless on Linux, but not on Windows.

elf-alchemist:
This is the singular instance of a text print using \r\n in the entire DSDA codebase, literally everywhere else is already using just a \n. I was wondering why my levelstat.txt `diff`s were not working between DSDA and Woof, and this was the sole exclusive reason :V

Windows has never had any issues in _any_ situation handling just a single \n in my experience. This is a problem solely on DSDA's end :v

It very much _is_ pointless.

rfomin:
> I was wondering why my levelstat.txt `diff`s were not working between DSDA and Woof, and this was the sole exclusive reason :V

Woof has a much simpler levelstat printing function; there are more differences than just the `\r`. The standard `diff -w` (ignores whitespace) should work; that's what we use in the `woof/demotest` script.

I don't think we should change this, it may break the parsing in people's tools.

Pedro-Beirao:
I would like this changed, but breaking people's tools is a valid concern.
Maybe directly warning ZM and 4shock and mentioning this in the next changelog is enough

andrey-budko:
> Windows has never had any issues in _any_ situation handling just a single \n in my experience. This is a problem solely on DSDA's end :v

Only since 2018
https://devblogs.microsoft.com/commandline/extended-eol-in-notepad/

Pedro-Beirao:
ZM and 4shock confirmed that this change wont affect their scripts, but no point in merging this anyway right?

PR #816, opened 2025-11-30
Allow Translucency to be Customisable for Boom and Higher

andrikpowell:
Strict Mode no longer restricts the ability to change the boom_translucent_sprites config.

Only thing that is restricted now is translucency under Vanilla complevels. While you can still change the setting, sprites will never be translucent in Vanilla under Strict Mode.


Pedro-Beirao (approved):
Looks good

PR #819, opened 2025-12-07
Fix monochrome menu options in OpenGL

Adrad1234:
Fixes colors in the automap menu being monochrome, when invulnerable
before_c
after_c


Pedro-Beirao (approved):
Look good, thanks!

PR #820, opened 2025-12-14
Add `comp_thingsectorlight` property

elf-alchemist:
`Alongside the fake heights transfer (242), Boom also introduced two separate line specials that individually transfer floor and ceiling light levels (213 and 261) onto a tagged sector. Normally, in Boom and CL9, any Things present in said sector will be lit according to the sector's normal light level. MBF, however, changed this behavior to decide a Thing's light level using the average of the transferred floor and ceiling light levels, instead.`

Not only is the MBF behavior incredibly unpopular among mappers, the choice of which behavior is also widely inconsistent among ports. Some ports will only use the Boom behavior of defining a Thing's light level based on the true sector, some will only use the MBF behavior and others will only read the transferred floor light level, instead. This proposition formally defines the behavior in question as part of MBF21.

Relevant:
https://github.com/doom-cross-port-collab/issue-tracker/issues/1
https://github.com/kraflab/mbf21/pull/6
https://github.com/fabiangreffrath/woof/pull/2393
https://github.com/team-eternity/eternity/pull/740
https://github.com/odamex/odamex/pull/1499

### DIFF
--- a/prboom2/src/doomstat.c
+++ b/prboom2/src/doomstat.c
@@ -57,7 +57,7 @@ complevel_t compatibility_level;
 // it's required for demos recorded in "demo compatibility" mode by boom201 for example
 int demover;
 
-int comp[MBF_COMP_TOTAL];    // killough 10/98
+int comp[COMP_TOTAL];    // killough 10/98
 int default_comperr[COMPERR_NUM];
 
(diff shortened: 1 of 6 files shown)


Pedro-Beirao (approved):
Thanks!

PR #827, opened 2026-01-03
let IDDKT cheat skip spawned/ressurected monsters

### LINKED ISSUE #826: "iddkt" cheat points to spawned monsters when it probably shouldn't
Specifically, it points to monsters spawned by monster spawners used by the Icon of Sin and other final bosses, which are not counted as kills by the extended HUD or the level table, and are therefore probably not what a player entering this cheat is trying to look for. This can cause problems using the cheat on maps with a monster spawner that activates when there's still a lot of existing monsters to kill.

fabiangreffrath:
Fixes #826

### DIFF
--- a/prboom2/src/m_cheat.c
+++ b/prboom2/src/m_cheat.c
@@ -763,6 +763,11 @@ static void cheat_cycle_mobj(mobj_t **last_mobj, int *last_count, int flags, int
 
       mobj = (mobj_t *) th;
 
+      if (mobj->flags & MF_RESSURECTED)
+      {
+        continue;
+      }
+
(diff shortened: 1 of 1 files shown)


Pedro-Beirao (approved):
Looks good, thank you!

thecnoNSMB:
If I'm interpreting the code correctly, with this change the IDDKT cheat will now no longer point to monsters resurrected by arch-viles, is this desired?

fabiangreffrath:
> If I'm interpreting the code correctly, with this change the IDDKT cheat will now no longer point to monsters resurrected by arch-viles, is this desired?

Nope, here's the update:
https://github.com/kraflab/dsda-doom/pull/828

PR #836, opened 2026-01-27
Fixes autoloaded HELP lumps being treated as missing

Adrad1234:
Autoloading a WAD that replaces HELP and/or CREDIT lump, causes those lumps not to appear while while pressing F1, instead defaulting to dsda-doom credits.

### DIFF
--- a/prboom2/src/w_wad.c
+++ b/prboom2/src/w_wad.c
@@ -610,7 +610,7 @@ int W_LumpNumExists(int lump)
 
 int W_PWADLumpNumExists(int lump)
 {
-  return W_LumpNumExists(lump) && (lumpinfo[lump].source == source_pwad);
+  return W_LumpNumExists(lump) && (lumpinfo[lump].source == source_pwad || lumpinfo[lump].source == source_auto_load);
 }
 
 int W_LumpNameExists(const char *name)


Pedro-Beirao:
This looks good, but @andrikpowell worked on the HELP/CREDIT loop, so can you review it?

andrikpowell:
> This looks good, but @andrikpowell worked on the HELP/CREDIT loop, so can you review it?

I recently actually tweaked how I deal with this in Nyan, by adding a `W_PWADLumpNumExists2()` which includes pwad and autoload:
https://github.com/andrikpowell/nyan-doom/commit/b3b7fe2ed726670ea84e6cde3b2d82f0924133c6

Although, I think the function could be named better. I think it's worth differentiating between "pwad" and "pwad or autoload" for certain situations.

andrikpowell:
Ohhhh yeah... hmmm this is tricky...

The problem is, I don't want HELP or CREDIT to be included in autoload... However this is due to my personal setup... basically by default, in my iwad autoload folders, I have widescreen versions of the default IWAD assets (Doom, Doom 2, TNT, Plutonia, Heretic, Hexen, etc).

And I guess the issue is that since that's always in my autoload, the condition of showing the dynamic credits would just never happen in this case.

However from the logic of autoloading for a PWAD, you'd probably want to include the HELP / CREDIT from autoload for that...

I think that's the main problem with the current autoload system is that there's no way to determine iwad vs pwad autoload.

Pedro-Beirao:
Yeah this is why I wanted to get your opinion. I could imagine not wanting autoload to affect this is in some cases.

PR #840, opened 2026-02-03
Save/Load Hexen stats per map (Hub Stats Fix)

andrikpowell:
Not sure if this is the best way, but it does work.

I could be wrong, but I don't think "visited" is needed.

I'm a bit unfamiliar with Hexen's save system compared to DSDA's (Doom)...

>Edit: set this to draft until I fix this up correctly

### DIFF
--- a/prboom2/src/dsda.c
+++ b/prboom2/src/dsda.c
@@ -394,6 +394,10 @@ int dsda_MaxKillRequirement() {
   return dsda_max_kill_requirement;
 }
 
+void dsda_SetMaxKillRequirement(int max_kills) {
+  dsda_max_kill_requirement = max_kills;
+}
+
 void dsda_WatchPTickCompleted(void) {
   dsda_FlipLineActivationTracker();
 }
--- a/prboom2/src/dsda.h
+++ b/prboom2/src/dsda.h
@@ -85,6 +85,7 @@ dboolean dsda_MovieComplete(void);
 void dsda_DecomposeILTime(dsda_level_time_t* level_time);
 void dsda_DecomposeMovieTime(dsda_movie_time_t* total_time);
 int dsda_MaxKillRequirement(void);
+void dsda_SetMaxKillRequirement(int max_kills);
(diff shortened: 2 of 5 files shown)


fabiangreffrath:
What problem does this solve exactly?

andrikpowell:
> What problem does this solve exactly?

Dsda's stat system (kills, items, secrets) is not designed for hexen's hub system.

Say you are in the seven portals and kill 32 enemies, then you go into the guardian of ice and kill 2 enemies, and then come back to the seven portals.

The stats will first show 32 on seven portals, then when entering the guardian of ice, it will still show 32. When you kill 2 enemies, the stats will then say 34 (32+2). If you come back to the seven portals, it will continue to say 34. If I kill 10 enemies in the seven portals, we now have 44 kills. Return back to the guardian of ice, and we still have 44 kills.

So the stats would show:
Seven portals: 32/maptotal
Guardian of ice: 34/maptotal
Return to seven portals: 34/maptotal
Kill enemies on seven portals: 44/maptotal
Guardian of ice: 44/maptotal

This pr (which I need to update with Nyan's implementation), keeps track of kills per map in a hub.

So it will remember that you had 32 kills on the seven portals, and only 2 in the guardian of ice.

So the stats would show:
Seven portals: 32/maptotal
Guardian of ice: 2/maptotal
Return to seven portals: 32/maptotal
Kill enemies on seven portals: 42/maptotal
Guardian of ice: 2/maptotal

So it keeps track of the stats per map in a hub, and restores them when you come back to that map.

fabiangreffrath:
I see, thank you!

andrikpowell:
Ok, i've just updated the PR to what i have in Nyan... Tricky thing is that there may be a better way, i'm just not as familiar with the hexen save system.

Pedro-Beirao (approved):
This seems fine, but Im not familiar enough with Hexen nor familiar enough with **testing stuff** in Hexen to properly review.

PR #841, opened 2026-02-03
Fix fighter flechette exploding on the player when using freelook

andrikpowell:
Yeah so fun thing... while using freelook and playing the Fighter class in Hexen, throwing a flechette would result in it blowing up where the player was...

This fixes that.

### DIFF
--- a/prboom2/src/dsda/aim.c
+++ b/prboom2/src/dsda/aim.c
@@ -36,7 +36,7 @@ fixed_t dsda_PlayerSlope(player_t* player)
 
 int dsda_PitchToLookDir(angle_t pitch)
 {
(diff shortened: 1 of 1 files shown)


elf-alchemist:
Is this just a signedness issue? Could it work with just `(int64_t)` instead of `(long long)(int)`

andrikpowell:
> Is this just a signedness issue? Could it work with just `(int64_t)` instead of `(long long)(int)`

It looks like we can use `(int64_t)` instead of `(long long)`, but the `(int)` is required here... when testing just `(int64_t)` the bomb did the same behaviour as it did before under `(uint64_t)`.

Also `(uint64_t)(int)` does not work if you were curious.

Pedro-Beirao (approved):
Good catch!

PR #842, opened 2026-02-03
Fix garbage pixels for software fuzz in low resolutions

andrikpowell:


Pedro-Beirao (approved):
Thanks for looking into this

PR #847, opened 2026-02-18
Print date and time in the Load/Save Game menus in the current locale

### LINKED ISSUE #818: Quicksave name doesn't respect locale's date format
For example, running `dsda-doom` with `LC_TIME=en_GB.UTF-8` env var still results in date like `mm/dd/yy` instead of `dd/mm/yy`.

fabiangreffrath:
Fixes #818

### DIFF
--- a/prboom2/src/SDL/i_main.c
+++ b/prboom2/src/SDL/i_main.c
@@ -68,6 +68,7 @@
 #include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <locale.h>
 
 #include "e6y.h"
 
@@ -292,6 +293,9 @@ int main(int argc, char **argv)
   lprintf(LO_DEBUG, "M_LoadDefaults: Load system defaults.\n");
   M_LoadDefaults();              // load before initing other systems
   lprintf(LO_DEBUG, "\n");
+
+  // Print date and time in the Load/Save Game menus in the current locale
+  setlocale(LC_TIME, "");
 
   /* Version info */
   PrintVer();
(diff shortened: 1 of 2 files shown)


Pedro-Beirao:
Right I had fixed this a few months ago but never did a PR, thanks!

In locales where the year is not abbreviated to 2 numbers, we need to increase `SAVESTRINGSIZE` to 26. Can you do it in this PR? Or I'll make another.

fabiangreffrath:
> In locales where the year is not abbreviated to 2 numbers, we need to increase `SAVESTRINGSIZE` to 26. Can you do it in this PR? Or I'll make another.

This will immediately introduce an incompatiblity with any older savegame.

fabiangreffrath:
We could drop seconds:

```patch
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -1482,7 +1482,7 @@ static void M_QuickSave(void)
   time (&now);
   timeinfo = localtime (&now);

-  strftime(description, sizeof(description), "quick %x %X", timeinfo);
+  strftime(description, sizeof(description), "quick %x %H:%M", timeinfo);

   G_SaveGame(QUICKSAVESLOT, description);
   doom_printf("%s", description);
```

fabiangreffrath:
Or we abbreviate the "quick" string with e.g. a mere "q"?

Pedro-Beirao:
I dont mind making older saves incompatible, we do it pretty often in dsda. But yea we would have to increase the `SAVEVERSION` as well.

Its possible users want multiple saves during a minute, so I think we should keep the seconds.
And I dont like how just the 'q' looks.

Screenshot 2026-02-18 at 15 41 04

fabiangreffrath:
Strictly speaking, we don't even need neither "QUICK" nor "Q" at all. We have an entire savegame page called "Q" just for these.

Pedro-Beirao:
True 😅

PR #848, opened 2026-02-18
Switch right and middle mouse button order

### LINKED ISSUE #782: The inputs for mouse buttons 2 and 3 appear to be reversed
Specifically, the default keybinds say that MB2 is bound to the strafe modifier, and MB3 is bound to walking forward, but with my actual mouse these actions are swapped (right click walks forward, middle click is the strafe modifier). Attempting to bind these mouse buttons keeps the backwards behavior. I'm not sure if this is somehow a problem with my mouse.

I first noticed this issue in 0.28.3 (compiled manually on Fedora Linux), and it reproduces in 0.28.4 as well. My mouse is a Logitech G305.

fabiangreffrath:
Fixes #782


Pedro-Beirao (approved):
Looks good!

PR #837, opened 2026-01-28
Autoload Tweaks

andrikpowell:
- rename `source_pre` -> `source_port_wad` for clarity
- actually utilise `source_port_wad`
- differentiate iwad/main autoload from pwad autoload
- Add 2 autoload lump exist functions
- Add new PWAD lump exist functions (I'd like a better name for this)

This is in response to #836 . Also in turn fixes the the issue that the PR raised.

Let me know if there's a downside from this approach. Also for the love of god help me come up with a better name than `W_PWADLumpNameExists2`.

PR #599, opened 2025-02-04
ENDOOM Features

andrikpowell:
I'm sure you knew this was coming eventually... This PR does not address the drawing of ENDOOM, but rather the behaviour of how it is treated.

Tweaks / Features:
- `ansi_endoom` is no longer tied as the on/off switch for showing the ENDOOM.
- `show_endoom` is now the on/off switch + has an option to only show PWAD ENDOOMs.
- Add an "ENDOOM" menu option
- Disable the ENDOOM if playing/recording a demo (I added this feature because I can't think of a single reason why speedrunners would want to see the ENDOOM when doing demos)

PR #849, opened 2026-02-19
Better ENDOOM Support

andrikpowell:
It's finally here!

This replaces PR #599 .

This is my implementation of ENDOOM for Nyan. It is based off the simpler code from Woof, but with the added support of drawing ENDOOM via OpenGL. Drawing using OpenGL was required in order to draw the ENDOOM in the same window.

I've made the decision to disable the ENDOOM entirely when recording / playing back demos, as most speedrunners don't care to see that when making attempts.

There are now 3 config options for ENDOOM (just 1 in the menu):
- `show_endoom` - This is the menu option, and has 3 options (off, on, smart). Smart is very similar to pwad-only, but allows for the ENDOOM to show for IWADs, if no pwad map lumps are found. I've set the default for this to 2 (smart), as due to the change for disabling ENDOOM for demos, I think it's better to show off artist's ENDOOM art, rather than keep it hidden for casual play.
- `export_endoom` - This option when set to `true`, will show the old terminal based ENDOOM. I called it "export" because it's essentially exporting the ENDOOM data to the terminal... Maybe there's a better name?
- `ansi_endoom` - This option previously allowed enabling/disabling the ENDOOM, while also customising the ansi type. Can we stop trying to tie multiple functionalities to one config :/   This option now only controls the ansi type, and has no bearing on enabling the ENDOOM. Setting this option to 0 will now default to 1 (format_cp437).

A little minor note about the "smart" option... Currently because the PR #837 hasn't been integrated, it doesn't take autoload into account. In Nyan Doom, this is corrected to use `W_PWADLumpNumExists2()`.


elf-alchemist:
Bumping for interest.

Pedro-Beirao:
> Bumping for interest.

Not forgotten, its just that annoying to review 😅 
Need to prioritize things that wont make me burn out.

elf-alchemist:
> Not forgotten, its just that annoying to review 😅 
> Need to prioritize things that wont make me burn out.

Good thing that at oeast most of the new LoCs are battle tested library code from Chocolate, Woof and Nyan 😌 the DSDA specific changes aren't too many, thankfully. Was there any standout issue with this, at any rate?

Pedro-Beirao:
I'll leave this for post v0.30.0

prboom2/src/dsda/configuration.c
```diff
@@ -1324,6 +1324,14 @@ dsda_config_t dsda_config[dsda_config_count] = {
     "invert_analog_look", dsda_config_invert_analog_look,
     CONF_BOOL(0),
   },
+  [dsda_config_show_endoom] = {
+    "show_endoom", dsda_config_show_endoom,
+    dsda_config_int, 0, 2, { 2 }
+  },
+  [dsda_config_export_endoom] = {
+    "export_endoom", dsda_config_export_endoom,
```
Pedro-Beirao, line 1332: `terminal_endoom` maybe is better?

prboom2/src/dsda/endoom.c
```diff
@@ -341,12 +352,24 @@ static void RestoreOldMode(void) {
 }
 #endif
 
+int is_opengl = false;
```
Pedro-Beirao, line 355: Why are you using this variable? I think its better to just use `(V_IsOpenGLMode()` like everywhere else in the codebase.

Pedro-Beirao (commented):
Alright, sorry this took way too long. Im pretty happy with it, just left 2 comments.

Im also thinking that it should probably be disabled by default. What do you think?

prboom2/src/dsda/endoom.c
andrikpowell: I'm pretty sure at this point V_IsOpenGLMode() is unavailable, because the main program has been shut down. This is why this variable exists, so I can track if the player was in opengl, before we lose the ability to detect it at this point.

prboom2/src/dsda/configuration.c
andrikpowell: I could... Altho export makes it super clear that it's dumped into the terminal. Terminal_endoom could be misconstrued as something that is drawn by the application, while it's actually being dumped after the application has exited.

andrikpowell:
> Im also thinking that it should probably be disabled by default. What do you think?

I'm gonna be honest, the terminal endoom as it is, is one of the main reasons I started Nyan Doom. Because on windows it was so bad. So bad, that it made me start a fork of DSDA, because I thought it was unacceptable.

So I personally think that the new endoom should be the default behaviour. I can maybe add a terminal setting in the menu? Most people prob prefer it to be in the same game window (Nyan, woof, uzdoom, etc).

I'm curious if you're concerned about endoom behaviour with dsda-launcher? Tbh you should prob do what I did with nyan-launcher and have it auto detect the terminal endoom. Imo endoom logic should be inside dsda and not the launcher. If you really wanna have the endoom terminal as default for dsda-launcher, you could maybe force an arg I guess.

I tend to prefer the in window endoom in general. I notice I sometimes get garbage characters in the dsda-launcher endoom.

prboom2/src/dsda/endoom.c
Pedro-Beirao: That sounds weird since the function simply compares a variable that doesnt get cleared ever. But whatever, this is contained to this file, Im ok with it.

andrikpowell:
@Pedro-Beirao Latest commit makes this pr good to go right?

PR #850, opened 2026-02-20
Remove path when checking IWAD name

andrikpowell:
- Fixes stuff like paintitdoom.wad

Let me know if I missed something, I used the directory format from `HasTrailingSlash()`, but I'm a bit unfamiliar with how Linux/Mac paths work.

### DIFF
--- a/prboom2/src/SDL/i_system.c
+++ b/prboom2/src/SDL/i_system.c
@@ -451,6 +451,28 @@ static const char *I_GetBasePath(void)
   return executable_dir;
 }
 
+const char* I_FileNameFromPath(const char *path)
+{
+  const char *filename = path;
+  const char *p;
+
+  // Find last path separator
+  for (p = path; *p; ++p)
+  {
+    if (*p == '/'
+#if defined(_WIN32)
+        || *p == '\\'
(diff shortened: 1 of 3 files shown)


fabiangreffrath:
You may want to have a look at `M_BaseName()` in Woof:

https://github.com/fabiangreffrath/woof/blob/d632b50becb2d1977a130a5afeb54699ed8e5963/src/m_misc.c#L219-L243

andrikpowell:
> You may want to have a look at `M_BaseName()` in Woof:
> 
> https://github.com/fabiangreffrath/woof/blob/d632b50becb2d1977a130a5afeb54699ed8e5963/src/m_misc.c#L219-L243

Thanks, my version was pretty similar... I noticed that you just allow `:`, while DSDA's  `HasTrailingSlash` currently only allows that for `AMIGA`...

Should we gate the `:` only for `AMIGA`? (like in `HasTrailingSlash`)

fabiangreffrath:
https://github.com/kraflab/dsda-doom/commit/76ff04a1784b32d5f9c5cc762b2c8346c03bf8e3

PR #851, opened 2026-02-20
Add development release

Pedro-Beirao:
Adds a new CI job to upload the artifacts to a development release. The body of the release are the latest commits that ran in the action, maybe it would be cool to have the whole changelog since the latest release.

See it here https://github.com/Pedro-Beirao/dsda-doom/releases

Mostly taken from https://github.com/Murmele/Gittyup/releases

---

Also went ahead and made the macos arm64 runner use macos15 instead of latest to allow a broader support of versions.


Pedro-Beirao:
Will squash the commits obviously

PR #853, opened 2026-03-02
Fix REKKR bow being offset to the right when using opengl

gera2342:
[Someone mentioned this](https://desuarchive.org/vr/thread/12413141/#12422947) in /vr/doom/. 
doom04
Now matches the software mode.


andrikpowell:
Thanks for the PR.

I'm not a huge fan of this solution. Software clamps `x1` for a reason. I think it would be better to pass the original x1 data before the clamp over to OpenGL instead. See #856

Pedro-Beirao:
Closed in favor of #856

PR #856, opened 2026-03-22
OpenGL wide weapon alignment fix

andrikpowell:
This is kind of a response to #853 .

I didn't like how that PR was changing the software drawing path. Software clamps `x1` for a reason.

I think it would be better to pass the original x1 data before the clamp over to GL instead.


andrikpowell:
~Well seems that this looks slightly wrong in 4x3 ratio, so i'll see if i can revise it.~

look i'm dumb and accidently loaded an old version of DSDA :P ignore this

andrikpowell:
Re-pushed to give equal credit to @gera2342 .

PR was mostly their idea, I just tweaked it so that it applied only to OpenGL and not Software as well.

Pedro-Beirao:
Thank you both!

PR #860, opened 2026-03-25
Bring back the `-loadgame` parameter (but not `-recordfrom`)

### LINKED ISSUE #572: Reintroduce `-loadgame` and `-recordfrom` flags
It was useful for scripting. It was suddenly removed in v0.22.3: https://github.com/kraflab/dsda-doom/blob/master/patch_notes/v0.22.md#v0223

fabiangreffrath:
Fixes #572

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -1919,7 +1919,7 @@ static void IdentifyVersion (void)
 
 static void D_DoomMainSetup(void)
 {
-  int p;
+  int p, slot = -1;
   dsda_arg_t *arg;
   dboolean autoload;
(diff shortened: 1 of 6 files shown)


fabiangreffrath:
Some code paths have been changed in an unrecognizable way, so I couldn't just revert commit f90e37aa08366879d8c3e4739d63eb444cc6ce1c. However, I have tested the code change and it appears to work as expected. I haven't done any exhaustive testing, though, so it may still be possible to hit a combination with other parameters that emerges chaos.

Pedro-Beirao:
Thanks!

PR #861, opened 2026-03-25
Fix demo playback bar desync

andrikpowell:
It's a pretty specific bug that I'm not sure how common it'd be to run across... Basically it's related to the temp game modifiers (fast, respawn, nomo, coop_spawns) and loading save games...

To be more specific, playing a demo front to back is fine and works with no issues... However if you have a demo that one of the above modifiers, AND seek forward using the demo progress bar, it seems that it could desync the demo. This caused a desync because when pressing on the demo progress bar, it is essentially loading a save game.

Again, playing demos in full or just playing normal demos play back fine.  It's only when fast, respawn, nomo, or coop_spawns were active and you clicked on the progress bar.

### DIFF
--- a/prboom2/src/dsda/save.c
+++ b/prboom2/src/dsda/save.c
@@ -142,15 +142,36 @@ static void dsda_UnArchiveContext(void) {
   true_basetic = gametic - true_logictic_value;
 }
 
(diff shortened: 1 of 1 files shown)


Pedro-Beirao (approved):
I like this solution Thanks!

PR #862, opened 2026-03-25
Fix SSG fire "bleeding line"

andrikpowell:
So this does work... I'm partly concerned that it could maybe be slightly taxing on the software renderer?

The PrBoom renderer is annoying because when it comes to drawing the player's weapons, it can sometimes bleed outside of the bounds, due to how it stretches sprites. This clamps it so that it never "wraps around".

This was extremely relevant in some resolutions, where when the SSG was fired on it's lowest point, the top of the SSG firing sprite would "wrap around" to the bottom of the sprite creating a red/orangish line as seen here:

image

My resolution of testing was 2560x1440, where this could happen consistently. It's much more noticeable when running the [Minor Sprite Fix](https://www.doomworld.com/forum/topic/62403-doom-2-minor-sprite-fixing-project-v20-release-updated-112822/) which fixes the SSG firing animation (Nyan includes this [automatically](https://github.com/andrikpowell/nyan-doom/blob/0539454dfbeedb1b575f35adc485ef9a47d05e03/prboom2/src/deh/compatibility.c#L247-L254)).

Related is a PrBoom PR https://github.com/coelckers/prboom-plus/pull/181 and issue https://github.com/coelckers/prboom-plus/issues/95 that tried to fix the opposite with the bleeding on the top of the weapon. Maybe there's a way to simplify these fixes?

### DIFF
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -349,6 +349,9 @@ void R_SetDefaultDrawColumnVars(draw_column_vars_t *dcvars) {
   dcvars->edgeslope = dcvars->drawingmasked = 0;
   dcvars->flags = 0;
 
+  // [AR] mark weapon sprite
+  dcvars->isplayersprite = false;
+
   // heretic
   dcvars->baseclip = -1;
 }
--- a/prboom2/src/r_draw.c
+++ b/prboom2/src/r_draw.c
@@ -351,6 +351,7 @@ void R_SetDefaultDrawColumnVars(draw_column_vars_t *dcvars) {
 
   // [AR] mark weapon sprite
   dcvars->isplayersprite = false;
+  dcvars->pspritepostheight = 0;
 
   // heretic
   dcvars->baseclip = -1;
--- a/prboom2/src/r_draw.h
+++ b/prboom2/src/r_draw.h
@@ -79,6 +79,9 @@ typedef struct draw_column_vars_s
   int                 drawingmasked;
   unsigned int        flags; //e6y: for detect patches ind colfunc()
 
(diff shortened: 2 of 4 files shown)


andrikpowell:
Ok, so new update.

My last fix actually did work for what was in the code, but it had a few specific issues. The issue was really easy to see in [Eye Juice](https://www.doomworld.com/vb/thread/158606).

Basically as a brief explanation of how the doom graphic format works, instead of mapping out pixels via x and y, it instead reads pixels by y columns, and then there's a marker that ends a column and starts the next column. Now transparent pixels in each column are skipped, and when actual pixels are detected, it starts a "post", and continues down until it reaches another transparent pixel or the end of a column.

Now the reason I'm explaining this is that the last "fix" clamped posts to make sure they didn't go out of bounds (and vertically loop back around). This worked for the example of the SSG fire sprite.

But what I failed to consider is that tall doom graphics (> 128 pixels tall), can actually forcibly broken up at the 128 mark. Id Software actually did this with a tall reload shotgun sprite, so it would render correctly.

[See this bug](https://doomwiki.org/wiki/Shotgun_sprite_wraps_vertically_in_Final_Doom) on how using DeuTex for Final Doom messed this up, so the shotgun sprite actually wraps around.

Now back to Eye Juice, which all of the weapon sprites are around 168 pixels tall. This means that tools like SLADE add a manual post break in the middle of the image. So when I said there was an issue with my previous fix regarding Eye Juice, it was that it was clamping each post from bleeding over. The result is that you'd get a weird "missing line" on the weapons from Eye Juice.

My new solution is to go higher up and use the full height of the player weapon sprite, so it doesn't interfere with any posts.

PR #867, opened 2026-04-10
ignore and remove invalid `stats.txt` files

### LINKED ISSUE #866: Allow starting with empty `stats.txt` file
If `stats.txt` file is empty which can happen if engine crashes and truncates this file, subsequent run of the game will crash with `Encountered invalid wad stats` message. However, since absence of `stats.txt` is not treated as error and file is recreated in such case, empty `stats.txt` should also be treated this way, mostly due to the mentioned crash scenario.

fabiangreffrath:
Fixes #866


Pedro-Beirao (approved):
Looks good, thank you!

PR #868, opened 2026-04-11
Add StartupWMClass to desktop file to fix icon display on some Linux desktops

electricbrass:
Some Linux desktop environments, such as KDE, match the running program to the desktop file based on the filename, which appears to be correct already, but others, like GNOME, instead use the StartupWMClass entry of the desktop file. DSDA's was missing this.

Before:
image

After:
image


Pedro-Beirao (approved):
Looks good, thank you!

PR #871, opened 2026-04-14
Ensure integer parsing in config flags does not discard data.

yrriban:
When reading the values stored in the -assign and -update command line variables, sscanf is used to convert integer values.  However, sscanf has no ability to report an inability to parse the string in full, instead simply returning the number of arguments successfully converted. This means a flag value such as "a=1,b=2", will set a to 1 and silently discard the rest.

Here we replace sscanf with strtol.  By setting endptr and checking that its value is '\0' after being called, we ensure that the whole string was indeed read.  We also check the value of errno; the only possible error here should be EINVAL for a value outside the range of a long.

### DIFF
--- a/prboom2/src/dsda/configuration.c
+++ b/prboom2/src/dsda/configuration.c
@@ -15,6 +15,8 @@
 //	DSDA Config
 //
 
+#include <errno.h>
+#include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


fabiangreffrath (approved):
LGTM, thank you!

Pedro-Beirao:
Thank you!

PR #874, opened 2026-04-21
Minor fix to "demo loop" logic and Dynamic Credits

andrikpowell:
This PR has some cleanup to make a few of these functions more clear...

Also I fixed a bit of an obscure condition that Xyzzy ran into, where the wad had:

```
TITLEPIC
DEMO1
CREDIT
DEMO2 - Blank
```

Which mostly worked, but due to how Dynamic Credits tends to replace the first CREDIT screen (since the second CREDIT would show after DEMO2), I added a condition where the PWAD CREDIT would be preferred when only one CREDIT is reached.

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -656,15 +656,41 @@ void D_PageTicker(void)
     D_AdvanceDemo();
 }
 
-// Check whether to skip IWAD Demos
-static int dsda_SkipIwadDemos(void)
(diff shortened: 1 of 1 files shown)


Pedro-Beirao (approved):
Looks good, thanks!

elf-alchemist:
What's left to merge this one?

Pedro-Beirao:
> What's left to merge this one?

Nothing, I just like to keep them open for a while after approving. In case I decide to review it again.

PR #875, opened 2026-04-23
Fix interaction between -warp and negative -skiptic values.

yrriban:
The original code here has two separate clauses to handle -warp not being specified and not being specified, respectively.  However, only the first of them gives negative skiptic values the special handling promised in the flag help text.  The second just compares the current tics to the skiptic value, terminating immediately if it's negative.

This change calculates whether or not we're at the target tic once, then reuses it in both clauses of the if statement.  We also exit early if we're not in skip mode or the gametic isn't positive to reduce the overall nesting needed in this function.


Pedro-Beirao (approved):
Looks good, thank you

PR #879, opened 2026-04-30
Add ID24 music changer linedef support (for Doom / Heretic)

andrikpowell:
Does what it says.

Included Elf on the commit because the code is based off his Woof implementation

Also tweaked console commands to allow distinguishing the front / backside of a linedef.


Pedro-Beirao:
Dont want to merge this in dsda until it is part of a future standard

PR #880, opened 2026-04-30
Don't use Colormap for duplicate palette detection

### LINKED ISSUE #843: Weapon sprite is displaying a vertical "bleed" error - bug seemingly unique to modern versions of DSDA
I noticed that when playing in DSDA-Doom, my new BFG sprite replacement has some weird vertical pixel bleed on the firing frames. Pictured left is DSDA in-game, with some of the bleeding pixels circled in red. The same areas when seen within slade or other ports (UZDoom, Woof, etc) do not feature these stray pixels.

([Image](https://i.gyazo.com/ff4606dda9f468458077df0eeaecda87.png))

Here is a wad with the sprite isolated, replacing all frames of the BFG for convenience: [https://www.doomworld.com/applications/core/interface/file/attachment.php?id=394661](https://www.doomworld.com/applications/core/interface/file/attachment.php?id=394661)

After discussing with users on Discord, we've managed to ascertain that the issue is not present in older copies of DSDA-Doom - the way the sprite renders was changed somewhere between v0.24.3 and v0.26.2, but I'm unable to source any builds from between those versions to identify when the change happened any more accurately.

andrikpowell:
Fixes #843

### DIFF
--- a/prboom2/src/dsda/palette.c
+++ b/prboom2/src/dsda/palette.c
@@ -83,18 +83,12 @@ void dsda_FreePlayPal(void) {
 }
 
 static dboolean dsda_DuplicatePaletteEntry(const byte *playpal, int i, int j) {
-  int colormap_i;
-
   if (
     playpal[3 * i + 0] != playpal[3 * j + 0] ||
     playpal[3 * i + 1] != playpal[3 * j + 1] ||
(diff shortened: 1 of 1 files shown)


fabiangreffrath (approved):
I have no idea why this code was introduced in the first place, but apparently it creates at least one more bug than it fixes.

PR #881, opened 2026-04-30
Fix tntfast cheat not toggling fast monsters

### LINKED ISSUE #676: TNTFAST cheat broken
The [TNTFAST](https://doomwiki.org/wiki/Source_port_cheat_codes#Boom_.2F_MBF) cheat seems to have no effect and always prints "fast monsters on" each time you use it (or, if you launched with the `-fast` parameter, it always prints "fast monsters off", again with no effect). Reloading the level does not help.

As a workaround, the "fast monsters" toggle in the compatibility menu or the `-fast` parameter can be used instead.

nathan-casabieille:
The `tntfast` cheat was broken: it always printed "Fast Monsters On" and had no effect on monster speed.

**Root cause:** `cheat_fast()` in `m_cheat.c` was toggling `fastparm` directly, then calling `dsda_RefreshGameSkill()`. That function calls `dsda_ResetGameModifiers()` when `allow_incompatibility` is true (normal gameplay), which reads `fastparm` back from `dsda_config_fast_monsters` - overwriting the toggle immediately.

**Fix:** Use `dsda_ToggleConfig(dsda_config_fast_monsters, true)` instead, consistent with the pattern used by `cheat_hom()` and other cheats. This updates the config value so `dsda_ResetGameModifiers()` reads the toggled state correctly.

Fixes #676

### DIFF
--- a/prboom2/src/m_cheat.c
+++ b/prboom2/src/m_cheat.c
@@ -815,8 +815,8 @@ static void cheat_hom()
 // killough 3/6/98: -fast parameter toggle
 static void cheat_fast()
 {
-  dsda_AddMessage((fastparm = !fastparm) ? "Fast Monsters On" : "Fast Monsters Off");
-  dsda_RefreshGameSkill(); // refresh fast monsters
+  dsda_AddMessage(dsda_ToggleConfig(dsda_config_fast_monsters, true) ? "Fast Monsters On" : "Fast Monsters Off");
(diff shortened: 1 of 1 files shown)


andrikpowell:
Heh, this is what Nyan Doom already does, and since it updates `fastparm` via `dsda_ResetGameModifiers()` this should be good 👍

Pedro-Beirao:
Thank you!

PR #882, opened 2026-04-30
m_menu: fix crash when pressing backspace in save dialog without binding

### LINKED ISSUE #833: Not binding Keybinds/Menus/Backspace to actual backspace key ("back") results in crash in save game dialog when using backspace key
I had bound this option to "Q" only in order to use WASD in menus comfortably. Trying to save a game I wanted to correct something in the savefile name and hit backspace out of habit - game crashes without writing a savegame and giving the following message "createPatch: unknown patch format: PLAYPAL"

Binding "backspace" to both "Q" and "Back" resolved this.

nathan-casabieille:
Pressing backspace while editing a save name crashes with `createPatch: Unknown patch format PLAYPAL` if `input_menu_backspace` isn't bound to the physical backspace key.

The character filter in `M_SaveResponder` accepts `ch >= 32 && ch <= 127`. Since `KEYD_BACKSPACE = 127` and `HU_FONTEND = 0x7f`, the character lands inside the font array bounds but with no valid lump — it defaults to lump 0 (PLAYPAL), which `createPatch` can't parse.

Handle `KEYD_BACKSPACE` explicitly before the printable character check, and tighten the filter to `ch < KEYD_BACKSPACE`. Backspace now also works as a delete key in the save name field regardless of the binding.

Closes #833

### DIFF
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -5773,7 +5773,18 @@ static dboolean M_SaveResponder(int ch, int action, event_t* ev)
     }
     else if (ch > 0)
     {
-      if (ch >= 32 && ch <= 127 &&
+      if (ch == KEYD_BACKSPACE)
+      {
+        if (saveCharIndex > 0)
+        {
+          if (!strcmp(savegamestrings[saveSlot], dsda_MapLumpName(gameepisode, gamemap)))
+            saveCharIndex = 0;
+          else
+            saveCharIndex--;
+          savegamestrings[saveSlot][saveCharIndex] = 0;
+        }
+      }
+      else if (ch >= 32 && ch < KEYD_BACKSPACE &&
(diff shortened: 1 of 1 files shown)


fabiangreffrath:
We should probably just hard-code the basic menu inputs (which this PR already kind of does) and not let them bind/unbind individually. We have done this in Woof and nobody complained that it became impossible to bind "backspace" to a different key than "backspace", yet.

Pedro-Beirao:
> We should probably just hard-code the basic menu inputs (which this PR already kind of does) and not let them bind/unbind individually. We have done this in Woof and nobody complained that it became impossible to bind "backspace" to a different key than "backspace", yet.

Agreed. I will look into it after the menu mouse PR is merged

Pedro-Beirao:
Closing in favor of #934

Issue #883, opened 2026-05-04
additional dehacked text mnemonics


elf-alchemist:
Some dehacked strings available in Woof!/Crispy aren't available in DSDA. The MBF beta emulation pickup sprites are useful for creating pickups that do not affect the player directly, but still leave a message.

```cpp
#define HUSTR_SECRETFOUND "A secret is revealed!"

// [NS] Beta pickups.
#define BETA_BONUS3 "Picked up an evil sceptre."
#define BETA_BONUS4 "Picked up an unholy bible."
```

The rest of the pickup implementation is very minimal:
https://github.com/fabiangreffrath/woof/blob/539e826c7dc383a61d2ce5c4c1131bafecbc3d84/src/p_inter.c#L346-L352

andrikpowell:
Hmm that's weird... I don't even see the ability to pickup an evil sceptre nor bible in `P_TouchSpecialThing()`... nor even the messages for them...

The secret is revealed message is easy enough to add.


Edit: ah so what you really want is for us to add the pickup functionality for MBF pickups? Also doesn't that mean that picking them up does nothing??? I like the idea of allowing these powerups, but I am a bit concerned if a wad tries to rely on them and if other ports support this differently.

elf-alchemist:
> Hmm that's weird... I don't even see the ability to pickup an evil sceptre nor bible in `P_TouchSpecialThing()`... nor even the messages for them...

Yes, DSDA is indeed missing the message pickup case on the P_TouchSpecialThing switch statement.

I previously added them on Woof as thus: https://github.com/fabiangreffrath/woof/commit/70aa28dc827836a0d0bba8d5cb60830af9380a40

andrikpowell:
> > Hmm that's weird... I don't even see the ability to pickup an evil sceptre nor bible in `P_TouchSpecialThing()`... nor even the messages for them...
> 
> Yes, DSDA is indeed missing the message pickup case on the P_TouchSpecialThing switch statement.
> 
> I previously added them on Woof as thus: [fabiangreffrath/woof@70aa28d](https://github.com/fabiangreffrath/woof/commit/70aa28dc827836a0d0bba8d5cb60830af9380a40)

My slight concern is how other ports deal with this and if they allow picking up said items...

Cuz like say I added this into a wad for a custom message, if other ports don't see this as a pickable item or have a message, then my wad breaks in their port.

[...]

PR #885, opened 2026-05-05
Add Customizable secret message via Dehacked

andrikpowell:
Part of #883 .

This is pretty straight forward.

I don't really like that we allow parsing deh files in Heretic here (secret message shows in Heretic). In Nyan, the dehacked / hehacked is completely rewritten, so there is a distinct difference in what's allowed (altho I do allow [some](https://github.com/andrikpowell/nyan-doom/blob/master/docs/hehacked.md#extra-non-hehacked-nyan-doom-strings) doom strings in Heretic, which this secret message will be one of them).


Pedro-Beirao (approved):
Looks good, thank you!

PR #888, opened 2026-05-05
Document using "x-update-baseline" to use the VS2026 vcpkg

KernelLeak:
The "C/C++" workload for Visual Studio installs it's own version of vcpkg by default, but this version expects adding a "builtin-baseline" setting to `prboom2/vcpkg.json`.

The easiest way to do this is open a Visual Studio Developer Command Prompt, change to the `prboom2` folder and run

```
vcpkg integrate install
vcpkg x-update-baseline --add-initial-baseline
```

which will add the required setting and should allow Visual Studio's CMake support to just build everything.

This took me a while to figure out, so I thought I'd submit a PR to add this to the docs, and maybe add the setting to `vcpkg.json`...

### DIFF
--- a/docs/guides/building_on_windows.md
+++ b/docs/guides/building_on_windows.md
@@ -40,6 +40,19 @@ vcpkg integrate install
 
 **IMPORTANT: vcpkg does not update itself automatically.** To get newer versions of libraries, you will need to regularly come back and run `git pull`. Similarly, to update the tool, run `./bootstrap-vcpkg.bat -disableMetrics` again after pulling the changes.
 
+### Using the vcpkg installed by Visual Studio
+
+The "C/C++" workload for Visual Studio installs it's own version of vcpkg by default, but this version expects adding a "builtin-baseline" setting to `prboom2/vcpkg.json`.
+
+The easiest way to do this is open a Visual Studio Developer Command Prompt, change to the `prboom2` folder and run
+
+```
+vcpkg integrate install
+vcpkg x-update-baseline --add-initial-baseline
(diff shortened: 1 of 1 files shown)


KernelLeak:
I haven't tested whether adding a "builtin-baseline" setting to `vcpkg.json` conflicts with a manual installation of vcpkg - in that case I'd just add the first commit, but leave out the second one.

Pedro-Beirao:
Seems to break in the MSVC CI

KernelLeak:
I've pushed just the documentation change; I'm pretty sure the MSVC x64 build will still be busted (and has been since https://github.com/kraflab/dsda-doom/pull/885), but I'm afraid fixing that exceeds my CMake/vcpkg knowledge - sorry... :(

(I'm mainly doing C# and Typescript at work, and last worked on anything C based 15 to 20 years ago... all I can tell is it's something to do with ZLIB dependencies that suddeny don't resolve anymore... :/)

andrikpowell:
> The "C/C++" workload for Visual Studio installs it's own version of vcpkg by default, but this version expects adding a "builtin-baseline" setting to `prboom2/vcpkg.json`.

I'm confused why we need this, when I'm able to just open dsda-doom in Visual Studio 2022 (Windows) and it just sets everything up for me.

Is this a specific requirement that Visual Studio 2026 added / does it act differently than older versions?

Pedro-Beirao:
I will close this for now. but if someone raises an issue about this, I will reconsider.
Thank you for the PR anyway!

PR #889, opened 2026-05-05
Teleporters optimization to fix lethal Oversaturation lag

KernelLeak:
After ZomB1986's pull request in https://github.com/kraflab/dsda-doom/pull/514 wasn't merged, it always bugged me that MAP04 of Oversaturation basically turns into a slideshow at about two thirds in.

After looking at the situation for a bit, I fixed the problem by leaving the `MT_TELEPORTMAN`-type thinkers where they are, but instead building a separate list of just those thinkers that gets tossed as soon as such a thinker is added or removed, and rebuilt the next time `P_TeleportDestination` is looking for a teleport target.

While i was at it, I also sorted those collected thinkers by their sector-id (and am only using the one with the lowest index in the thinkers list per sector-id), allowing the linear scan across all thinkers in `P_TeleportDestination` to be replaced with a binary search in just the array of all teleport-thinkers. In the case of Oversaturation, that means doing a binary lookup in an array of 959 entries instead of scanning 90000+ thinkers each and every time.

To test this, I ran [JonOfAllGames' ITYTD-max demo of Oversaturation MAP04](https://dsdarchive.com/wads/oversat?level=Map+04) at full tilt until about 208 minutes in, then made [a save file](https://drive.google.com/file/d/1FoDJ8NVTCQQV7oqOftjQ63FuTUlmlwho/view?usp=drive_link) - just hit the button, dispatch the Pain Elementals, drop down, go left, kill the Spider Demon and shoot the recessed button to [start the slideshow](https://youtu.be/fzdMrEXks5Y) (less than 1 FPS on my machine), or [with my changes go on killing things normally](https://youtu.be/Ej2Rk29ewlk) (50-100 FPS depending on which part of the horde you stare down).

I'm pretty confident this should not cause any demo desyncs, and while it probably won't matter much on smaller maps (though it also shouldn't worsen performance) it will probably make a difference on other big slaughtermaps.,

### DIFF
--- a/prboom2/src/p_telept.c
+++ b/prboom2/src/p_telept.c
@@ -144,14 +144,8 @@ static mobj_t* P_TeleportDestination(short thing_id, int tag)
 
   FIND_SECTORS(id_p, tag)
   {
-    register thinker_t* th = NULL;
-    while ((th = P_NextThinker(th,th_misc)) != NULL)
-      if (th->function == P_MobjThinker) {
-        register mobj_t* m = (mobj_t*)th;
-        if (m->type == MT_TELEPORTMAN  &&
-            m->subsector->sector->iSectorID == *id_p)
-            return m;
-      }
+    register teleport_t* teleport = P_FindTeleport(*id_p);
+    if (teleport) return teleport->mobj;
   }
   return NULL;
 }
--- a/prboom2/src/p_tick.c
+++ b/prboom2/src/p_tick.c
@@ -68,6 +68,8 @@ int init_thinkers_count = 0;
 
 void P_InitThinkers(void)
 {
+  P_CleanTeleports();
+
   int i;
 
   for (i=0; i<NUMTHCLASS; i++)  // killough 8/29/98: initialize threaded lists
@@ -120,6 +122,13 @@ void P_UpdateThinker(thinker_t *thinker)
 
 void P_AddThinker(thinker_t* thinker)
 {
+  if (thinker->function == P_MobjThinker)
+  {
+    mobj_t* m = thinker;
+    if (m->type == MT_TELEPORTMAN)
+      P_CleanTeleports();
+  }
+
   thinkercap.prev->next = thinker;
   thinker->next = &thinkercap;
   thinker->prev = thinkercap.prev;
@@ -188,6 +197,13 @@ void P_RemoveThinkerDelayed(thinker_t *thinker)
 
 void P_RemoveThinker(thinker_t *thinker)
 {
+  if (thinker->function == P_MobjThinker)
+  {
+    mobj_t* m = thinker;
+    if (m->type == MT_TELEPORTMAN)
+      P_CleanTeleports();
+  }
+
   R_StopInterpolationIfNeeded(thinker);
   thinker->function = P_RemoveThinkerDelayed;
 
@@ -267,6 +283,8 @@ static void P_RunThinkers (void)
 
 void P_CleanThinkers (void)
 {
+  P_CleanTeleports();
+
   for (currentthinker = thinkercap.next;
        currentthinker != &thinkercap;
        currentthinker = currentthinker->next)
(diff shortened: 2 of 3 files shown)


Pedro-Beirao:
This is great, thank you! Ive had this in my TODO list for quite some time.
But Im seeing that the CI is failing. You should probably enable CI in your fork.

I should say that it will take awhile for me to properly review this, just know it wont be forgotten.

KernelLeak:
Eh, it was unfixed for so long it can wait a bit longer... :)

I'll look into enabling CI (I only have VS2026 and Windows here), but is only being able to see "Error:" without any error message (and for every CI step, too) normal?

Pedro-Beirao:
> but is only being able to see "Error:" without any error message (and for every CI step, too) normal?

I see the error logs fine. Not sure what you are talking about.

fabiangreffrath:
I haven't looked at the code in detail, but as an alternative approach, couldn't we just add a `thinker_t *teleport` field to `struct sector_t` that contains a linked list to the `mobj_t` pointers of all the teleporter spawn points in that sector?

KernelLeak:
> I haven't looked at the code in detail, but as an alternative approach, couldn't we just add a `thinker_t *teleport` field to `struct sector_t` that contains a linked list to the `mobj_t` pointers of all the teleporter spawn points in that sector?

But a linked list would mean you'd have to do a linear scan again, instead of a binary search in an array that's usually allocated once...

Also, is it just me, or is the MSVC x64 build fail unrelated to my changes? I'm pretty sure I haven't touched anything relating ZLIB... :/

KernelLeak:
> > but is only being able to see "Error:" without any error message (and for every CI step, too) normal?
> 
> I see the error logs fine. Not sure what you are talking about.

Sorry, user error - uMatrix was blocking the XHR requests to fetch the logs in Firefox.

fabiangreffrath:
> But a linked list would mean you'd have to do a linear scan again, instead of a binary search in an array that's usually allocated once...

That'd be one list *per sector* that should contain either no or exactly one element in most cases.

KernelLeak:
True, but I was going for being able to trash the extra structure I built here easily - is it guaranteed that no sectors or thinkers are going to be added, removed or changed during the game loop?

fabiangreffrath:
You could stack the teleport mobj pointers in variable sized arrays and add or remove them as needed.

fabiangreffrath:
> You could stack the teleport mobj pointers in variable sized arrays and add it remove them as needed.

And even this might be too complicated. Let's have a look at the loop in question. 

It iterates through all the sectors with the given tag. It takes the first sector that it finds and then iterates through all thinkers to see if it finds a mobj thinker for a teleport spawn point that's in that sector. Again, it uses the first one that it finds. There is no `continue` or similar in that loop.

So, there is a 1:1 connection between a teleport spawn point and its sector. I guess, we just have to add a pointer field to `struct sector_t` that points to the *first* mobj thinker for a teleport spawn point that's added to this sector. Should we ever remove a spawn point mobj, we will have to check if it is the one that its sector is pointing to and set this pointer to `NULL`. Whenever we add a spawn point mobj to a sector, we will have to check if the pointer is already set and only have it point to the new mobj if it was `NULL` before.

fabiangreffrath:
Even doomwiki seems to support this theory:

`If there are several teleport destinations in the tagged sector (or sectors), only one will be used; the one with the lowest thing number in the sector with the lowest sector number.`

https://doomwiki.org/wiki/Teleporter

KernelLeak:
> `If there are several teleport destinations in the tagged sector (or sectors), only one will be used; the one with the lowest thing number in the sector with the lowest sector number.`

Which is why I sorted them by sector-id and index...

andrikpowell:
This PR causes desyncs...
After testing 124,950 demos, 577 desync due to this PR...

For an easy demo to test and see this, check out:
https://dsdarchive.com/files/wads/doom2/1169/doomiiii.zip
https://dsdarchive.com/files/demos/doomiiii/82285/d411-780.zip

andrey-budko:
Is this due to a typo here?
`return ((const teleport_t*)a)->index - ((const teleport_t*)a)->index;`
(second `a` should be `b`)

andrikpowell:
so good news, that typo fixes the example demo, and around half of the demo failures (so we now are only at 222 demos failing)... Bad news, we are still getting desyncs.

Here's a demo that desyncs even with the typo fixed:
https://dsdarchive.com/files/wads/doom2/2680/station1.zip
https://dsdarchive.com/files/demos/station1/22653/station1-020.zip

KernelLeak:
Whoops... I really should have noticed that typo... ☹️

fabiangreffrath:
> so good news, that typo fixes the example demo, and around half of the demo failures (so we now are only at 222 demos failing)... Bad news, we are still getting desyncs.

Could you run the demo test with #899 as well, please?

Pedro-Beirao:
Closing in favor of #899

PR #890, opened 2026-05-08
Fix zlib for MSVC static build

andrikpowell:
Ok, so I'm a bit more confident about this now... Although it required me to understand cmake a bit...

Basically the reason builds are failing is because the latest ZLIB update to 1.3.2 changed how static builds are referenced. I'm still a little unclear who's at fault here, vcpkg or zlib...

But the main point is that for static builds, we must now use `ZLIB::ZLIBSTATIC` instead of just `ZLIB::ZLIB`.


rfomin (approved):
Looks good.

PR #895, opened 2026-06-13
fix zdoom automap line style property

elf-alchemist:
adjust enum to use the correct values

### DIFF
--- a/prboom2/src/r_defs.h
+++ b/prboom2/src/r_defs.h
@@ -286,6 +286,7 @@ typedef byte r_flags_t;
 
 typedef enum
 {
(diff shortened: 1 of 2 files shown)


prboom2/src/r_defs.h
```diff
@@ -286,6 +286,7 @@ typedef byte r_flags_t;
 
 typedef enum
 {
+  // Exposed via Action Speicals or UDMF
```
fabiangreffrath: typo: "Speicals"

elf-alchemist: Fixed in 99b96de

Pedro-Beirao:
Thank you!

PR #896, opened 2026-06-14
treat missing patches in texture definitions as non-fatal

elf-alchemist:


andrikpowell:
I realise that I'm gonna be annoying about this... and I realise that Woof is more lenient with missing patches.

But as a modder / mapper, I do actually kinda prefer the fatal approach, as it makes it immediately obvious that I've loaded up the wrong IWAD, or that I've setup up my WAD incorrectly. A majority of wads don't have a problem with this, as they are usually set up correctly.

I do realise that from a user perspective, it kinda makes some sense to be more lenient... altho I do worry about players using the wrong IWAD and it breaking maps.

I'm not sure if it's worth adding an arg that allows the more strict texture checking for modding purposes or not...

I kinda don't like how the "%s seems to be incompatible with %s.\nAre you using the right IWAD?" message was removed, because I think it could be useful feedback for the user after a certain amount of errors... The problem with that how many errors would be needed before a message like that is shown?

Edit:
Ok, to clarify, maybe a pop up message would better than a fatal crash... although that could also be annoying when replaying that particular wad if it was set up incorrectly.

Pedro-Beirao:
I also side with keeping it as a fatal error.

PR #899, opened 2026-06-20
vanilla teleport destination optimzation

elf-alchemist:
alternative to https://github.com/kraflab/dsda-doom/pull/889, pulled from https://github.com/fabiangreffrath/woof/pull/2684

### DIFF
--- a/prboom2/src/p_map.h
+++ b/prboom2/src/p_map.h
@@ -126,6 +126,7 @@ extern mobj_t *crosshair_target;
 extern msecnode_t *sector_list;                             // phares 3/16/98
 extern fixed_t tmbbox[4];         // phares 3/20/98
 extern line_t *blockline;   // killough 8/11/98
+extern mobj_t **sectors_telept;
 
 // heretic
 
--- a/prboom2/src/p_map.h
+++ b/prboom2/src/p_map.h
@@ -126,7 +126,6 @@ extern mobj_t *crosshair_target;
 extern msecnode_t *sector_list;                             // phares 3/16/98
 extern fixed_t tmbbox[4];         // phares 3/20/98
 extern line_t *blockline;   // killough 8/11/98
-extern mobj_t **sectors_telept;
 
 // heretic
 
--- a/prboom2/src/p_maputl.c
+++ b/prboom2/src/p_maputl.c
@@ -371,7 +371,7 @@ void P_UnsetThingPosition (mobj_t *thing)
 
     if (thing->type == MT_TELEPORTMAN)
     {
-        P_ResetTeleptFromSector(thing->subsector->sector - sectors);
+        P_ResetTeleptFromSector(thing->subsector->sector->iSectorID);
     }
 }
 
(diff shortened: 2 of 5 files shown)


andrikpowell:
I haven't tested the entire database, but this causes a desync / crash with this demo:
https://dsdarchive.com/wads/01
https://dsdarchive.com/files/demos/01/2287/011-032.zip (```01 blind.lmp```)

the signal 11 crash happens here:
https://github.com/kraflab/dsda-doom/blob/76ff04a1784b32d5f9c5cc762b2c8346c03bf8e3/prboom2/src/p_telept.c#L194

andrikpowell:
When testing plutonia demos, this one also seems to have a problem:
```Plutonia.wad``` (IWAD)
https://dsdarchive.com/files/demos/plutonia/60033/ple1cn850.zip (```ple1sn850.lmp```)

Note that these crashes seem to be inconsistent, in that sometimes it seems to crash via signal 11, other times it seems to heavily desync? I'm not sure why. I have a concern that some demos may be "randomly passing", since it creates a levelstat.txt, but it ends up in the "heavy desync" category after a while and doesn't crash...

Also both of these demos are very different, but do seem to span multiple levels.

The first demo is a blind playthrough, so there are a couple "map restarts", and it does go into multiple maps.

The second demo is a 2P demo, and continues through multiple maps.

fabiangreffrath:
Well, for some reason dsda-doom removed the zone memory "user" concept, and this is a perfect example of what it was good for. A pointer gets automatically set to `NULL` if it's allocated memory gets freed.

fabiangreffrath:
NB: Checked both demos with my Woof PR and they all run through.

andrikpowell:
Should clarify that on the DPCP server, we found that we had to NULL `sectors_telept` in `P_SetupLevel()`.

The issue was caused because both demos spanned multiple levels and we weren't resetting the teleports per map.

Once I did that, the demos seem to work. I just need to do some batch testing again to fully check if we're good.

fabiangreffrath:
Thank you!

andrikpowell:
A bit of a nitpick, but I prob would've kept `sectors_telept` static, and then added an extern function called `P_FreeLevelTeleports() / P_FreeSectorTeleports()` and call that in `P_SetupLevel()` instead.

I don't see a reason why `sectors_telept` needs to be external, plus the function would make it more clear why it's in `P_SetupLevel()`

fabiangreffrath:
> A bit of a nitpick, but I prob would've kept `sectors_telept` static, and then added an extern function called `P_FreeLevelTeleports() / P_FreeSectorTeleports()` and call that in `P_SetupLevel()` instead.
> 
> I don't see a reason why `sectors_telept` needs to be external, plus the function would make it more clear why it's in `P_SetupLevel()`

Agreed. And please merge the latest changes from the Woof PR.

elf-alchemist:
Addressed in 8f04e847890083ab58216b2e636b031005967ec2

fabiangreffrath:
> Addressed in [8f04e84](https://github.com/kraflab/dsda-doom/commit/8f04e847890083ab58216b2e636b031005967ec2)

Thank you! Now we will need some extensive demo testing for this PR.

andrikpowell:
Got a desync for this demo:
https://dsdarchive.com/wads/teledest
https://dsdarchive.com/files/demos/teledest/87184/teledestp014.zip

All other demos seem to work correctly via the test.

Pedro-Beirao:
Thank you everyone. Ready to merge, yes?

elf-alchemist:
Yes, indeed.

Issue #894, opened 2026-06-12
UDMF: Door Raise cannot be closed while open


0n-s:
UDMF doors, unlike the vanilla Doom doors, cannot be closed as they are opening, or manually closed.

Per this statement (emphasis mine):
> In principle, udmf features supported by both dsda-doom and gzdoom should work the same in each. However, there may be maps that do not work in one port or the other due to differences in the underlying physics, similar to other map formats. **Significant deviation in behaviour is likely a bug.**

This behavior is not mirrored in UZDoom. UZDoom also has a nice feature (also unlike vanilla doors, except in Crispy Doom) where manually closing the door, or opening while the door is closing, plays the appropriate sound again, which would be nice to see.

PR #902, opened 2026-06-22
fix zdoom pusher TID search

elf-alchemist:
Another implementation typo similar to #894, the call to Hexen's TID search isn't correct due to both the DSDA TID and Hexen TID implementations being separated.


elf-alchemist:
Doesn't look like I have permissions to dispatch the workflow, but I still vividly remember this being a networking issue with whatever API the MSVC instance was calling to run the CI job.

Pedro-Beirao (approved):
Looks good, thanks!

PR #903, opened 2026-06-23
Add a track_reality_kill option which causes all player damage to be fatal

Portponky:
This was suggested by PJPlaysDoom who is one of the most prolific reality players/streamers. He thought it would be useful to have a mode where any damage kills the player, as it would be more exciting (for viewers) and more immediate than the track_reality notification.

This PR adds a command line option `-track_reality_kill` which causes any player damage to be increased to 500. It does this after the knockback happens, and it prevents it from happening in sector 11 for almost reality runs on maps like e1m8, so it should be demo compatible for reality runs.

### DIFF
--- a/prboom2/src/dsda/args.c
+++ b/prboom2/src/dsda/args.c
@@ -273,6 +273,11 @@ static arg_config_t arg_config[dsda_arg_count] = {
     "tracks reality and almost reality categories restrictions",
     arg_null,
   },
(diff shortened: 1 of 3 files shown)


Pedro-Beirao:
I dont think this should be merged. The notification could be more noticeable, but I dont want to have a new flag just for this.
PJ can just use your build when doing reality runs. Thank you anyway!

PR #910, opened 2026-07-04
fix non-repeatable `A_LineEffect` behavior being session based, instead of map based

elf-alchemist:
As reported on the Discord server, this PR was tested and compared with DSDA, Woof and Eternity on the map posted there:
https://discord.com/channels/218106532406099969/776233266411339786/1522739662895054961

The buggy behavior in question:
https://youtu.be/WtP0gfkOFGg

Changes pulled from Woof:
https://github.com/team-eternity/WinMBF/blob/3af60d718b1fe272769ee9eeb768b88e29a6cd7e/Source/p_enemy.c#L2548-L2563

This same logic is also in Eternity, so it can be presumed to be have been fixed early on by Quasar, since it is also present in WinMBF:
https://github.com/fabiangreffrath/woof/blob/49f14665d5aee8b0bcf2441056e6e06ce56b56ce/src/p_enemy.c#L2843-L2862

Eternity itself also has some additional changes, but I can only presume these to be due to EE-specific reasons:
https://github.com/team-eternity/eternity/blob/0b5760ee475dfb9bc72c82154a2b7661678ad9a7/source/a_general.cpp#L371-L399

### DIFF
--- a/prboom2/src/p_enemy.c
+++ b/prboom2/src/p_enemy.c
@@ -3160,26 +3160,28 @@ void A_RandomJump(mobj_t *mo)
 
 void A_LineEffect(mobj_t *mo)
 {
-  static line_t junk;
-  player_t player;
-  player_t *oldplayer;
-
   if (compatibility_level < lxdoom_1_compatibility &&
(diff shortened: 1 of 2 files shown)


Pedro-Beirao (approved):
Looks good, thanks!

PR #913, opened 2026-07-08
Dont mute when out of focus when viddumping

### LINKED ISSUE #912: [Enhancement] "Mute when out of focus" shouldn't apply during viddumps.
"Mute when out of focus" as of current still applies when viddumping a demo, which creates unintended mutes when trying to viddump in the background.

Of course, this could be solved by having a seperate config specifically for viddumps, or manually disabling the setting. But this problem can also be solved by ignoring the option when `-viddump` is used.

In my own (limited) testing, this can be very easily fixed by doing this in `dsda/settings.c`.
```patch
 dboolean dsda_MuteSfx(void) {
   return dsda_IntConfig(dsda_config_mute_sfx) ||
-         (!I_WindowFocused() && dsda_IntConfig(dsda_config_mute_unfocused_window));
+         (!I_WindowFocused() && dsda_IntConfig(dsda_config_mute_unfocused_window) && !dsda_Flag(dsda_arg_viddump));
 }
 
 dboolean dsda_MuteMusic(void) {
   return dsda_IntConfig(dsda_config_mute_music) ||
-         (!I_WindowFocused() && dsda_IntConfig(dsda_config_mute_unfocused_window));
+         (!I_WindowFocused() && dsda_IntConfig(dsda_config_mute_unfocused_window) && !dsda_Flag(dsda_arg_viddump));
 }
```
I'm not sure how correct this approach is, but it would be an option.

Pedro-Beirao:
Fixes #912

### DIFF
--- a/prboom2/src/dsda/settings.c
+++ b/prboom2/src/dsda/settings.c
@@ -23,6 +23,7 @@
 #include "g_game.h"
 #include "gl_struct.h"
 #include "lprintf.h"
+#include "i_capture.h"
 #include "i_main.h"
 #include "i_video.h"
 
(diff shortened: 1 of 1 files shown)


fabiangreffrath (approved):
Cool, thanks!

PR #914, opened 2026-07-08
Add support for "TrueColor" OpenGL light fade

JNechaevsky:
This shader mode implements a TrueColor-style smooth fade by interpolating directly between `COLORMAP` rows 0 and 31, bypassing the intermediate 32-level banding. Must be equivalent of implementation from Inter/Nugget, producing seamless lighting transitions without visible steps.

I suppose this might not be entirely canonical, but I'm more than confident there will be those who appreciate this mode.

doom02

### DIFF
--- a/prboom2/data/lumps/gls_main.lmp
+++ b/prboom2/data/lumps/gls_main.lmp
@@ -48,6 +48,28 @@ vec3 indexRGB(float ci, float li)
   return texture2D(colormap, vec2(ci, li)).rgb;
 }
 
+// [PN+JN] TrueColor smooth fade mode.
+// Interpolates directly between COLORMAP row 0 (brightest)
+// and row 31 (darkest), bypassing the intermediate 30 rows.
+vec3 indexRGB_TrueColor(float ci, float li)
+{
+  // Strictly clamp li to 0..31 range (excluding the 32nd invulnerability row!)
+  li = clamp(li, 0.0, 31.0);
+
+  // Normalize to 0.0 .. 1.0
+  float t = li / 31.0;
+
+  // V coordinates for rows 0 and 31 (texel centers)
+  float v_bright = 0.0  / 32.0;
+  float v_dark   = 31.0 / 32.0;
+
+  vec3 color_bright = texture2D(colormap, vec2(ci, v_bright)).rgb;
+  vec3 color_dark   = texture2D(colormap, vec2(ci, v_dark)).rgb;
+
+  // Interpolate between the brightest and darkest colors
+  return mix(color_bright, color_dark, t);
+}
+
 uniform sampler2D tex;
 uniform int fade_mode;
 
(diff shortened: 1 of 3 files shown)


Pedro-Beirao (approved):
Wow I was not expecting this. Thank you!

@elf-alchemist so this wont affect any colormap tricks right? I am not familiar with true color at all.

elf-alchemist:
> @elf-alchemist so this wont affect any colormap tricks right? I am not familiar with true color at all.

Traditionally, the older truecolor implementations perform a fade, from the first colormap row color, to black. This is the kind that breaks effects such as "colormap based faux brightmaps", as seen in in Soft-FX and KDiKDiZD.

In this particular implementation, Julia N. and PVictress figured out how to fade colos between the first and last(!) colormap row colors, this being the key to maintaining the aforementioned effects in both KDiKDiZD and Soft-FX. A quick test with either can show it in action.

JNechaevsky:
Yes, that’s correct. I should add some clarification: despite the excellent visual result, this approach takes a small step back in terms of compatibility with `COLORMAP` brightmap tricks.

In the simplest case, there **won’t be any** issues if a single color is used across all levels here:

изображение

.. And here’s how it looks in-game (the test map is right here [dsda-tc-bmap.zip](https://github.com/user-attachments/files/29863873/dsda-tc-bmap.zip)):

<details>

изображение

</details>

.. However, given that only the first and last rows are actually used, if someone employs a truly exotic brightmap trick where only certain levels are highlighted rather than all of them, the result will differ from the "Smooth" mode and may look incorrect. That said, I honestly have no idea whether such tricks were ever actually used. A trick like this:

изображение

Not sure what else to add, I just have a thing for TrueColor and beautiful colors. Otherwise, DSDA's performance absolutely blew me away, it feels like getting behind the wheel of a premium car after years of driving a single-threaded software dinosaur. @PVictress, anything you'd like to add?

pvictress:
Nothing much to add, really, just curiosity doing its thing, and it all came together as a four-handed effort. This is a GLSL shader; from Inter/Nugget we only borrowed the general idea of using the first and last `COLORMAP` rows. Everything else is self-explanatory, and the original `indexRGB` function served as an excellent foundation. I just added a bit of arithmetic on top.

elf-alchemist:
> изображение

For reference, the use of such effects is non-existent. At best, there was a demo from Linguica on his custom colormap thread -- [which instead of fading, it performs a HSL hue shift](https://www.doomworld.com/forum/post/1422351) -- but as of yet, there's no known examples of such a thing being in use, it is indeed _extremely_ esoteric. The trickiest colormap effects, that actually have any real world use, are already covered, as per above.

Pedro-Beirao:
Thank you for the explanations!

JNechaevsky:
Thank you for the excellent port as well! I believe I'll be using it as my main one from now on. For years I was seeking various eye-candy features, but in the end, all what's _really needed_ was proper TrueColor with amazing and performant OpenGL renderer.

PR #915, opened 2026-07-08
Website for downloads and wiki

Pedro-Beirao:
Based of what I had done for dsda-launcher. Ive wanted to make the readme here more appealing for so long.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,10 @@ CMakeUserPresets.json
 
 # clangd
 /.cache/
+
+# jekyll
+_site
+.sass-cache
+.jekyll-cache
+.jekyll-metadata
+vendor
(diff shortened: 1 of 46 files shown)


elf-alchemist:
image

I like the look much better. Personally, I wouldn't remove the mention of Doom-in-hexen, but I can see why it would be fine to try to not emphasize that one as much. Otherwise, maybe it could receive a complete refresh? The comparisons to PrBoom+/PrBoom+um just seem outdated at this point, since basically everyone has already moved over, there's not much need to use it as a reference point anymore. Maybe it could be saved for a README.old ( maybe some other archived folder)? I don't know thinking out loud, here.

Pedro-Beirao:
Yeah I completely agree with removal of most prboom references. It is still a draft, not sure what should be left in the readme.

I am also in the process of setting up a jekyll page. Im thinking something like https://just-the-docs.com/ would be good enough to host a website for the port, downloads page, and documentation for everything we support

Pedro-Beirao:
Website live at https://pedro-beirao.eu/dsda-doom/ (test url)

I still dont know what domain would be best to use. maybe https://dsdarchive.com/dsda-doom or https://dsda-doom.github.io

PR #916, opened 2026-07-10
Fix keypad Enter displaying as empty string in key bindings menu

JNechaevsky:
A tiny patch. The pointer `s` was assigned the string "PADE", but it was never copied into `menu_buffer`. The code fell through without updating the buffer or the offset, resulting in an empty display for Keypad Enter. This fix adds the missing `strcpy` and necessary offset increment. 

Before and after:
pade

### DIFF
--- a/prboom2/src/m_menu.c
+++ b/prboom2/src/m_menu.c
@@ -4259,8 +4259,11 @@ static int M_GetKeyString(int c,int offset)
     // cph - Keypad keys, general code reorganisation to
     //  make this smaller and neater.
     if ((0x100 <= c) && (c < 0x200)) {
-      if (c == KEYD_KEYPADENTER)
+      if (c == KEYD_KEYPADENTER) {
   s = "PADE";
+  strcpy(&menu_buffer[offset], s);
+  offset+=4;
+      }
       else {
   strcpy(&menu_buffer[offset], "PAD");
   offset+=4;


JNechaevsky:
Actually, doing it this way might be slightly simpler since we can skip the pointer altogether, but the diff would be a bit larger, though:

```c
      if (c == KEYD_KEYPADENTER) {
  strcpy(&menu_buffer[offset], "PADE");
  offset+=4;
      }
```

I'll correct it if needed.

andrikpowell:
heh
https://github.com/andrikpowell/nyan-doom/commit/e52bd723abca8d99549af33fddead1231dd800cb

JNechaevsky:
I thought I was the only one nitpicking over tiny details. Probably that should have just committed to DSDA straight away.

Once we wrap up this PR, how about we also add toggling via Alt + both Enter keys, like it was done in [Chocolate Doom](https://github.com/chocolate-doom/chocolate-doom/commit/75499e054a80aff7add8a0e634f4cdcb3c77274c) back in the day?

Pedro-Beirao:
Thank you!


> Once we wrap up this PR, how about we also add toggling via Alt + both Enter keys, like it was done in [Chocolate Doom](https://github.com/chocolate-doom/chocolate-doom/commit/75499e054a80aff7add8a0e634f4cdcb3c77274c) back in the day?

Agreed

JNechaevsky:
Thanks! Will do as soon I'll get home after weekend.

PR #917, opened 2026-07-10
Do not read patches that are outside the PNAMES table

andrikpowell:
This is similar but also very different from #896 

Basically some wads can have malformed `TEXTUREX` / `PNAMES` set ups, that reference patch indexes outside the full `PNAMES` lookup...

This PR skips over them, and doesn't include such patches in the lookup

The previous behaviour was undefined and resulted in random crashes when loading wads like this (example: https://dsdarchive.com/wads/frost)

For demo testing especially, we need this behaviour to be consistent.

### DIFF
--- a/prboom2/src/r_data.c
+++ b/prboom2/src/r_data.c
@@ -288,19 +288,33 @@ static void R_InitTextures (void)
       mpatch = mtexture->patches;
       patch = texture->patches;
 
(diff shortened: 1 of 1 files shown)


elf-alchemist:
This can be safely merged, both Crispy and Woof already do this.

Pedro-Beirao (approved):
Looks good, thank you!

PR #921, opened 2026-07-12
Add mouse navigation to menus

El-Juancho:
Adds mouse support to the menus, similar to how it works in Woof.

You can move the mouse over menu items to select/highlight them, left click to activate, and right click to go back/cancel. This also works for setup tabs/pages and save/load pages/slots.

The hover highlight brightens the menu item's existing color, so custom menu colors and PWAD palettes should still look right.

The mouse is also released while menus are open in fullscreen/exclusive mode, so the cursor can actually be used there.

Heretic/Hexen menus are laid out a bit differently from Doom's, so I added the extra handling needed for their save slots and sliders to behave properly too.


andrikpowell:
Really cool to see the mouse used in the menus...

Interestingly last night I just implemented CR_LIGHTEN in Nyan, but I did completely rewrite the cr table function so that it was more readable and more clear on the cr variants..

I'm wondering what you think about this method:
https://github.com/andrikpowell/nyan-doom/commit/71ed2798f6a0598e853c031255ae784ed1a5ec10
https://github.com/andrikpowell/nyan-doom/commit/9863515803c6d04fbe494769caf91e463a907114

I did notice that I took a different approach to highlighting save / load slots by highlight the boxes instead of the text (more closer to Woof). I also have extra darken text logic for the save/load slots, but Nyan's "item disabled" system is far more robust compared to DSDA's current "disabled" system.

I was kinda trying to decide whether to highlight items when using the keyboard as well... which I did implement somewhat in Nyan... like if you have a thermo item selected, the thermo itself is now highlighted. But it might make more sense to only highlight items when the mouse hovers over instead of keyboard selected.

andrikpowell:
Actually thinking on this more, I think the highlight part should be a separate PR from the mouse support... cuz they both seem like different features.

El-Juancho:
> Really cool to see the mouse used in the menus...
> 
> Interestingly last night I just implemented CR_LIGHTEN in Nyan, but I did completely rewrite the cr table function so that it was more readable and more clear on the cr variants..
> 
> I'm wondering what you think about this method: [andrikpowell@71ed279](https://github.com/andrikpowell/nyan-doom/commit/71ed2798f6a0598e853c031255ae784ed1a5ec10) [andrikpowell@9863515](https://github.com/andrikpowell/nyan-doom/commit/9863515803c6d04fbe494769caf91e463a907114)
> 
> I did notice that I took a different approach to highlighting save / load slots by highlight the boxes instead of the text (more closer to Woof). I also have extra darken text logic for the save/load slots, but Nyan's "item disabled" system is far more robust compared to DSDA's current "disabled" system.
> 
> I was kinda trying to decide whether to highlight items when using the keyboard as well... which I did implement somewhat in Nyan... like if you have a thermo item selected, the thermo itself is now highlighted. But it might make more sense to only highlight items when the mouse hovers over instead of keyboard selected.


i looked at both commits and i think your variant table method is cleaner and easier to extend than my CR_BRIGHT implementation, specially because it creates the lighter colors directly. i kept mine simple intentionally because i didnt want to refactor dsda's current color translation code in an already pretty big mouse PR.
For save/load i highlighted the text mostly to keep it consistent with the other menus and raven games, and yes i think the highlight makes more sense only when using the mouse.

> Actually thinking on this more, I think the highlight part should be a separate PR from the mouse support... cuz they both seem like different features.

About making it a separate PR, honestly i think the highlight is directly related to mouse navigation, since it gives visual feedback of what the mouse is selecting, specially on tabs and things without a cursor. It would feel weird to me to have mouse navigation without it, but if the maintainers really prefer it separated i can do it.

Pedro-Beirao:
S_CHOICE ("Video Mode" option for example) should cycle when clicked

Pedro-Beirao:
I think its ok to go with this LIGHTEN approach for now. But after switch to @andrikpowell's

Pedro-Beirao:
It is impossible to select "USE TIMEBOMB" in Key Bindings -> Raven menu with the mouse since it sits in the place where the page scrolls.

Pedro-Beirao:
This is a very difficult to review PR. It wouldve been much easier if it was divided in multiple commits.
But it seems safe anyway.

There are quite a few changes that I want to make though, but it would be easier if I make them myself instead of trying to explain.

@El-Juancho is it ok if I make the changes and make a new PR with them? This commit will be included and authored by you

El-Juancho:
> This is a very difficult to review PR. It wouldve been much easier if it was divided in multiple commits. But it seems safe anyway.
> 
> There are quite a few changes that I want to make though, but it would be easier if I make them myself instead of trying to explain.
> 
> @El-Juancho is it ok if I make the changes and make a new PR with them? This commit will be included and authored by you

I'm okay with that :). I fixed both problems you described.

Pedro-Beirao:
I'll merge this, but note that in the future (when I add mouse controls on demo playback) I will implement a widget system and will replace a lot of the logic done here.

`m_menu.c` is just chaos so I dont mind adding all this new logic to it.

PR #922, opened 2026-07-14
Allow to toggle full screen by pressing alt + kp_enter

JNechaevsky:
This allows to toggle full screen mode by not just `Alt+Enter`, but also by `Alt+KPEnter` keys.

### DIFF
--- a/prboom2/src/SDL/i_video.c
+++ b/prboom2/src/SDL/i_video.c
@@ -329,7 +329,8 @@ static void I_GetEvent(void)
             break;
           }
           // Switch windowed<->fullscreen if pressed Alt-Enter
-          else if (Event->key.keysym.sym == SDLK_RETURN)
+          else if (Event->key.keysym.sym == SDLK_RETURN ||
+                   Event->key.keysym.sym == SDLK_KP_ENTER)
           {
             V_ToggleFullscreen();
             break;


Pedro-Beirao (approved):
Thanks!

JNechaevsky:
I’d like to propose two more improvements:

1. There’s a minor bug where the window loses focus upon minimizing. It looks like this:

<details>

https://github.com/user-attachments/assets/824ee491-42de-4dc5-974a-abf074a3c616

</details>

I'm clicking on active window, but nothing happens until I change focus and gain it again. The fix is very simple:

```diff
+++ /dsda-doom/prboom2/src/SDL/i_video.c

@@ -413,6 +414,9 @@
           {
           case SDL_WINDOWEVENT_FOCUS_GAINED:
           case SDL_WINDOWEVENT_FOCUS_LOST:
+          case SDL_WINDOWEVENT_MINIMIZED:
+          case SDL_WINDOWEVENT_MAXIMIZED:
+          case SDL_WINDOWEVENT_RESTORED:
             UpdateFocus();
             break;
           case SDL_WINDOWEVENT_SIZE_CHANGED:
```

<hr>

2. Implement a "throttled mode" that ensures no CPU or GPU load while the window is minimized. 
This is slightly different, inspired by Crispy Doom, but I'm not sure if it's optimal aproach here in DSDA:

```diff
+++ dsda-doom/prboom2/src/SDL/i_video.c
@@ -100,6 +100,10 @@
 
 dboolean window_focused;
+dboolean screen_visible = true;
+
 // Window resize state.
 static void ApplyWindowResize(SDL_Event *resize_event);
 

@@ -417,6 +422,13 @@
             break;
           case SDL_WINDOWEVENT_SIZE_CHANGED:
             ApplyWindowResize(Event);
             break;
+          case SDL_WINDOWEVENT_MINIMIZED:
+            UpdateFocus();
+            screen_visible = false;
+            break;
+          case SDL_WINDOWEVENT_MAXIMIZED:
+          case SDL_WINDOWEVENT_RESTORED:
+            UpdateFocus();
+            screen_visible = true;
+            break;
           }
         }

+++ /dsda-doom/prboom2/src/d_main.c

@@ -602,6 +602,13 @@
     if (players[displayplayer].mo) // cph 2002/08/10
       S_UpdateSounds();// move positional sounds
 
+    extern dboolean screenvisible; // should be in .h file of course
+    if (!screenvisible)
+    {
+      I_uSleep(50000);
+      continue;
+    }
+
     // Update display, next frame, with current state.
     if (!movement_smooth || !WasRenderedInTryRunTics || gamestate != wipegamestate)
     {


```

Pedro-Beirao:
> I'm clicking on active window, but nothing happens until I change focus and gain it again. The fix is very simple:

Sounds good.



> 2\. Implement a "throttled mode" that ensures no CPU or GPU load while the window is minimized.
>         This is slightly different, inspired by Crispy Doom, but I'm not sure if it's optimal aproach here in DSDA:

We already have a way to limit the framerate when out of focus, which reduces the CPU and GPU load.
I dont think we should do a sleep() call.

JNechaevsky:
Alright, I’ll prepare a PR with the first fix then. 

As for the second one, I just have a huge, if not fanatical, obsession with any kind of noise (or rather, the absence of it), including fan noise. But DSDA doesn’t have this issue at 4K/OpenGL on my mini-PC, so this is really just an extra idea to consider.

PR #923, opened 2026-07-14
Fix window focus loss on minimize/restore

JNechaevsky:
This fixes the game window losing focus when minimized and then restored.

Video example:

<details>

https://github.com/user-attachments/assets/c553fa77-4cdd-494c-b45f-2395686cd217

</details>


Pedro-Beirao:
Thank you!

PR #927, opened 2026-07-16
Ci: Add static universal macOS build using vcpkg

Pedro-Beirao:
Continuation of #852 
I thought itd be best to do it all in one PR.

This makes 2 builds on macOS, and then uses `lipo` to merge them.
`macOS arm64 (brew)` stays so that we can keep track of any `brew` issues


Pedro-Beirao:
Self reminder to change the updater script in dsda-launcher

PR #928, opened 2026-07-18
MBF21: Add Weapon Recoil Customisation

andrikpowell:
## Fixing Weapon Recoil for MBF21

_Note: I tested demos with this, and had no desyncs._

"Weapon Recoil" has been added as a weapon property similar like "Ammo per Shot" that allows you to customise the amount of recoil in MBF21.

Currently regarding MBF21 codepointers, I've only added recoil to `A_GunFlashTo` to match `A_GunFlash`

## Current Recoil Breakdown
MBF does apply recoil to:
```
A_GunFlash
A_FirePlasma
A_FirePistol
A_FireShotgun
A_FireShotgun2
A_FireCGun
```

MBF does NOT apply recoil to:
```
A_Punch
A_Saw
A_FireMissile
A_FireBFG
```

I added recoil to:
```
A_GunFlashTo
```

I did NOT add recoil to:
```
A_WeaponProjectile
A_WeaponBulletAttack
A_WeaponMeleeAttack
```

## Ways to deal with this
I think the most important part is to figure out how we wanna deal with recoil in the future...

MBF's implemention is jank because it doesn't apply recoil to the fist, or the chainsaw (note that the default recoil value for the saw was 0 anyway, but the fist has the same recoil value as the pistol, but since it doesn't call a flash, it doesn't have recoil by default). Technically it also doesn't apply to the rocket launcher or BFG, but those indirectly get recoil from the `A_GunFlash`. This is why I'm iffy about adding recoil to MBF21 attack codepointers, because it's already missing from the rocket launcher and BFG.

We could do a clean slate approach by removing the recoil from weapon attack codepointers in MBF21, or we could retroactively add recoil to missing attack codepointers for MBF21.

Another completely separate approach we could do is add a whole new codepointer called like "A_WeaponCustomRecoil" and allow any value as well as direct control over recoil (regardless of OPTIONS `weapon_recoil`). If we did this, I'd connect it with the "Weapon Recoil" property, and if that property isn't used, just go with the default recoil set for that weapon.

### DIFF
--- a/prboom2/src/d_deh.c
+++ b/prboom2/src/d_deh.c
@@ -1351,6 +1351,7 @@ static const char *deh_weapon[] = // CPhipps - static const*
   "Shooting frame", // .atkstate
   "Firing frame",   // .flashstate
   "Ammo per shot",  // .ammopershot [XA] new to mbf21
+  "Weapon recoil",  // .recoil [AR] new to mbf21
   "MBF21 Bits",     // .flags
 };
 
@@ -2525,7 +2526,12 @@ static void deh_procWeapon(DEHFILE *fpin, char *line)
       weaponinfo[indexnum].ammopershot = (int)value;
       weaponinfo[indexnum].intflags |= WIF_ENABLEAPS;
     }
-    else if (!deh_strcasecmp(key, deh_weapon[7]))  // MBF21 Bits
+    else if (!deh_strcasecmp(key, deh_weapon[7]))  // Weapon recoil
+    {
+      weaponinfo[indexnum].recoil = (int)value;
+      weaponinfo[indexnum].intflags |= WIF_ENABLERECOIL;
+    }
+    else if (!deh_strcasecmp(key, deh_weapon[8]))  // MBF21 Bits
     {
       if (bGetData == 1)
       {
--- a/prboom2/src/d_items.c
+++ b/prboom2/src/d_items.c
@@ -64,6 +64,7 @@ weaponinfo_t doom_weaponinfo[NUMWEAPONS+2] =
     S_NULL,
     1,
     0,
+    0,
     WPF_FLEEMELEE | WPF_AUTOSWITCHFROM | WPF_NOAUTOSWITCHTO
   },
   {
@@ -76,6 +77,7 @@ weaponinfo_t doom_weaponinfo[NUMWEAPONS+2] =
     S_NULL,
     S_PISTOLFLASH,
     1,
+    0,
     0,
     WPF_AUTOSWITCHFROM
   },
@@ -90,6 +92,7 @@ weaponinfo_t doom_weaponinfo[NUMWEAPONS+2] =
     S_SGUNFLASH1,
     1,
     0,
+    0,
     WPF_NOFLAG
   },
   {
@@ -102,6 +105,7 @@ weaponinfo_t doom_weaponinfo[NUMWEAPONS+2] =
     S_NULL,
     S_CHAINFLASH1,
     1,
+    0,
     0,
     WPF_NOFLAG
   },
@@ -116,6 +120,7 @@ weaponinfo_t doom_weaponinfo[NUMWEAPONS+2] =
     S_MISSILEFLASH1,
     1,
     0,
+    0,
     WPF_NOAUTOFIRE
   },
   {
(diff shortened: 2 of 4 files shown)


elf-alchemist:
The only thing I disagree with, is that the recoil values array should just be moved into the weapon definitions, if we're doing this we should go all out and just merge it in. Keeping the legacy entry around makes for some awkward code. We donlt have any previous definitions for a "Weapon Recoil" property anywhere else, so it wouldn't cause a desync.

andrikpowell:
> The only thing I disagree with, is that the recoil values array should just be moved into the weapon definitions, if we're doing this we should go all out and just merge it in. Keeping the legacy entry around makes for some awkward code. We donlt have any previous definitions for a "Weapon Recoil" property anywhere else, so it wouldn't cause a desync.

I approached it like this, because that's how "Ammo Per Shot" was implemented... with it being set to 0 in the main table, and then applied via mbf21 dehacked... like otherwise "Ammo Per Shot" would use the default values not in that table.

elf-alchemist:
Additionally, since it was mentioned on the DCPC server, how does this implementation compare to how Eternity already handles recoil? It's worth a look over to make sure it is not _too_ much of a problem to adapt on EE's end.

elf-alchemist:
Actually, I'm thinking a cut back might be more ideal, at least for now, on how recoil is introduced by weapons, and it might be best left for the next feature set spec.

Haxxx showed this following ZDoom code pointer, which performs the task of recoil, and thinking about it this might be more preferred over my idea of retroactively adding the Weapon Recoil property to MBF21, though adding it to A_GunFlashTo is still definitely interesting.

https://zdoom.org/w/index.php?title=A_Recoil

To be more specific, this would allow modders to have more fine-tuned control over different "firing modes" on each weapon, such as with charge-based weapons (example: LoR's Calamity Blade), and it still leaves the door open for altfires, if we ever consider those (not that its planned for at all, anyway, but a single Weapon Recoil _would_ look a bit silly in hindsight if alts were added).

andrikpowell:
## Understanding what MBF did

Ok, I fully understand why MBF has the recoil done the way it does... Even though it does start to get weird when you have custom dehacked.

In MBF, these codepointers receive recoil:
```
A_FirePlasma
A_FirePistol
A_FireShotgun
A_FireShotgun2
A_FireCGun
```

It's worth noting that these codepointers call the flash state, so they do not include `A_GunFlash`.

Things start to get a little weird though with:
```
A_FireMissile
A_FireBFG
```

Those codepointers do not call recoil... but there's a specific reason that they don't. Those 2 weapons call `A_GunFlash` before the actual projectile happens... So to make the recoil fit better with the weapon, recoil was added to `A_GunFlash`.

Note that the fist and chainsaw do have recoil... Although through dehacked, if you call one of the codepointers that triggers recoil, you can get some with the fist slot. The chainsaw slot however has a recoil value of 0, so it can never have any recoil.

## Fixing MBF21 recoil

So now that we know this for MBF, there is an issue with MBF21 and `A_GunFlashTo`... basically if you are doing anything fancy using new MBF21 codepointers, it can never call recoil... So I think that adding recoil to `A_GunFlashTo` would fix this oversight... But there is a catch. Depending on how fancy a setup is, the user needs to be able to customise recoil per weapon, in case they use `A_GunFlashTo` in a unique way (Example: I have an axe weapon that "remembers" it's energy state... So I would probably set the recoil for this weapon to 0 to avoid weirdness)

@elf-alchemist I sorta agree, but also disagree.

I think that we should fix recoil in MBF21 for `A_GunFlashTo` and I think we should add a default recoil property to compensate for if modders don't want recoil for unique uses of the codepointer. The reason I think this is that it is a continuation of MBF's `weapon_recoil` OPTIONS feature, which was overlooked in MBF21.

However I do not think we should add recoil to other MBF21 codepointers because they are often used stacked / together for effects.

## What to do for MBF2y

I do however think that in MBF2y, we should add `A_Recoil` (similar to Eternity / ZDoom) that allows direct control of recoil regardless of the OPTIONS value. By extension of MBF21's "weapon recoil" property can be used as the default recoil value for that codepointer, unless a specific value is used.

What I'm really saying is that in MBF2y, we should disable OPTIONS `weapon_recoil` and just allow `A_Recoil` in dehacked to add kick back directly.

But I think in MBF21, we should fix the current oversight with recoil added to `A_GunFlashTo` and add a default "weapon recoil" property for weapons

elf-alchemist:
Mhm, fine by me! As long as we disable the OG recoil behavior for the next spec it should be good, yes.

Pedro-Beirao (approved):
Looks good! Lets wait until the spec is official and then we can merge this.

PR #929, opened 2026-07-18
OpenGL: Fix textures using old boom colormap while under invulnerabilty

### LINKED ISSUE #925: Sectors being colored due to prior Boom colormaps in OpenGL w/ Invuln
This only appears with OpenGL while the invulnerability powerup is active. 

The bug is weirdly inconsistent; going into the water outside will only make the switch within the Multi-Action Switch room turn blue, while going into the water in the first room will do it to the stairs and switch. So I'm assuming the bug is also potentially location-based.
Additionally, it doesn't fix during map resets.

## Reproduction
* Set vidmode to OpenGL
* Load into [BOOMEDIT.WAD](https://www.doomworld.com/idgames/themes/TeamTNT/boom/boomedit)
* Give player invuln (IDBEHOLDV or similar)
* Drop into the water pool near the start
* Exit and enter the door to the right
* Drop Stair sector is now blue

## Example
Image

## Additional Information
**Version:** v0.29.4
**GPU:** RTX 5080
**OS:** Arch Linux (Kernel v7.1.3-arch1-2)
**Display Manager:** Wayland

andrikpowell:
Fixes #925.

Depending on how a texture was first seen, if it was first seen under a boom colormap, it wouldn't update the texture under the invulnerability colormap

### DIFF
--- a/prboom2/src/gl_texture.c
+++ b/prboom2/src/gl_texture.c
@@ -194,8 +194,8 @@ static void gld_GetTextureTexID(GLTexture *gltexture, int cm)
       bcm = 0;
     }
   }
-  gltexture->texflags_p = &gltexture->texflags[cm][gltexture->player_cm];
-  gltexture->texid_p = &gltexture->glTexExID[cm][gltexture->player_cm][bcm];
+  gltexture->texflags_p = &gltexture->texflags[cm][player_cm];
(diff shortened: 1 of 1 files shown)


Pedro-Beirao (approved):
Good solution, thanks!

PR #932, opened 2026-07-22
Better sounds for menu mouse controls

Pedro-Beirao:
And set g_sfx_menu and g_sfx_pistol as important so that they dont get interrupted

### DIFF
--- a/prboom2/src/hexen/p_acs.c
+++ b/prboom2/src/hexen/p_acs.c
@@ -1669,7 +1669,7 @@ static int CmdSectorSound(void)
         mobj = (mobj_t *) & ACScript->line->frontsector->soundorg;
     }
     volume = Pop();
-    S_StartSoundAtVolume(mobj, S_GetSoundID(StringLookup(Pop())), volume, 0);
+    S_StartSoundAtVolume(mobj, S_GetSoundID(StringLookup(Pop())), volume, false, 0);
     return SCRIPT_CONTINUE;
 }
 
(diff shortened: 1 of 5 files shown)


andrikpowell:
I think we should add these as menu specific aliases (that point to those sounds) instead of using in-game sound aliases

Because it could do weird stuff with in-game sound priority

Pedro-Beirao:
I assumed they were seen as separate, but guess not. Alright need a different approach.

Pedro-Beirao:
@andrikpowell do you like this solution? 2e4eeefe290d69873a8c342df3753b667e115df5

elf-alchemist:
Random suggestion, but perhaps could DSDA potentially include support for Crispy's optional menu sounds?

https://github.com/fabiangreffrath/crispy-doom/blob/7775ef82d1e9dfd50eb9d2824acefaeff7247458/src/doom/sounds.c#L307-L319

Otherwise falling back on the original vanilla sounds, where applicable.

Pedro-Beirao:
> Random suggestion, but perhaps could DSDA potentially include support for Crispy's optional menu sounds?
> 
> https://github.com/fabiangreffrath/crispy-doom/blob/7775ef82d1e9dfd50eb9d2824acefaeff7247458/src/doom/sounds.c#L307-L319
> 
> Otherwise falling back on the original vanilla sounds, where applicable.

Oh I like that. Would like to see these sounds part of a future standard to encourage mappers to use them.

elf-alchemist:
> Oh I like that. Would like to see these sounds part of a future standard to encourage mappers to use them.

We can definitely look into it, but as the next spec will take a bit to solidify, I don't see much harm in adopting the cosmetic feature out-of-spec in the mean time.

elf-alchemist:
For reference, here's the original PR that implemented the feature in question: https://github.com/fabiangreffrath/crispy-doom/pull/783/changes

PR #940, opened 2026-08-09
Name the screenshots "dsda" instead of the generic "doom"

### LINKED ISSUE #933: rename screenshots from `doom*.png` to `dsda*.png`
Something I've had in mind for a bit is to not want to see possible conflicts between different engines when it comes to saving screenies. Woof/Nugget use the EXE name's first 4 letters, and also an extended 4 wide digit count:
https://github.com/fabiangreffrath/woof/blob/5aa6408b8c1a0f1e1a2993dce132011ff880db4f/src/v_video.c#L1142-L1149

The relevant section in DSDA:
https://github.com/kraflab/dsda-doom/blob/d69db3a22a7b1ee6936c82ce7ad25588bc7f6a57/prboom2/src/m_misc.c#L915-L918

Pedro-Beirao:
Fixes #933

### DIFF
--- a/prboom2/src/m_misc.c
+++ b/prboom2/src/m_misc.c
@@ -912,9 +912,9 @@ void M_ScreenShot(void)
     startshot = shot; // CPhipps - prevent infinite loop
 
     do {
-      int size = snprintf(NULL, 0, "%s/dsda%02d" SCREENSHOT_EXT, shot_dir, shot);
+      int size = snprintf(NULL, 0, "%s/dsda%04d" SCREENSHOT_EXT, shot_dir, shot);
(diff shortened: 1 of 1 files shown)


elf-alchemist:
In addition, could the number also be increased to use four digits? i.e `%04d` instead of `%02d`? I forget exactly how the C fornatter handles overflown digit counts here, but if it just starts starts writing 3, then 4, etc digits, then increasing the digit count ahead of time means screenshots can be easily sorted without 'smart sorting', which is something not all file browsing/sorting tool tools have.

elf-alchemist:
Very nice

PR #942, opened 2026-08-09
Added BEX mnemonic for beta pickup sprite

Pedro-Beirao:
Fixes #883 

@elf-alchemist as co-author since it is taken directly from his PR for woof

PR #948, opened 2026-08-11
Add copyright free beta sprites

andrikpowell:
Due to https://github.com/kraflab/dsda-doom/pull/942, I wanted sprites for the messages, so here you go.

Unholy Bible:
image

Evil Sceptre:
image


fabiangreffrath:
Where are these images from? Who's the author? What's the license?

andrikpowell:
> Where are these images from? Who's the author? What's the license?

The author is me. I made it from scratch. Basically looked at the originals and made graphics that looked similar. It's possible they may be a little similar, but that was kinda the point tbh. The goal was to be just different enough, but still keep the same idea.

License can be whatever

Pedro-Beirao:
I do prefer these over the ones in elf's PR.
Only the skull could be considered "too close to the original", but I think its ok.

Might merge this one instead if there are no objections.

elf-alchemist:
> Might merge this one instead if there are no objections.

I know I'm not a juror in this case, but I would be a bit miffed if one port used different assets when I already made PRs for multiple in order to stay consistent (I know, I know, I need to save the christmas spirit here too, I'll fix those as well).

PR #957, opened 2026-08-23
Do pre-tic interpolation for all mobjs

andrikpowell:
So I've noticed a flaw with mobj interpolation...

Basically, usually when interpolation is added at the start of a map, the item thinkers are processed before the lift thinkers....

This means that when they go up and down on lifts, interpolation works correctly.

However there is a flaw, in that if an enemy drops an item, that item ends up getting processed after the lift thinker... resulting in z "jittering" as it uses the wrong interpolation value.

By calculating interpolation for all mobjs before it reaches the thinkers, we fix the jitter problem and ensure their previous z positions are always recorded before the lift thinkers can move them.

The slight downside is that we are forced to use the thinker list processing for every mobj, whereas before we did it manually inside the mobj itself. I'm not sure how much this would affect things, but UZDoom seems to do this right now.

Worth noting that Eternity and Woof have this problem when it comes to items dropped on lifts.

Test wad:
[Lift-test-map.zip](https://github.com/user-attachments/files/31355582/Lift-test-map.zip)

Video of the issue this PR fixes:
https://youtu.be/i_htRQtw_7Q

### DIFF
--- a/prboom2/src/p_map.c
+++ b/prboom2/src/p_map.c
@@ -1866,7 +1866,11 @@ dboolean P_ThingHeightClip (mobj_t* thing)
       (thing->z - thing->floorz < 9 * FRACUNIT) ||
       (thing->flags & MF_NOGRAVITY)
     )
+    {
+      thing->PrevZ = thing->z;
       thing->z = thing->floorz;
+      thing->intflags |= MIF_NOINTERPOLATEZ;
+    }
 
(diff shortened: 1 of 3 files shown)


Pedro-Beirao (approved):
Seems like a nice solution. Thanks!

andrikpowell:
Updated this to be simpler

elf-alchemist:
I'd approve it, again, if I could. This is a much nicer solution than previously 👍

PR #959, opened 2026-08-26
Remove mapinfo

elf-alchemist:
Briefly tested on Eviternity II, appears to work without issue. It's probably for the best. Skill info internal structure maintained for both A) possible future custom skill defs lump, and B) maintaining support for existing custom skill menu work by @andrikpowell.

Part 1 of a 3 step process:
* Remove parsing/use/implementation of ZDoom MAPINFO
* Refactor UMAPINFO parsing/use/implementation code
* Reintroduce previous properties in UMAPINFO


Pedro-Beirao:
Alright, looks great! Onto the next.
Thanks!

PR #962, opened 2026-08-31
Fix "Always Pistol Start" from breaking args when recording demos

andrikpowell:
Yay. Took like a year to figure this one out. :/

Seems that this only happened when you had "Always Pistol Start" on.

Basically the pistol start configs kinda overwrote the flags, and then later on, they were never updated... So this makes sure that flags are always respected.

### DIFF
--- a/prboom2/src/dsda/skill_info.c
+++ b/prboom2/src/dsda/skill_info.c
@@ -227,6 +227,8 @@ void dsda_InitSkills(void) {
 // "Always Pistol Start" is the only persistent cfg (saved in cfg file)
 //
 
+static void dsda_ResetGameModifiers(void);
+
 void dsda_InitGameModifiers(void)
 {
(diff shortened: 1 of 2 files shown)


prboom2/src/dsda/skill_info.c
```diff
@@ -227,6 +227,8 @@ void dsda_InitSkills(void) {
 // "Always Pistol Start" is the only persistent cfg (saved in cfg file)
 //
 
+static void dsda_ResetGameModifiers(void);
+
```
Pedro-Beirao, line 239: Could've just moved the definition of `dsda_ResetGameModifiers` up here

andrikpowell: Technically yes I can do that.

I think my thoughts process was that init should go first, since it's the first function called.

andrikpowell: Fixed :)

PR #963, opened 2026-08-31
Do pre-tic interpolation for all mobjs (Take II)

andrikpowell:
The last method (#957) fixed things on lifts, but added extra jitter to everything else.

This approach is probably better, fixes that issue and avoids checking the thinkerlist every tic.


Pedro-Beirao (approved):
Sounds good, thanks!

PR #965, opened 2026-09-02
Fix view jolting up/down too much when running across tiny height varariations

bangstk:
Players and mappers may have noticed that when running around on sectors which are just 1px or 2px higher than surroundings, the view jolts around up and down a lot, way more than the actual height difference. In some wads with a lot of details on the floor this can be very distracting.

This is caused by the viewbobbing getting reset to 0 when the players enter midair. When running off of a 1px or 2px high ledge, (such as rugs, loose tiles, etc. or other small floor details in maps) the player does enter midair for just 1 tic, so the viewbobbing instantly snaps to 0. Then, on the next tic when the player touches the ground again, the view immediately snaps back to the full bob magnitude. This causes quite a spike (depending on where view was in the bob cycle).

Some example WADs to see this prominently:
The left & right rooms at the start of 10x10.wad MAP01
When Them Demons Cry (onicry.wad)

This is fixed by tracking how long the player has been in midair, and uses this to scale the viewbobbing down to 0 across a few tics instead of instantly. This buffer time absorbs the sudden jolt when running off of 1px/2px/etc high sectors.

This part of the view code affects only the camera view height, and does not affect the weapon sprite, so demo compatibility should be unaffected.

I made sure that the raven nuances are unaffected (in raven the player viewbobbing continues as long as the player is in midair)

Here's a before and after clip of a particularly rough example (running up stairs which have very slightly elevated planks at the edge of each step) and you can also see the phenomenon when walking into and out of the 1px deep dirt areas.

Before:

https://github.com/user-attachments/assets/897bc1d6-b146-47da-b13e-30edcf966c21

After:

https://github.com/user-attachments/assets/f9b3c8e6-1f79-463a-8808-f9368df7b02a

Doom Retro and Z/GZ/UZDoom also address this (by having the player bob continue indefinitely in midair just like raven) but this PR keeps the behavior a lot closer to Doom's (where bobbing does stop in midair - now its just after a couple tics). [A PR is also open for Woof with this same change.](https://github.com/fabiangreffrath/woof/pull/2815)

### DIFF
--- a/prboom2/src/dsda/configuration.c
+++ b/prboom2/src/dsda/configuration.c
@@ -1041,9 +1041,9 @@ dsda_config_t dsda_config[dsda_config_count] = {
     "dsda_weaponbob_pct", dsda_config_weaponbob,
     dsda_config_int, 0, 4, { 4 }
   },
-  [dsda_config_viewbob_floor_jolt] = {
-    "dsda_viewbob_floor_jolt", dsda_config_viewbob_floor_jolt,
-    CONF_BOOL(0)
+  [dsda_config_fix_viewbob_floor_jolt] = {
+    "dsda_fix_viewbob_floor_jolt", dsda_config_fix_viewbob_floor_jolt,
+    CONF_BOOL(1)
   },
   [dsda_config_quake_intensity] = {
     "dsda_quake_intensity", dsda_config_quake_intensity,
--- a/prboom2/src/dsda/configuration.c
+++ b/prboom2/src/dsda/configuration.c
@@ -1041,6 +1041,10 @@ dsda_config_t dsda_config[dsda_config_count] = {
     "dsda_weaponbob_pct", dsda_config_weaponbob,
     dsda_config_int, 0, 4, { 4 }
   },
+  [dsda_config_viewbob_floor_jolt] = {
+    "dsda_viewbob_floor_jolt", dsda_config_viewbob_floor_jolt,
+    CONF_BOOL(0)
+  },
   [dsda_config_quake_intensity] = {
     "dsda_quake_intensity", dsda_config_quake_intensity,
     dsda_config_int, 0, 100, { 100 }
--- a/prboom2/src/dsda/configuration.h
+++ b/prboom2/src/dsda/configuration.h
@@ -229,7 +229,7 @@ typedef enum {
   dsda_config_switch_weapon_on_pickup,
   dsda_config_viewbob,
   dsda_config_weaponbob,
-  dsda_config_viewbob_floor_jolt,
+  dsda_config_fix_viewbob_floor_jolt,
   dsda_config_quake_intensity,
   dsda_config_map_blinking_locks,
   dsda_config_map_secret_after,
--- a/prboom2/src/dsda/configuration.h
+++ b/prboom2/src/dsda/configuration.h
@@ -229,6 +229,7 @@ typedef enum {
   dsda_config_switch_weapon_on_pickup,
   dsda_config_viewbob,
   dsda_config_weaponbob,
+  dsda_config_viewbob_floor_jolt,
   dsda_config_quake_intensity,
   dsda_config_map_blinking_locks,
   dsda_config_map_secret_after,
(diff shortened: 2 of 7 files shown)


andrikpowell:
I saw this from the Woof PR. This looks really cool.

My first thought though is when recording demos, could this be seen as an advantage... Like can recording a demo with or without the view smoothing give an advantage over recording a demo without.

I realise that comparing against vanilla demos may be kinda weird already since we have features like visible stats, but this could change how the game is presented. This probably warrants a discussion on the DSDA server.

For casual play though, this seems really cool, and I'd like to try it out.

bangstk:
Considering you can already turn off the view bobbing entirely, and that is allowed even in strict mode, I don't think this change should be considered as an advantage if that is not.

Pedro-Beirao:
> Considering you can already turn off the view bobbing entirely, and that is allowed even in strict mode, I don't think this change should be considered as an advantage if that is not.

View bobbing is only allowed to be turned off because it gives some people nausea.

This PR introduces a good feature, but I am sure that there will be cases in speedrunning where an alignment is easier without this, so there needs to be setting for this.

bangstk:
I have serious doubts that the original behavior is useful, because of how inconsistent it is (how much the view is jolted depends on where the view currently is in the bob cycle at the exact tic the player walks off of the slightly raised sector, and its a much faster cycle compared to looking at the weapon sprite bobbing to cheese damaging floors). It seems like something that could only be exploitable by TAS.

Note that the only time the view is different from before is the couple of tics after walking off of a raised sector. Outside of this the view bobbing will be identical to how it was before. It is not a general smoothing of the view across all height changes.

But, if a setting is what it takes to get merged I can do it.
Is it ok for the setting to be in the cfg only or do you want it to be in the settings menu too?

Pedro-Beirao:
I think I would prefer it to be in the setting menu. In `Display -> Options`, somewhere around the `View Bobbing` options.

bangstk:
Do you have a preference for the default value?
If not my choice is to turn the fix on by default because the application for the original behavior seems very theoretical and the fix benefits casual players.

Pedro-Beirao:
Yeah I agree, make the fix the default.

bangstk:
Ok, it is now in the config (dsda_viewbob_floor_jolt) and options menu

Screenshot 2026-09-02 114741

andrikpowell:
> I have serious doubts that the original behavior is useful, because of how inconsistent it is

It's not really about being useful. It's if something like this can be exploited in demos. Although, at this time it's kinda hard to see a downside to this, so I'd say it's fine to keep in demos at this time. In the future, if an exploit is found, we can address it then.

I'm sure there may be some purists that would prefer the original behaviour.... For some reason.... So it's nice to have an option.


> Ok, it is now in the config (dsda_viewbob_floor_jolt) and options menu

I'll admit I'm not really a fan that this option is reversed. Like it should be something like "Small Height Variation View Fix", to make it clear that this is something added to the engine. It's fine for it to be enabled by default.

bangstk:
Inverted

Pedro-Beirao (approved):
Perfect, thank you very much!

There is still an ongoing discussion in the dsda discord about allowing this in strict mode, but I am confident in saying that it will be allowed.

bangstk:
If it adds to the strict mode discussion - I should point out that this behavior is already different between DSDA prior to this fix and DOOM2.EXE.

https://github.com/doom-cross-port-collab/boom/blob/2a11d37040be9ac40a61f4c7b7f4541bef2411bf/src/p_user.c#L117
(line 117)

This line, which was active in DOOM2.EXE, but commented out in Boom and derived ports (and omitted entirely in DSDA), causes player->viewheight to be used when the player is in midair in vanilla doom, but not used for midair view height in Boom and up.

This means that in Boom and up, the view jolting is actually much more violent when going up steps with small drops on them such in the video in the OP, than it is in DOOM2.EXE. This is because player->viewheight is involved in the smooth view transition that normally happens when walking up stairs, so not using it in midair makes the view spike even larger.

I will probably have to make a vanilla compatible example wad to demonstrate this.

andrikpowell:
> If it adds to the strict mode discussion - I should point out that this behavior is already different between DSDA prior to this fix and DOOM2.EXE.

Hmm you do make a point that even when playing vanilla demos using voodoo dolls falling in mikoconveyors, it doesn't affect the view of the player in DSDA, like it does in DOOM2.exe. I'm assuming it's something related to this, possibly.

I will admit my main concern was for Vanilla demos, but we already stray away from some vanilla view behaviour it seems.

PR #966, opened 2026-09-02
Better MUSINFO/UMAPINFO Fix

andrikpowell:
In Nyan Doom, I had to rework this slightly to get switching between midi players / soundfonts, to work.

(It's worth noting that in Nyan Doom, in Fluidsynth switching soundfonts does NOT restart the music track, it just changes the soundfont wherever the midi is).

But this also fixes MUSINFO tracks from getting lost when restarted. Also seems to fix UMAPINFO music as well.

I wasn't a fan of #812 nor #951.

Edit: I did make the choice to have unknown MUSINFO track to play "silence". My thought process was as a mapper, it makes it easier to tell when I've made a mistake with MUSINFO.

### DIFF
--- a/prboom2/src/dsda/music.c
+++ b/prboom2/src/dsda/music.c
@@ -96,6 +96,10 @@ void dsda_InitializeMusic(musicinfo_t* source, int count) {
 
   S_music = source;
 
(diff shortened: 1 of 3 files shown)


Pedro-Beirao (approved):
Seems fine, but with music code its not easy to figure out if it breaks.

Thanks!

PR #967, opened 2026-09-02
Add back support for Windows 7

andrikpowell:
I have a friend who has Windows 7 and `windows-latest` makes it so they can't use the latest CI builds.


Pedro-Beirao:
Awesome thanks! We should also use an older macOS version so that older versions are supported.

I'll maybe have to setup a VM to test.

PR #968, opened 2026-09-03
Automap / Minimap Fixes and Tweaks

andrikpowell:
First off, I fixed the minimap coordinates to work correctly. When I rewrote Nyan's hud to allow for multiple fullscreen huds, I learned how you're supposed to do it, and notice my last PR #799 was kinda wrong.

Second, I made tweaks so that the automap doesn't change zoom scaling when changing resolution or renderer.

Third, I replaced the current automap variables with ones that make more sense (and match Nyan). It become harder to follow once the overlay automap was introduced.

This reworking also removed the requirement of having a minimap refresh function inside `V_ChangeScreenResolution`.

### DIFF
--- a/prboom2/src/am_map.c
+++ b/prboom2/src/am_map.c
@@ -850,7 +850,7 @@ void AM_Stop (dboolean minimap)
   automap_full = false;
 
   if (minimap && dsda_ShowMinimap())
-    AM_Start(false);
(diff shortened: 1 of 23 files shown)


Pedro-Beirao (approved):
Looks great! Nice rework.
Always appreciate easier to read code.

PR #969, opened 2026-09-04
Accept string values for -complevel.

yrriban:
These string values match the ones available for the COMPLVL lump, and reuse the same logic after refactoring.


elf-alchemist:
I like this, I miss the aliases from Woof each time I try to use DSDA. These also existed on PrBoom+um funnily enough. There's even a PR ready for UZDoom that adds a `-complevel` arg, which _also_ supports the same `COMPLVL` strings.

Pedro-Beirao (approved):
This is great, thank you!

Yeah I like these aliases, even though I never put any effort into bringing them to dsda-doom.

PR #970, opened 2026-09-05
Fix Optional Sounds Demo Desyncs

andrikpowell:
Move port / optional sounds into separate sfx list
- Append them dynamically after the base game sfx, but before dehextra and dsdhacked
- Allow different sounds per game for optional sounds
- Change raven games to use separate secret lump (SECRET instead of DSSECRET)
- Have raven games respect secret sounds not getting cut off

The main reason I took this approach is it means we can still just reference `sfx_secret` in other parts of the codebase.

### DIFF
--- a/prboom2/src/d_main.c
+++ b/prboom2/src/d_main.c
@@ -105,6 +105,7 @@
 #include "dsda/skill_info.h"
 #include "dsda/skip.h"
 #include "dsda/sndinfo.h"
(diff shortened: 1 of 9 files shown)


Pedro-Beirao (approved):
Looks pretty good, thanks!

Its just the bunch of defines that I dont really enjoy, as I commented before. Its fine anyway.

PR #972, opened 2026-09-06
Optional SFX Fix (Take II)

andrikpowell:
Instead we are going to keep the internal sfx table shifted with the port sounds near the top... Then in dehacked, we will translate the dehacked values into our internal sfx table.

Also Heretic/Hexen use different sound values. I'm not sure what's correct to use for secret / optional sounds, but we need to pay attention, because they do NOT work the same as in Doom.

Take a look at the individual commit to see what I tweaked regarding that.

### DIFF
--- a/prboom2/src/SDL/i_video.c
+++ b/prboom2/src/SDL/i_video.c
@@ -425,7 +425,7 @@ static void I_GetEvent(void)
         break;
 
       case SDL_QUIT:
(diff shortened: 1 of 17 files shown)


Pedro-Beirao (approved):
Alright, back to this solution. Not the prettiest thing, but seems to be fine.

PR #976, opened 2026-09-08
Allow players to advance textscreens in Heretic

andrikpowell:
Also did a E2 palette screen fix that Nyan does, but improved it :)


Pedro-Beirao:
Awesome, thank you very much!

PR #978, opened 2026-09-09
Menu Mouse Refactor

andrikpowell:
You can look at each of my commits to see stuff more in detail. Just a bunch of cleanup and fixing some mouse stuff


Pedro-Beirao:
So calling the menu when clicking during demos is a bad behaviour in my opinion. It would make sense to do that in the demo loop, but not during normal demo playback.

Other than that, great changes! Really good work!

Pedro-Beirao (approved):
Thank you!

PR #979, opened 2026-09-09
Fix Raven Font Kerning

andrikpowell:
Nyan has had this for a while

Changes:
- Raven message font has -1 kerning
- Menu kerning matches messages
- Note that in Vanilla, the finale screens do not use -1 kerning, and instead use 0 kerning, so they are currently correct


Pedro-Beirao:
Awesome, thank you!

PR #982, opened 2026-09-13
Fix incorrect playerstart values in Hexen

andrikpowell:
This used to cause a crash in 0.29.0

I'm not sure why it doesn't now, but we should guard this regardless... maps like HEXDD.wad MAP39 have bad arg values that go outside the accepted values for Hexen's playerstarts.


Pedro-Beirao (approved):
Sounds plausible. Thanks.

PR #983, opened 2026-09-14
Fix "Vanilla Transparency" for strict mode

andrikpowell:
- Add support for strict ranges

Figuring out where to update `deh_changeCompTranslucency()` was tough because you need to call it to reset it for strict, but if you call it too early, dehacked/mobjinfo hasn't been initialised yet.


Pedro-Beirao (approved):
Alright looks good. Good think this was found now.

PR #988, opened 2026-09-15
Menu fixes / tweaks

andrikpowell:
- simplify thermo logic / calls
- thermos can be disabled (removed highlight)
- use disabled instead of strict (disabled includes strict)
- add some failure sounds
- remove `&& CR_BRIGHT`
- replace LoadDef with SaveDef in `M_DrawSave()`


prboom2/src/m_menu.c
```diff
@@ -858,6 +859,15 @@ static void M_DeleteSaveGame(int slot)
 // Load/Save Highlight
 //
 
+dboolean M_MenuItemHighlighted(int item)
+{
+  // Mouse highlight is always enabled
+  if (M_MouseHovered(item))
+    return true;
```
Pedro-Beirao: I assume this is a remnant from nyan? Since the function is useless for us in dsda.

I think its ok to keep it here in dsda since Im sure there will be many more menu changes in the future.

Pedro-Beirao (approved):
Looks good, thanks!

prboom2/src/m_menu.c
andrikpowell: You are correct.

Tbh I think it may be worth just calling `M_MouseHovered()` instead of `M_MenuItemHighlighted()` in dsda.

It's only used in Nyan since Nyan allows highlighting options via keyboard navigation as well, but I don't see dsda adding this in the future.

The only benefit this has is that it keeps the code slightly closer to Nyan, but I don't think it's doing much benefit in dsda.

PR #989, opened 2026-09-15
Hexen / Heretic Killcount Fixes

andrikpowell:
This is a companion pr to #840 

It adds the correct tracking for hexen / heretic thing spawning for the killcount.

Heretic / Hexen actually check when to spawn in a mobj, so we need to make sure to remove monsters from the killcount that aren't actually spawned.

Heretic D'sparil spawns are also taken into consideration.

I did also add a rewrite for the stats, so that map and exhud stats both use the same functions. Also added stats for deathmatch demos... The easiest place to see this is the internal demos for Alien Vendetta.


prboom2/src/p_mobj.h
```diff
@@ -245,9 +245,10 @@ enum {
   MIF_SCROLLING             = (1<<2), // Object is affected by scroller / pusher / puller
   MIF_PLAYER_DAMAGED_BARREL = (1<<3),
   MIF_SPAWNED_BY_ICON       = (1<<4),
-  MIF_FAKE                  = (1<<5), // Not a real thing, transient (e.g., for cheats)
-  MIF_LINEDONE              = (1<<6), // Object has activated W1 or S1 linedef via DEH frame
-  MIF_INTERP_CAPTURE        = (1<<7), // [AR] Capture interpolation once per tic
+  MIF_SPAWNED_BY_DSPARIL    = (1<<5),
```
elf-alchemist, line 248: I was about to say, this messes up saves, but we're ditching compatibility in this update, anyway :P

andrikpowell: This is a bigger deal if this is not getting added to this update. It could be worth, if we are adding this in the future to just add the flag by itself while not adding the full functionality, for savegame compat.

elf-alchemist: No. I mean that you shifted the existing flags over by one bit, which means they would be corrupted when restoring a save from a previous version, which coincidentally isn't happening on this update because we bumped the save version, anyway.

andrikpowell: Well it would become an issue if this update doesn't get this PR, because then the flags would be shifted in a future version

I put MIF_SPAWNED_BY_DSPARIL next to MIF_SPAWNED_BY_ICON because it's basically Heretic's equivalent, so it kinda makes sense to keep them together.

prboom2/src/p_inter.c
```diff
@@ -865,11 +865,15 @@ static void P_KillMobj(mobj_t *source, mobj_t *target)
...
 
-  if (!((target->flags ^ MF_COUNTKILL) & (MF_FRIEND | MF_COUNTKILL)))
+  if (dsda_IsCountedKill(target))
     totallive--;
 
   dsda_WatchDeath(target);
 
+  // Transfer kill to the second form
+  if (heretic && target->type == HERETIC_MT_SORCERER1)
+    target->intflags |= MIF_SPAWNED_BY_DSPARIL;
```
andrikpowell: I don't like this very much at the moment... it's fine for vanilla heretic, but for hehacked, this seems finicky tbh.

Pedro-Beirao (approved):

PR #991, opened 2026-09-15
Disable "Death Use Action: Nothing" in Demos

andrikpowell:
Revert #798 as per discussion on DSDA server


Pedro-Beirao (approved):
Perfect thanks

PR #993, opened 2026-09-16
Heretic UMAPINFO Cleanup / Fixes

andrikpowell:
Various fixes for Heretic UMAPINFO.

A big fix for WI_SHOW_NEXT_DONE, which wasn't working as expected (See commit description for more info).


prboom2/src/w_wad.c
```diff
@@ -648,8 +648,6 @@ int W_PWADLumpNumExists2(int lump)
 
 int W_LumpNameExists(const char *name)
 {
-  if (!name)
-    return false;
```
Pedro-Beirao, line 652: Might be redundant, but it doesnt hurt. I think it should be kept.

andrikpowell: If you ever reach the point where you are passing NULL into `W_LumpNameExists()`, then you are not using the function, the way it was intended. Typically we'd use something like "`W_CheckNumForName() != LUMP_NOT_FOUND`" to avoid NULL issues.

Keeping this in makes the function unclear, and makes it less obvious when you've made a mistake. Whereas getting an error from NULL makes it obvious you'd don't something incorrect.

Although tbh my bigger problem is keeping this in here, while not applying it to every alike function. It feels like a weird bandaid, and feels incomplete. If we do want to go this route for some reason, it should be done to other functions like this. For now, it feels out of place.

Pedro-Beirao (approved):
Eh sure. The whole finale logic is confusing, so trying to go over all possibilities in my head doesnt get me far. I assume it passes demo regression.

PR #997, opened 2026-09-19
Allow spaces before and after value in Complevel Parser

andrikpowell:
This is in response to #996

The issue with that PR is that it disables parsing words like "vanilla" or "boom".

This tweaks the parser to act closer to the previous method to allow spaces before and after the value (whether it be a number or word).

I may have gone a little overboard on comments.


Pedro-Beirao:
> The issue with that PR is that it disables parsing words like "vanilla" or "boom".

No it doesn't? Works fine.

Your solution is more strict, but also adds complexity that I dont think is needed here.

andrikpowell:
> > The issue with that PR is that it disables parsing words like "vanilla" or "boom".
> 
> No it doesn't? Works fine.
> 
> Your solution is more strict, but also adds complexity that I dont think is needed here.

"Vanilla" / "Boom" / "MBF" / "MBF21" all resolve as complevel 0 in your PR, Type TNTCOMP, and you'll see what I mean.

I was wrong in saying it disables parsing... I should've said it doesn't parse them correctly :P

andrikpowell:
Should clarify you get 0 when including spaces before or after the text values:

-complevel "      mbf21     " resolves as complevel 0.

Pedro-Beirao (approved):
Alright, looks good!

PR #998, opened 2026-09-21
Fix `warning: mixing declarations and code is incompatible with standards before C99 [-Wdeclaration-after-statement]`

AbitTheGray:
Just a minor thing that compiler notified me about. I don't care either way, but it was an easy-to-fix warning.


elf-alchemist:
This is PR is fine to merge, but the real question is why haven't we updated to an actually modern C standard version. Why are we still doing C89 as if this was the dark ages or something, even Woof is already using C11. Maybe if we update all this we might actually discover some manner of undefined behavior that the later versions already address which who knows may or may not be the key to fixing some unreproducible Signal 11 crash or another.

andrikpowell:
> This is PR is fine to merge, but the real question is why haven't we updated to an actually modern C standard version. Why are we still doing C89 as if this was the dark ages or something, even Woof is already using C11. Maybe if we update all this we might actually discover some manner of undefined behavior that the later versions already address which who knows may or may not be the key to fixing some unreproducible Signal 11 crash or another.

Well this is only half of the issue. In Nyan, I've fixed all windows warnings and casts, so when I see something like this, I can immediately fix it. So that would help tremendously and avoid situations like this.

Pedro-Beirao:
I agree, we should consider moving to C11.

PR #999, opened 2026-09-21
Hexen's `me1c4537.lmp` test fails due to compiler deciding to follow only defined behavior

AbitTheGray:
Hi, I was messing around with the codebase. I've ended up changing the Ruby testing to C++ GTest, at least to the point that the one demo kept failing in Release build but not in Debug build. I've used AI to bisect the code and it traced it to code that calculates difference. The problem is in CLang optimizing-away signed integer overflow/underflow which it considers an Undefined Behavior.
It was the `HexenEpisode1Skill4MaxByPvs` test (`me1c4537.lmp`) and the problem I've encountered was Hexen-specific.

You can hotfix this using `-fwrapv` argument (or equivalent). Or apply this MR - I may have done more than the absolute minimum of changes, so if you want less, feel free to apply just them.

There is more Undefined Behavior in the codebase. I worry that some people may be disqualified due to no fault of their own.
As I understand it, most players will be fine due to running a Windows version as MSVC does not use those optimizations. GCC should, and I encountered it due to CLang using the optimization in my case.

PS: I've used AI to help find the problem, and discuss compiler behavior due to me not having experience with older C/C++ versions.


elf-alchemist:
Undefined behavior baby!

Pedro-Beirao:
`(int)a - (int)b` is not the same as `(int)(a - b)`

take a = 2.1; b = 1.9

`2 - 1 = 1` is not the same as `(int)(0.8) = 0`

Pedro-Beirao:
So this PR is not safe at all... I gotta complain about AI usage man.

fabiangreffrath:
Wait a minute. None of the code here is dealing with `float`, only `int` -- either `unsigned` or `signed`. And explicitly telling the compiler to use `unsigned` arithmetic when a wrap around is expected is absolutely common sense.

Pedro-Beirao:
Oooh I feel stupid. Right `angle_t` is an `unsigned int`... makes sense.

AbitTheGray:
@Pedro-Beirao you are right with floating points case, obviously.
As for AI, I would probably just ignore it (just report it on Discord) and not bisect it myself. AI had a simple goal and acted as a smarter script, bisecting the code, switching files from `-O0` to `-O2` and running tests. It made it faster to find the problem, but I was the one who made the changes in the branch, hence why there are some unnecessary changes as well :)

And I've discussed source of the problem with the AI, in chat, because I was to lazy to search and read about how different compilers react to this specific UB.

---

If you, in general, do not want this change (or want to solve it a different way), feel free to take this just as a report. I was just worried about some runs of some people not counting due to desync caused by a compiler.

Pedro-Beirao (approved):
All changes seem correct. Thank you!

